PR #1, opened 2025-11-19
RTD documentation bootstrap

izmalk:
Bootstrapping RTD documentation with [starter pack](https://github.com/canonical/sphinx-docs-starter-pack).

For now, the docs are available at the https://canonical-charmed-valkey.readthedocs-hosted.com/en/latest/ address. The address will change slightly later.


izmalk:
The [Check for removed URLs / build-docs (pull_request)](https://github.com/canonical/charmed-valkey-operator/actions/runs/19520180949/job/55881674931?pr=1) test fails because there is no `docs` directory in the targeted branch (`main`).

reneradoi (commented):
Hey @izmalk thank you for proactively setting this up!

For the Valkey operators, we will likely implement some kind of "monorepo" structure for the different flavors that we will have (k8s, vm, non-sharded, sharded). Would the docs structure also work for this?

I know the question is pretty broad given that there is a variety of different ways for monorepo structure. To narrow it a bit, the likely repository structure could look something like this:
```
/
|-.github
|-docs
|-shared-code
|-valkey-k8s-operator
|-valkey-vm-operator
|-valkey-cluster-k8s-operator
|-valkey-cluster-vm-operator
```

We aim at one documentation set for the valkey operators. Do you think the bootstrapped docs structure would work for that?

izmalk:
I don't see any significant problem with the monorepo approach docs-wise. Especially so if we keep the docs directory in the root of the repo.

reneradoi (approved):
Thank you @izmalk!

PR #2, opened 2025-12-19
[DPE-9013] Bootstrap charm and repo structure

reneradoi:
This PR sets up the basic repository structure for the charmed Valkey operators and creates a basic Valkey K8s operator.

Features of the operator:
- deployment on K8s
- starting Valkey on the leader unit, not starting non-leader units (horizontal scaling not implemented yet)
- a config manager that initializes a Valkey config file on the charm unit based on a template 
- charm structure pattern
- advanced statuses
- data interfaces v1 for peer relation data
- basic unit testing (scenario)
- basic integration testing (jubilant, spread + concierge)
- CI workflows for build, testing and release to Charmhub

The charm code targets both K8s and VM substrates. For now, it doesn't include substrate-specific logic (meaning: differentiating per substrate via the cloud type as described [here](https://github.com/canonical/charmed-valkey-operator/wiki/multiple-substrates-design)). This has to wait until [this issue](https://github.com/canonical/operator/issues/2267) in ops is fixed.


common/common/core/cluster_state.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python3
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Objects representing the cluster state of Valkey."""
+
+import logging
+from typing import Dict, Set
```
skourta: We do not need to use these types with Python 3.10+ we can just use dict and set

skourta (approved):
Overall LGTO. Thank you @reneradoi. I left one small comment. I think this will be a great starting point for the charm.

common/common/core/cluster_state.py
reneradoi: Yes, that makes sense. Thanks for pointing it out.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,86 @@
...
+  integration-test:
+    name: Integration test
+    needs:
+      - lint
+      - unit-test
+      - build
+    uses: ./.github/workflows/integration_test.yaml
+    with:
+      artifact-prefix: ${{ needs.build.outputs.artifact-prefix }}
+    secrets: inherit
```
Mehdi-Bendriss: applies to all yaml files incl. metadata.yaml etc
```suggestion
    secrets: inherit

```

src/core/base_workload.py
```diff
@@ -0,0 +1,30 @@
...
+
+"""Base objects for workload operations across different substrates."""
+
+from abc import ABC, abstractmethod
+
+
+class WorkloadBase(ABC):
+    """Base interface for common workload operations."""
+
+    @abstractmethod
```
skourta, line 14: This is a property when implemented in k8s.

src/core/models.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class PeerAppModel(PeerModel):
+    """Model for the peer application data."""
+
+
+class PeerUnitModel(PeerModel):
+    """Model for the peer unit data."""
+
+    started: bool | None = Field(default=None)
```
skourta: Should we just make it always present with a default of False?

src/events/base_events.py
```diff
@@ -0,0 +1,36 @@
...
+
+from literals import PEER_RELATION
+
+logger = logging.getLogger(__name__)
+
+
+class BaseEvents(ops.Object):
+    """Handle all base events."""
+
+    def __init__(self, charm: ops.CharmBase):
```
skourta: We should use `ValkeyCharm` for the charm type

src/workload_k8s.py
```diff
@@ -0,0 +1,73 @@
...
+        if not container:
+            raise AttributeError("Container is required.")
+
+        self.container = container
+        self.config_file = pathops.ContainerPath(CONFIG_FILE, container=container)
+        self.valkey_service = "valkey"
+        self.metric_service = "metric_exporter"
+
+    @override
+    @property
```
skourta:
```suggestion
    @property
    @override
```

skourta (approved):
Looks good to me. Thank you, @reneradoi. One thing to keep in mind is we still do not expose the service, so the users have to do it themselves.

src/events/base_events.py
reneradoi: Yes we should, thank you. That was a leftover from the previous structure.

src/core/models.py
reneradoi: Agree, I've applied this and all your other suggestions - thank you!

.github/workflows/integration_test.yaml
```diff
@@ -0,0 +1,171 @@
...
+        if: ${{ success() || (failure() && steps.spread.outcome == 'failure') }}
+        uses: actions/upload-artifact@v5
+        with:
+          name: logs-integration-test-${{ matrix.job.name_in_artifact }}
+          path: ~/logs/
+          if-no-files-found: error
+      - name: Disk usage
+        timeout-minutes: 1
+        if: ${{ success() || (failure() && steps.spread.outcome == 'failure') }}
+        run: df --human-readable
```
Mehdi-Bendriss:
```suggestion
        run: df --human-readable
        
```

src/charm.py
```diff
@@ -0,0 +1,65 @@
...
+            logger.warning("Container not ready yet")
+            event.defer()
+            return
+
+        if not self.unit.is_leader():
+            logger.warning("Scaling not implemented yet, services not started")
+            return
+
+        self.config_manager.set_config_properties()
+        self.workload.start()
```
Mehdi-Bendriss, line 59: Can you also add the rollingops lib to the lib/charms directory?

pyproject.toml
```diff
@@ -0,0 +1,92 @@
...
+[tool.poetry]
+package-mode = false
+requires-poetry = ">=2.0.0"
+
+[tool.poetry.dependencies]
+python = "^3.12" # switch to 3.14 once charm base is 26.04
+ops = "^3.5.0"
+charmlibs-pathops = "^1.2.0"
+data-platform-helpers = ">=0.1.7"
+
```
Mehdi-Bendriss, line 14: We should also include valkey-glide

pyproject.toml
```diff
@@ -0,0 +1,92 @@
...
+ops-scenario = "*"
+
+[tool.poetry.group.integration.dependencies]
+pytest = "*"
+pytest-asyncio = "*"
+allure-pytest = "*"
+allure-pytest-default-results = "^0.1.2"
+data-platform-helpers = ">=0.1.7"
+jubilant = "^1.6.0"
+
```
Mehdi-Bendriss, line 52: We should also include valkey-glide

Mehdi-Bendriss (commented):
Thanks René, looks good! I have a few small comments

Mehdi-Bendriss (approved):

PR #4, opened 2026-01-20
[DPE-9135]: Internal user management

skourta:
This pull request introduces several significant changes to support secure internal user management, password rotation, and improved dependency handling for the Valkey charm. The main updates include the introduction of a secret-based internal system user, mechanisms for updating and rotating its password via Juju secrets, and the necessary supporting code and configuration changes. Additionally, the build and test workflows are updated to support new dependencies.

**Internal user management and password rotation:**

- Added a new `system-users` config option (as a Juju Secret URI) for securely configuring the internal system user and its password, with auto-generation if not set. (`config.yaml` [config.yamlR1-R11](diffhunk://#diff-d8d0422389f03d783e32e627250fe29834bd09c6361640d1ff00661dd6820034R1-R11))
- Introduced the `INTERNAL_USER` and related constants in `literals.py` to standardize internal user naming and configuration. (`src/literals.py` [src/literals.pyR12-R20](diffhunk://#diff-959ba637043ab4ac88c6f9f9109022f78a7df1b9a2ab3a0407bff4f124c2db64R12-R20))
- Implemented a `ValkeyClient` utility class for connecting to Valkey servers and updating user passwords, including error handling via a new `ValkeyUserManagementError` exception. (`src/common/client.py` [[1]](diffhunk://#diff-f5c8dd5ff5ebd3473ee519d51e59ddb12e43e04365489c1c429256d3d20c3bc4R1-R96) `src/common/exceptions.py` [[2]](diffhunk://#diff-e793bad618b16c64102391f92909ac368210513be36cc7ff3ea633e458e54feeR1-R8)
- Extended the cluster state and models to store and retrieve internal user credentials, including secret resolution and password storage. (`src/core/cluster_state.py` [[1]](diffhunk://#diff-740c8c6f03bd927ed1459e32310a5858ffb39f5bd0e66ab7f0b50df4d7f7a7caR102-R120) `src/core/models.py` [[2]](diffhunk://#diff-e46c88bb687d77b94764a732b3e08a0a7bd544a64e606429c0b787c13e1c226aR12-R35) [[3]](diffhunk://#diff-e46c88bb687d77b94764a732b3e08a0a7bd544a64e606429c0b787c13e1c226aR128-R135)
- Updated the `ClusterManager` to use the new client for updating credentials and to track admin user/password state. (`src/managers/cluster.py` [[1]](diffhunk://#diff-20b77aa4083c3162fac7cd352e1b8ceef186a967a37f906277cb77346967c113R13-R17) [[2]](diffhunk://#diff-20b77aa4083c3162fac7cd352e1b8ceef186a967a37f906277cb77346967c113R32-R56)

**Event handling and password rotation logic:**

- Enhanced the base event handler to react to leader election, config changes, and secret changes, ensuring the internal admin password is set and rotated as needed. This includes updating the password in Valkey and handling errors/statuses appropriately. (`src/events/base_events.py` [[1]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eR8-R15) [[2]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eR34-R36) [[3]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eR47-R151)

**Dependency and build/test workflow improvements:**

- Switched the `valkey-glide` dependency to a specific Git branch and subdirectory for custom build support. (`pyproject.toml` [[1]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L13-R13) [[2]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L51-R51)
- Updated CI and integration test workflows to install required system dependencies for building and testing, including Rust and protobuf. (`.github/workflows/ci.yaml` [[1]](diffhunk://#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR38-R41) `.github/workflows/integration_test.yaml` [[2]](diffhunk://#diff-f61384d930ecca9b388b7d547b0c10b6c977f087b106b0747d64dad87a96daa0R111-R115) `spread.yaml` [[3]](diffhunk://#diff-e1e5ecb5e7fe7aa228c64f680564673541f91452212c3c6c23cbbd27bb78dcedR120)

**General codebase improvements:**

- Added utility and abstract methods for file writing and improved code structure for future extensibility. (`src/core/base_workload.py` [src/core/base_workload.pyR32-R41](diffhunk://#diff-90d9cf1ebee4e42b159cab54e264b9947faa2f1435648c7857df493fc091d2fdR32-R41))
- Minor improvements to code style and configuration, such as linting exclusions. (`pyproject.toml` [pyproject.tomlL84-R88](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L84-R88))
- Added password generation and utility imports in config manager for future use. (`src/managers/config.py` [[1]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cR7-R10) [[2]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cR19)


src/managers/cluster.py
```diff
@@ -26,11 +29,33 @@ class ClusterManager(ManagerStatusProtocol):
...
+        """
+        try:
+            client = ValkeyClient(
+                username=self.admin_user,
+                password=self.admin_password,
+                hosts=self.cluster_hostnames,
+            )
+            client.update_password(username=username, new_password=password)
+            client.password = password
+            client.save_acl()
```
skourta: @reneradoi @Mehdi-Bendriss 
WDYT of doing this vs reconciling at the charm level?

src/common/client.py
```diff
@@ -0,0 +1,109 @@
...
+            Any result from the command.
+        """
+        client = None
+        try:
+            client = await self.create_client()
+            result = await asyncio.wait_for(client.custom_command(command), timeout=5)
+            return result
+        except Exception as e:
+            logger.error(f"Error running command {' '.join(command)}: {e}")
+            raise ValkeyUserManagementError(f"Could not run command {' '.join(command)}: {e}")
```
reneradoi: These two lines could leak passwords in the log output when `_run_custom_command` is called from `update_password`.

src/managers/cluster.py
reneradoi: Not sure if you mean that, but I think it is important that we decide for one approach here: While we use the client commands to update the password on `config/secret_changed`, we write to the acl file on `leader_elected`. Why two different ways?

Personally, I'd prefer the 2nd one: write an updated password to the acl file and call `ACL LOAD`, because I think this would be more resilient (allowing us password updates while the server is not running). But I'm open here, we could also choose the other approach, too, if you prefer for specific reasons.

tests/unit/test_charm.py
```diff
@@ -148,3 +161,231 @@ def test_update_status_non_leader_unit(cloud_spec):
...
+    container = testing.Container(name=CONTAINER, can_connect=True)
+
+    password_secret = testing.Secret(
+        tracked_content={"wrong-username": "secure-password"}, remote_grants=APP_NAME
+    )
+    state_in = testing.State(
+        leader=True,
+        relations={relation},
+        containers={container},
+        config={INTERNAL_USER_PASSWORD_CONFIG: password_secret.id},
```
reneradoi: This doesn't match the name of the test, instead this seems to be the same test as above.

tests/unit/test_charm.py
```diff
@@ -148,3 +161,231 @@ def test_update_status_non_leader_unit(cloud_spec):
...
+def test_config_changed_non_leader_unit():
+    ctx = testing.Context(ValkeyCharm)
+    relation = testing.PeerRelation(id=1, endpoint=PEER_RELATION)
+    container = testing.Container(name=CONTAINER, can_connect=True)
+
+    state_in = testing.State(
+        leader=False,
+        relations={relation},
+        containers={container},
+        config={INTERNAL_USER_PASSWORD_CONFIG: "secret:1tf1wk0tmfrodp8ofwxn"},
```
reneradoi: Shouldn't this be `password_secret.id` instead?

pyproject.toml
```diff
@@ -6,11 +6,11 @@ package-mode = false
 requires-poetry = ">=2.0.0"
 
 [tool.poetry.dependencies]
-python = "^3.12" # switch to 3.14 once charm base is 26.04
+python = "^3.12"                                                                                                                 # switch to 3.14 once charm base is 26.04
 ops = "^3.5.0"
 charmlibs-pathops = "^1.2.0"
 data-platform-helpers = ">=0.1.7"
-valkey-glide = "^2.2.5"
+valkey-glide = { git = "https://github.com/skourta/valkey-glide", subdirectory = "python/glide-async", branch = "add-build-rs" }
```
reneradoi, line 15: Let's add a TODO comment here, and at the other place where it is used, so that we don't forget to change it when the issue was fixed upstream.

reneradoi (changes requested):
Thank you Smail! Overall, the PR looks good, not many concerns. The main thing from my point of view is that we need to decide on a method for updating passwords: either via client command or by writing and loading the acl file (preferred).

src/events/base_events.py
```diff
@@ -38,3 +44,108 @@ def _on_update_status(self, event: ops.UpdateStatusEvent) -> None:
...
+                        INTERNAL_USER
+                    )
+                except (ops.ModelError, ops.SecretNotFoundError) as e:
+                    logger.error(f"Could not access secret {admin_secret_id}: {e}")
+                    raise
+            else:
+                password = self.charm.config_manager.generate_password()
+
+            self.charm.state.cluster.update({"charmed_operator_password": password})
+            self.charm.config_manager.set_acl_file()
```
reneradoi, line 68: We need to ensure that we write the acl file on every unit, not just the leader. Otherwise we could run into a situation where we experience corrupted auth after a leader change, when a new leader has to restart (for whatever reason) on an empty or non-existing acl file.

src/managers/cluster.py
skourta: Yes that is what I am asking. The 2nd approach has a slight issue if the `ACL LOAD` fails temporarily and we update the password on disk we might lose contact with the cluster when the network is reestablished. We can get around it by only updating the app's secret when the load succeeds but that does not protect us from a charm-independent service restart.

That is my main concern. Let's discuss this in our 1-1 later today.

src/common/client.py
skourta: That is true. I will sanitize it.

pyproject.toml
skourta: Yes, that makes sense. Whether upstream or move the repo to Canonical we need to update this in the near future.

src/events/base_events.py
skourta: Yes, I started implementing the scale up in a different branch. Do you think we should handle that on this PR instead (Only ACL I mean)?

tests/unit/test_charm.py
skourta: It is, yes. I seem to have missed this one.

src/events/base_events.py
skourta: As discussed I left a TODO comment. This will be handled along with updating password for other units as part of the scale up PR.

config.yaml
```diff
@@ -0,0 +1,11 @@
...
+# See LICENSE file for licensing details.
+
+options:
+  system-users:
+    type: secret
+    description: |
+      Configure the internal system user and it's password. The password will
+      be auto-generated if this option is not set. It is for internal use only
+      and SHOULD NOT be used by applications. This needs to be a Juju Secret URI pointing
+      to a secret that contains the following content: `root: <password>`.
```
Mehdi-Bendriss: nit: we should keep a blank line at the end of yaml files
```suggestion
      to a secret that contains the following content: `root: <password>`.
      
```

src/common/client.py
```diff
@@ -0,0 +1,75 @@
...
+        password: str,
+        hosts: list[str],
+    ):
+        self.hosts = hosts
+        self.user = username
+        self.password = password
+
+    async def create_client(self) -> GlideClient:
+        """Initialize the Valkey client."""
+        addresses = [NodeAddress(host=host, port=CLIENT_PORT) for host in self.hosts]
```
Mehdi-Bendriss, line 41: I like there is a list of addresses to choose from, that should hopefully reduce the uncaught random connection errors

src/common/client.py
```diff
@@ -0,0 +1,75 @@
...
+        credentials = ServerCredentials(username=self.user, password=self.password)
+        client_config = GlideClientConfiguration(
```
Mehdi-Bendriss, line 43: Why not use the sync glide client? that should avoid us from having `async / await` all over the place

src/common/client.py
```diff
@@ -0,0 +1,75 @@
...
+
+        Returns:
+            Any result from the command.
+        """
+        client = None
+        try:
+            client = await self.create_client()
+            result = await asyncio.wait_for(client.custom_command(command), timeout=5)
+            return result
+        except Exception as e:
```
Mehdi-Bendriss, line 65: can we be a more precise here as to the expected list of exceptions? or are those different per command?

src/common/client.py
```diff
@@ -0,0 +1,75 @@
...
+            logger.error("Error running custom command: %s", e)
+            raise ValkeyUserManagementError(f"Could not run custom command: {e}")
```
Mehdi-Bendriss: Why raise a UserManagement error here? if we are considering `_run_custom_command` to be a central command, we should rather raise something more specific here

src/common/client.py
```diff
@@ -0,0 +1,75 @@
...
+        finally:
+            if client:
+                await client.close()
+
+    def load_acl(self) -> None:
```
Mehdi-Bendriss: nit for clarity: 
```suggestion
    def reload_acl_rules(self) -> None:
```

src/common/client.py
```diff
@@ -0,0 +1,75 @@
...
+        """Load ACL content to the Valkey server."""
+        try:
+            result = asyncio.run(self._run_custom_command(["ACL", "LOAD"]))
+            logger.debug(f"ACL load result: {result}")
+        except Exception as e:
+            logger.error(f"Error loading ACL: {e}")
+            raise ValkeyUserManagementError(f"Could not load ACL: {e}")
```
Mehdi-Bendriss: similar comment

src/events/base_events.py
```diff
@@ -38,3 +44,109 @@ def _on_update_status(self, event: ops.UpdateStatusEvent) -> None:
...
+
+        if self.charm.unit.is_leader() and not self.charm.state.cluster.internal_user_credentials:
+            if admin_secret_id := self.charm.config.get(INTERNAL_USER_PASSWORD_CONFIG):
+                try:
+                    password = self.charm.state.get_secret_from_id(str(admin_secret_id)).get(
+                        INTERNAL_USER
+                    )
+                except (ops.ModelError, ops.SecretNotFoundError) as e:
+                    logger.error(f"Could not access secret {admin_secret_id}: {e}")
+                    raise
```
Mehdi-Bendriss, line 63: i'm not a fan of crashing hooks, we can address it in another PR, but we should look at a deferral + blocked status here

reneradoi (approved):
Thank you Smail!

src/events/base_events.py
```diff
@@ -38,3 +44,109 @@ def _on_update_status(self, event: ops.UpdateStatusEvent) -> None:
...
+            else:
+                password = self.charm.config_manager.generate_password()
+
+            self.charm.state.cluster.update({"charmed_operator_password": password})
+            self.charm.config_manager.set_acl_file()
+
+    def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
+        """Handle the config_changed event."""
+        self.charm.state.unit_server.update({"hostname": socket.gethostname()})
```
Mehdi-Bendriss, line 72: I believe the handling of VM code in this case is a bit of a low hanging fruit, we can afford to add the ip address as well

Mehdi-Bendriss (commented):
Thanks Smail! Looks good! I have left a few comments / questions

src/common/client.py
skourta: I wondered the same thing myself. René let me know it is just a wrapper for the async client and is not feature-full at the moment. I decided to resolve all asyncs in the `client.py` file so we abstract it away in case we change later on.

src/common/client.py
skourta: I went through the code of the client a little and it didn't specify so I decided to catch all errors from the client and rase a `UserManagementError` just like we do in etcd.

src/common/client.py
skourta: I tried to follow the valkey nomenclature but I am open to renaming this. WDYT?

src/common/client.py
skourta: I changed the exceptions to have a base `ValkeyClientError` that we can use in events to catch all Valkey client related error and then add more specific errors that we can use in the managers if we have to.

src/events/base_events.py
skourta: Yes, adding the IP is actually shared code between the workloads. I added it in the scale up branch as part of the `base_workload`. Should we add it in this PR instead?

src/events/base_events.py
skourta: I left a TODO comment

src/common/client.py
reneradoi: See [here](https://glide.valkey.io/how-to/synchronous-connection/). The project is continuously working on the sync client, but it is a work-in-progress. The way Smail has set it up, it should be relatively easy for us to switch the implementation to the sync client at some point if we decide to.

reneradoi (approved):

src/common/client.py
skourta: I have added the TODO as we discussed.

src/common/client.py
skourta: Renamed

Mehdi-Bendriss (approved):
Thanks Smail!

PR #5, opened 2026-01-28
[DPE-9135] - Create all charm users to interact with Valkey

skourta:
This pull request introduces a refactor to support multiple internal users (rather than a single admin user) for Valkey charm authentication and management. It replaces previous single-user logic with a more flexible, extensible approach using an enumeration of internal users and a mapping of their roles and passwords. The changes touch core models, managers, event handling, literals, and integration tests to ensure consistent multi-user support throughout the codebase.

### Multi-User Support and Model Refactor

* Introduced the `CharmUsers` enum to represent all internal users, and replaced the single `INTERNAL_USER` logic with multi-user handling throughout the codebase. This includes updating literals, models, and role mappings.
* Refactored the `PeerAppModel` to store passwords for all internal users, using a new `InternalUsersSecret` type for each.
* Updated the property and related logic from `internal_user_credentials` to `internal_users_credentials`, returning a dictionary of all internal user passwords.

### Event Handling and Secret Management

* Updated leader election, config change, and secret change event handlers to manage passwords for all internal users, including handling user-specified secrets, generating passwords as needed, and updating the cluster state accordingly. [[1]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL54-R95) [[2]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL77-R223)
* Improved error handling and status reporting for secret access and password update failures, including new status messages and scope corrections. [[1]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL77-R223) [[2]](diffhunk://#diff-074beea58df119261178a25e474cd0c240fa20f1c6fb6990867d357676e96cefL34-R34)

### ACL and Configuration Management

* Refactored ACL file generation to include all relevant internal users, using the new `CHARM_USERS_ROLE_MAP` and `CharmUsers` enum. The ACL file now correctly reflects all internal users and their roles.
* Updated configuration and cluster managers to use the new multi-user model for authentication and password management. [[1]](diffhunk://#diff-20b77aa4083c3162fac7cd352e1b8ceef186a967a37f906277cb77346967c113L17-R17) [[2]](diffhunk://#diff-20b77aa4083c3162fac7cd352e1b8ceef186a967a37f906277cb77346967c113L32-R35) [[3]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cL19-R19)

### Integration Test Updates

* Updated integration test helpers and test cases to use the new internal user model, including secret naming, password retrieval, and client creation. [[1]](diffhunk://#diff-7a77f865f0501e99ea65d27c5459322d56f8093d0f68a51b6cd9f51a73a7c8aaL17-R33) [[2]](diffhunk://#diff-7a77f865f0501e99ea65d27c5459322d56f8093d0f68a51b6cd9f51a73a7c8aaL179-R190) [[3]](diffhunk://#diff-7a77f865f0501e99ea65d27c5459322d56f8093d0f68a51b6cd9f51a73a7c8aaL206-R217) [[4]](diffhunk://#diff-7a77f865f0501e99ea65d27c5459322d56f8093d0f68a51b6cd9f51a73a7c8aaL231-R242) [[5]](diffhunk://#diff-642cf7dd7aa6642e06e3a0c51708ea494904fde750d3eea6a10a938850081d88L10-R19) [[6]](diffhunk://#diff-642cf7dd7aa6642e06e3a0c51708ea494904fde750d3eea6a10a938850081d88L61-R67)

These changes lay the groundwork for more robust and scalable internal user management within the Valkey charm, improving security and maintainability.


tests/integration/k8s/test_charm.py
```diff
@@ -126,7 +169,7 @@ async def test_user_secret_permissions(juju: jubilant.Juju) -> None:
     # deferred `config_changed` event will be retried before `update_status`
     with fast_forward(juju):
         juju.grant_secret(identifier=secret_name, app=APP_NAME)
-        sleep(10)  # allow some time for the permission to propagate
+        sleep(20)  # allow some time for the permission to propagate
```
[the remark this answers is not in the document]
skourta: Experimentally 10s didn't always generate an `update_status` event

src/events/base_events.py
```diff
@@ -51,20 +56,43 @@ def _on_leader_elected(self, event: ops.LeaderElectedEvent) -> None:
...
-            else:
-                password = self.charm.config_manager.generate_password()
+                    self.charm.status.set_running_status(
+                        CharmStatuses.SECRET_ACCESS_ERROR.value,
+                        scope="app",
+                        component_name=self.charm.cluster_manager.name,
+                        statuses_state=self.charm.state.statuses,
+                    )
+                    event.defer()
+                    return
```
reneradoi: If we don't have access to the secret and defer, we might end up in a situation where we start the database (on another event) without having created auth config before.

Previously, the code prevented that by raising. If we don't want to raise anymore, we might need to apply auto-generated passwords instead if we can't access the secret. This is to avoid ever starting the database without proper authentication set up.

src/events/base_events.py
```diff
@@ -74,63 +102,56 @@ def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
...
+                self._update_internal_users_password(str(admin_secret_id))
+            except (ops.ModelError, ops.SecretNotFoundError, ValkeyACLLoadError):
                 event.defer()
                 return
 
     def _on_secret_changed(self, event: ops.SecretChangedEvent) -> None:
         """Handle the secret_changed event."""
-        # TODO For a multi-node cluster the units should independently update their passwords.
-        # If they fail the event should be deferred and retried.
         if not self.charm.unit.is_leader():
```
reneradoi: Let's invert the logic for better readability:

```
if leader and admin_secret:id:
    <logic as is>
    return

non-leader code here
```

src/events/base_events.py
```diff
@@ -142,12 +163,64 @@ def update_admin_password(self, admin_secret_id: str) -> None:
...
-            ClusterStatuses.PASSWORD_UPDATE_FAILED.value,
+            CharmStatuses.SECRET_ACCESS_ERROR.value,
             scope="app",
             component=self.charm.cluster_manager.name,
         )
+
+        # Check which passwords have changed
+        old_passwords = self.charm.state.cluster.internal_users_credentials
+        passwords = {user.value: old_passwords.get(user.value, "") for user in CharmUsers}
+
```
reneradoi: It should be possible to simplify this code a lot. We could do something like:
```
if any(key not in CharmUsers for key in secret_content.keys()):
    <error handling for invalid usersname>
    return

if any(
    secret_content.get(user.value) != state.internal_users_credentials.get(user.value) 
    for user in CharmUsers
):
    logger.info("passwords changed")
    try:
        config_manager.set_acl_file
        cluster_manager.reload_acl_file
        state.cluster.update
    except ...
```

src/events/base_events.py
```diff
@@ -142,12 +163,64 @@ def update_admin_password(self, admin_secret_id: str) -> None:
...
+                    component_name=self.charm.cluster_manager.name,
+                    statuses_state=self.charm.state.statuses,
+                )
+                return
+
+        for user in CharmUsers:
+            new_password = secret_content.get(user.value)
+            if not new_password:
+                continue
+            # only update user credentials if the password has changed
```
reneradoi: This case should be covered by juju cli, I don't think it's possible to set a secret key to an empty value.

src/events/base_events.py
```diff
@@ -142,12 +163,64 @@ def update_admin_password(self, admin_secret_id: str) -> None:
...
+            try:
+                self.charm.config_manager.set_acl_file(passwords=passwords)
+                self.charm.cluster_manager.reload_acl_file()
+                self.charm.state.cluster.update(
+                    {
+                        f"{user.value.replace('-', '_')}_password": passwords[user.value]
+                        for user in CharmUsers
+                    }
+                )
+            except ValkeyACLLoadError as e:
```
reneradoi: We also have to catch the `ValueError` that is raised in `_get_user_acl_line` in the `ConfigManager`.

src/literals.py
```diff
@@ -14,7 +16,31 @@
...
+    VALKEY_SENTINEL = "charmed-sentinel-valkey"
+    VALKEY_REPLICA = "charmed-replication"
+    VALKEY_MONITORING = "charmed-stats"
+
+    # Sentinel users
+    SENTINEL_ADMIN = "charmed-sentinel-peers"
+    SENTINEL_CHARM_ADMIN = "charmed-sentinel-operator"
+
+
+CHARM_USERS_ROLE_MAP = {
```
reneradoi, line 39: I really like this way of defining the permissions!

tests/integration/k8s/test_charm.py
```diff
@@ -82,25 +87,61 @@ async def test_update_admin_password(juju: jubilant.Juju) -> None:
...
     juju.wait(lambda status: jubilant.all_agents_idle(status, APP_NAME), timeout=1200)
 
     # make sure we can still read data with the previously set password
     assert await get_key(
-        hostnames=hostnames, username=INTERNAL_USER, password=new_password, key=TEST_KEY
+        hostnames=hostnames,
+        username=CharmUsers.VALKEY_ADMIN.value,
+        password=new_password,
+        key=TEST_KEY,
     ) == bytes(TEST_VALUE, "utf-8")
```
reneradoi, line 129: It would be great to see a test where we update multiple passwords at the same time (which is something to be expected). We would have to adjust the `set_password` helper for that to accept a dict of `username: password` records.

reneradoi (commented):
Hey Smail! I've identified a few places where we can improve the code for better readability and a simpler solution. Would be happy to hear what you think about the suggestions.

Other than that, we should have test coverage for updating multiple passwords at the same time, while keeping those that have not been updated.

src/events/base_events.py
reneradoi: raising again to ensure the database does not get started without valid authentication.

src/events/base_events.py
reneradoi: structure was inverted for readability

src/events/base_events.py
reneradoi: simplified the logic

tests/integration/k8s/test_charm.py
reneradoi: Confirmed that it works by testing manually.

src/managers/config.py
```diff
@@ -76,27 +76,41 @@ def set_config_properties(self) -> None:
...
-        acl_content += f"user {INTERNAL_USER} on #{charmed_operator_password_hash} ~* +@all\n"
+        for user in CharmUsers:
+            # only process VALKEY users
+            # Sentinel users should be in the sentinel acl file
+            if "VALKEY_" not in str(user):
+                continue
+            acl_content += self._get_user_acl_line(user, passwords=passwords)
         self.workload.write_file(acl_content, ACL_FILE)
 
+    def _get_user_acl_line(self, user: CharmUsers, passwords: dict[str, str] | None = None) -> str:
```
Mehdi-Bendriss, line 96: nice

src/managers/config.py
```diff
@@ -76,27 +76,41 @@ def set_config_properties(self) -> None:
...
+
+        Returns:
+            str: ACL line for the user.
+        """
+        passwords = passwords or self.state.cluster.internal_users_credentials
+        if not (password := passwords.get(user.value, "")):
+            raise ValueError(f"No password found for user {user}")
+        password_hash = hashlib.sha256(password.encode("utf-8")).hexdigest()
+        acl_line = f"user {user.value} on #{password_hash} {CHARM_USERS_ROLE_MAP[user]}\n"
+        return acl_line
```
Mehdi-Bendriss: nit:
```suggestion
        return f"user {user.value} on #{password_hash} {CHARM_USERS_ROLE_MAP[user]}\n"
```

src/core/models.py
```diff
@@ -127,9 +137,13 @@ def model(self) -> PeerAppModel | None:
         return self.data_interface.build_model(self.relation.id) if self.relation else None
 
     @property
-    def internal_user_credentials(self) -> dict[str, str]:
+    def internal_users_credentials(self) -> dict[str, str]:
         """Retrieve the credentials for the internal admin user."""
```
Mehdi-Bendriss: docstrings comment to be updated

src/events/base_events.py
```diff
@@ -51,20 +56,42 @@ def _on_leader_elected(self, event: ops.LeaderElectedEvent) -> None:
             event.defer()
             return
 
-        if self.charm.unit.is_leader() and not self.charm.state.cluster.internal_user_credentials:
-            if admin_secret_id := self.charm.config.get(INTERNAL_USER_PASSWORD_CONFIG):
+        if self.charm.unit.is_leader() and not self.charm.state.cluster.internal_users_credentials:
```
Mehdi-Bendriss: could we invert the condition and return ?

src/events/base_events.py
```diff
@@ -51,20 +56,42 @@ def _on_leader_elected(self, event: ops.LeaderElectedEvent) -> None:
...
-                # TODO consider deferring and blocking the charm
                 except (ops.ModelError, ops.SecretNotFoundError) as e:
                     logger.error(f"Could not access secret {admin_secret_id}: {e}")
+                    self.charm.status.set_running_status(
+                        CharmStatuses.SECRET_ACCESS_ERROR.value,
+                        scope="app",
+                        component_name=self.charm.cluster_manager.name,
+                        statuses_state=self.charm.state.statuses,
+                    )
                     raise
```
Mehdi-Bendriss: Should we defer/return? the running status above will never show in this case 
In general, we should try to avoid crashing hooks, though in the case of the first leader-elected of the deployment if we don't have an admin user, there isn't much the charm can do..

src/events/base_events.py
```diff
@@ -74,63 +101,58 @@ def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
...
                     statuses_state=self.charm.state.statuses,
                 )
+                event.defer()
                 return
+            self.charm.state.statuses.delete(
+                ClusterStatuses.PASSWORD_UPDATE_FAILED.value,
+                scope="unit",
+                component=self.charm.cluster_manager.name,
+            )
+        return
```
Mehdi-Bendriss:
```suggestion
```

src/events/base_events.py
```diff
@@ -51,20 +56,42 @@ def _on_leader_elected(self, event: ops.LeaderElectedEvent) -> None:
...
-                password = self.charm.config_manager.generate_password()
 
-            self.charm.state.cluster.update({"charmed_operator_password": password})
+                self.charm.state.statuses.delete(
+                    CharmStatuses.SECRET_ACCESS_ERROR.value,
+                    scope="app",
+                    component=self.charm.cluster_manager.name,
+                )
+
+            # generate passwords for all internal users if not specified in the user secret
```
Mehdi-Bendriss: unlike in `_update_internal_users_password` we do not perform a validation check on the keys for the leader, or is this delegated fully config_changed?

tests/integration/k8s/test_charm.py
```diff
@@ -82,25 +87,61 @@ async def test_update_admin_password(juju: jubilant.Juju) -> None:
...
+
+    # wait for config-changed hook to finish executing
+    juju.wait(
+        lambda status: does_status_match(
+            status,
+            expected_app_statuses={APP_NAME: [ClusterStatuses.PASSWORD_UPDATE_FAILED.value]},
+        ),
+        timeout=1200,
+    )
+
```
Mehdi-Bendriss, line 156: could we add a check that all remaining user/passwords remain unchanged after changing the admin password?

Mehdi-Bendriss (commented):
Thanks Smail and René! Great work! I left a few comments / questions

src/events/base_events.py
reneradoi: Please see the previous comment from me ([link](https://github.com/canonical/valkey-operator/pull/5#discussion_r2745855047)) on that.

I've removed the status, as this indeed will never be shown.

src/events/base_events.py
reneradoi: I assume Smail's initial idea was to delegate and not block this events flow.

tests/integration/k8s/test_charm.py
reneradoi: updated

src/events/base_events.py
skourta: Yes, that was my intention.

src/managers/config.py
reneradoi: updated

src/core/models.py
reneradoi: done

Mehdi-Bendriss (approved):

reneradoi (approved):

PR #7, opened 2026-02-03
[DPE-9168] Support K8s and VM deployments

reneradoi:
This PR adds support for both K8s and VM substrates to the operator.

The main changes are:
- based on the result of `ops.model.get_cloud_spec()`, the charm recognizes the substrate it is deployed on; the information is added to the cluster state
- the operator instantiates the `ValkeyVmWorkload` class when deployed on VM clouds, while instantiating the `ValkeyK8sWorkload` class on Kubernetes clouds
- on VM, the operator installs the `charmed-valkey` snap and adjusts configuration to specific paths
- the deployment requires `--trust`
- when deployed on VM clouds, the image resource does not have to be specified
- startup of the database is moved from `pebble_ready` to the `start` event to account for both substrates
- furthermore, an integration test suite for VM deployments is added


src/literals.py
```diff
@@ -44,3 +52,10 @@ class CharmUsers(str, Enum):
     CharmUsers.SENTINEL_ADMIN: "~* +@all",
     CharmUsers.SENTINEL_CHARM_ADMIN: "~* +@all",
 }
+
+
+class Substrate(StrEnum):
```
Mehdi-Bendriss, line 58: can you change the `CharmUsers` enum above to inherit from `StrEnum`? we need to be consistent

src/managers/config.py
```diff
@@ -60,14 +69,16 @@ def config_properties(self) -> dict[str, str]:
                 config_properties[key.strip()] = value.strip()
 
         # Adjust default values
-        # port
-        config_properties["port"] = str(CLIENT_PORT)
-
-        # bind to all interfaces
         config_properties["bind"] = "0.0.0.0 -::1"
```
Mehdi-Bendriss: This needs to be the juju bind address on VM

src/managers/config.py
```diff
@@ -60,14 +69,16 @@ def config_properties(self) -> dict[str, str]:
...
+        config_properties["port"] = str(CLIENT_PORT)
+        config_properties["loglevel"] = "verbose"
 
-        # Use the ACL file
-        config_properties["aclfile"] = str(ACL_FILE)
+        if self.state.substrate == Substrate.VM:
+            config_properties["dir"] = f"{SNAP_COMMON_PATH}/var/lib/charmed-valkey"
+            config_properties["aclfile"] = f"{SNAP_CURRENT_PATH}/{SNAP_ACL_FILE}"
+        else:
+            config_properties["dir"] = "/var/lib/valkey"
```
Mehdi-Bendriss: this is okay for the time being, but we need to remember to rebuild the rock based on the latest snap and use `/charmed-valkey`

src/workload_vm.py
```diff
@@ -0,0 +1,110 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ValkeyVmWorkload(WorkloadBase):
+    """Implementation of WorkloadBase for running a Valkey container on VM."""
+
+    def __init__(self) -> None:
+        for attempt in Retrying(stop=stop_after_attempt(5), wait=wait_fixed(5)):
+            with attempt:
+                self.valkey = snap.SnapCache()[SNAP_NAME]
```
Mehdi-Bendriss, line 35: i saw this in opensearch too, but I can't remember of a good reason to have it in the constructor.. we already have an install method, why do it here too?

src/workload_vm.py
```diff
@@ -0,0 +1,110 @@
...
+
+        self.config_file = pathops.LocalPath(f"{SNAP_CURRENT_PATH}/{SNAP_CONFIG_FILE}")
+
+    @property
+    @override
+    def can_connect(self) -> bool:
```
Mehdi-Bendriss, line 44: missing docstrings

src/workload_vm.py
```diff
@@ -0,0 +1,110 @@
...
+            self.valkey.hold()
+            return True
+        except snap.SnapError as e:
+            logger.error(str(e))
+            if retry_and_raise:
+                raise RuntimeError
+            return False
+
+    @override
+    def start(self) -> None:
```
Mehdi-Bendriss, line 84: missing docstrings (applies to various methods in this module)

src/workload_vm.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @override
+    def write_config_file(self, config: dict[str, str]) -> None:
+        config_string = "\n".join(f"{str(key)}{' '}{str(value)}" for key, value in config.items())
+
+        path = self.config_file
+        path.write_text(config_string)
+
+    @override
+    def write_file(self, content: str, path: str) -> None:
```
Mehdi-Bendriss: I think these 2 methods can go to the base workload class, given we're using pathops

tests/integration/helpers.py
```diff
@@ -172,6 +172,11 @@ def get_cluster_hostnames(juju: jubilant.Juju, app_name: str) -> list[str]:
         A list of hostnames for all units in the Valkey application.
     """
     status = juju.status()
+    model_info = juju.show_model()
+
+    if model_info.type == "lxd":
```
Mehdi-Bendriss: Let's rather test on `!= "k8s"`, we will be able to run our integration tests against an existing non-lxd cloud deployment

tests/integration/vm/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+from platform import machine
+
+import jubilant
+import pytest
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope="package")
+def arch() -> str:
```
Mehdi-Bendriss, line 14: nice

tox.ini
```diff
@@ -69,4 +69,4 @@ commands_pre =
 commands =
     # on CI, concierge will setup the model `testing` - locally we need to do it ourselves
     sh -c "if [ -z "$CI" ]; then juju add-model testing && juju model-config logging-config='<root>=INFO;unit=DEBUG'; fi;"
-    poetry run pytest -v --tb native --log-cli-level=INFO -s --ignore={[vars]tests_path}/unit/ {posargs}
\ No newline at end of file
+    poetry run pytest -v --tb native --log-cli-level=INFO -s -x --ignore={[vars]tests_path}/unit/ {posargs}
```
Mehdi-Bendriss:
```suggestion
    poetry run pytest -v --tb native --log-cli-level=INFO -s -x --ignore={[vars]tests_path}/unit/ {posargs}
    
```

concierge-vm.yaml
Mehdi-Bendriss, line 1: `concierge-vm.yaml` would be a better qualifier

Mehdi-Bendriss (commented):
Thanks René! Good work! I left some comments

src/charm.py
```diff
@@ -24,8 +25,19 @@ class ValkeyCharm(ops.CharmBase):
 
     def __init__(self, *args) -> None:
         super().__init__(*args)
-        self.workload = ValkeyK8sWorkload(container=self.unit.get_container(CONTAINER))
-        self.state = ClusterState(self)
+        try:
+            cloud_spec = self.model.get_cloud_spec()
```
skourta, line 29: Did we decide to go with the cloud spec instead of the ENV VARs?

src/core/base_workload.py
```diff
@@ -39,3 +40,8 @@ def write_file(self, content: str, path: str) -> None:
             path (str): The file path where the content should be written.
         """
         pass
+
+    @abstractmethod
+    def exec(self, command: List[str]) -> str:
```
skourta: We don't need `List` type helper just use the normal `list`

src/literals.py
skourta: +1. I didn't even know it existed

src/managers/config.py
```diff
@@ -60,14 +69,16 @@ def config_properties(self) -> dict[str, str]:
...
-
-        # bind to all interfaces
         config_properties["bind"] = "0.0.0.0 -::1"
+        config_properties["port"] = str(CLIENT_PORT)
+        config_properties["loglevel"] = "verbose"
 
-        # Use the ACL file
-        config_properties["aclfile"] = str(ACL_FILE)
+        if self.state.substrate == Substrate.VM:
+            config_properties["dir"] = f"{SNAP_COMMON_PATH}/var/lib/charmed-valkey"
```
skourta: Why are we using the `SNAP_COMMON` for the config file?

src/workload_k8s.py
```diff
@@ -5,7 +5,7 @@
 """Implementation of WorkloadBase for running Valkey on K8s."""
 
 import logging
-from typing import override
+from typing import List, override
```
skourta: Also no need for List

skourta (commented):
Thank you, @reneradoi . I left small comments and a couple of question.

src/managers/config.py
reneradoi: I don't think this will work. `bind` is the address(es) that Valkey accepts communication from. If this is set to the machines IP address, only connections from localhost will be possible. How are clients supposed to connect to the instances then?

src/managers/config.py
reneradoi: Agree. I've added a todo comment.

src/workload_vm.py
reneradoi: To be able to easily access the snap in various methods, for example `self.valkey.start()` without having to instantiate the `SnapCache` in each of those methods.

src/workload_vm.py
reneradoi: All of these are overrides and the doc strings are in the abstract methods in the base class.

src/workload_vm.py
reneradoi: All of these are overrides and the doc strings are in the abstract methods in the base class.

concierge-vm.yaml
reneradoi: Agree, updated.

src/literals.py
reneradoi: Updated.

src/managers/config.py
reneradoi: `dir` is the working directory, for example the database snapshot files will be stored here. The config file location is not defined by that, this is given as a parameter to start the valkey server in the snap.

src/workload_vm.py
reneradoi: Good idea, have moved them.

src/charm.py
reneradoi: Yes I prefer that. Relying on environment variables that could randomly be set on a VM as well does not sound like a good solution to me.

src/core/base_workload.py
reneradoi: thx, updated.

src/charm.py
skourta: Just confirming. Thank you.

tests/integration/helpers.py
reneradoi: Updated.

src/managers/config.py
```diff
@@ -72,13 +69,13 @@ def config_properties(self) -> dict[str, str]:
         config_properties["bind"] = "0.0.0.0 -::1"
         config_properties["port"] = str(CLIENT_PORT)
         config_properties["loglevel"] = "verbose"
+        config_properties["aclfile"] = self.workload.acl_file.as_posix()
```
skourta, line 70: I like this. Would it be possible to do the same for the `dir` path?

skourta (approved):
One final minor question but LGTM.

src/managers/config.py
Mehdi-Bendriss: This is expected, we already talked about it in MM a couple of weeks back. 
`bind` controls which local network interfaces the server listens on, not which client IPs are allowed.

A few things to keep in mind:
- We think of clients in juju ecosystem, as juju clients (including data-integrator) first, we expect those to have access to the model where the database runs (offer etc..)
- It is up to the operators to handle the networking of their deployment
- Binding to 0.0.0.0 on **vm** is a security risk, that the user must take themselves consciously if they wish so, and protect their units with security groups for ex in public cloud
- We will not allow by default valkey to bind to public addresses out-of-the-box, we will use what juju tells us to bind on ==> `bind_address`. If the user opens up their instance, bind_address should be aware of it 
- We will introduce juju spaces support in the future for more granular control over the bind

reneradoi: Yes I'm sorry I was wrong about the `bind` configuration of Valkey. I've confirmed by testing and am updating the code.

reneradoi: Updated now.

src/managers/config.py
reneradoi: Added - handling it like this makes the code much cleaner.

Mehdi-Bendriss (approved):

PR #8, opened 2026-02-03
[DPE-9174] HA - Scale up

skourta:
This pull request introduces several enhancements and refactors to the Valkey charm's core event handling, workload management, and configuration logic. The main focus is on improving cluster initialization, unit startup sequencing, configuration file management, and internal user credential handling. The changes also add new dependencies and update exception handling for configuration operations.

The most important changes are:

**Cluster Initialization and Unit Startup Sequencing:**
- Added a new `UnitFullyStarted` event to signal when a unit has fully started, and implemented logic to defer unit startup until prerequisites are met (e.g., leader sets primary, internal user credentials, and starting member selection). Improved status handling and event sequencing for cluster formation and replica synchronization. [[1]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL13-R44) [[2]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eR54-R61) [[3]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL65-R253)
- Enhanced peer relation event handling to coordinate which unit starts next, using a new `starting_member` field in the cluster model and logic to clear it once the unit has started. [[1]](diffhunk://#diff-e46c88bb687d77b94764a732b3e08a0a7bd544a64e606429c0b787c13e1c226aR39-R49) [[2]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL65-R253)

**Workload and Configuration Management:**
- Refactored `WorkloadBase` to add methods for checking service liveness (`alive`), retrieving the private IP, creating directories with ownership/mode, and improving file writing to support k8s workloads. [[1]](diffhunk://#diff-90d9cf1ebee4e42b159cab54e264b9947faa2f1435648c7857df493fc091d2fdR7-R30) [[2]](diffhunk://#diff-90d9cf1ebee4e42b159cab54e264b9947faa2f1435648c7857df493fc091d2fdL29-R92) [[3]](diffhunk://#diff-90d9cf1ebee4e42b159cab54e264b9947faa2f1435648c7857df493fc091d2fdR123-R141)
- Added a new `set_runtime_config` method to the Valkey client to support dynamic configuration changes, and introduced a custom exception for configuration set failures. [[1]](diffhunk://#diff-f5c8dd5ff5ebd3473ee519d51e59ddb12e43e04365489c1c429256d3d20c3bc4R19) [[2]](diffhunk://#diff-f5c8dd5ff5ebd3473ee519d51e59ddb12e43e04365489c1c429256d3d20c3bc4R81-R97) [[3]](diffhunk://#diff-e793bad618b16c64102391f92909ac368210513be36cc7ff3ea633e458e54feeR19-R22)

**Internal User Credential and Secret Handling:**
- Improved logic for updating and propagating internal user credentials, ensuring the local admin password is updated on all units when secrets change, and refreshing secrets as needed. [[1]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eR311-R312) [[2]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL175-R371) [[3]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL198-R394)
- Extended the peer and unit models to track additional fields such as `primary_ip`, `starting_member`, `private_ip`, and `charmed_operator_password` for better state management. [[1]](diffhunk://#diff-e46c88bb687d77b94764a732b3e08a0a7bd544a64e606429c0b787c13e1c226aR39-R49) [[2]](diffhunk://#diff-e46c88bb687d77b94764a732b3e08a0a7bd544a64e606429c0b787c13e1c226aR123-R129)

**Dependency Updates:**
- Updated `pyproject.toml` to add dependencies on `python-dateutil`, `valkey`, and `tenacity`, and removed the custom `valkey-glide` dependency.

**Miscellaneous Improvements:**
- Added optional `refresh` parameter to secret retrieval for more precise control.
- Improved event logging and error handling throughout the event lifecycle. [[1]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL65-R253) [[2]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL175-R371)

These changes collectively improve the robustness, maintainability, and clarity of the cluster management and event handling logic in the Valkey charm.


src/managers/cluster.py
```diff
@@ -30,22 +35,174 @@ def __init__(self, state: ClusterState, workload: WorkloadBase):
...
         self.workload = workload
         self.admin_user = CharmUsers.VALKEY_ADMIN.value
-        self.admin_password = self.state.cluster.internal_users_credentials.get(
-            CharmUsers.VALKEY_ADMIN.value, ""
-        )
-        self.cluster_hostnames = [server.model.hostname for server in self.state.servers]
+        self.admin_password = self.state.unit_server.valkey_admin_password
+
+    @property
+    def number_units_started(self) -> int:
```
reneradoi: This should go to the state.

src/managers/cluster.py
```diff
@@ -30,22 +35,174 @@ def __init__(self, state: ClusterState, workload: WorkloadBase):
...
+        except ValkeyWorkloadCommandError:
+            raise ValkeyConfigSetError("Could not set primaryauth on Valkey server.")
+
+    @tenacity.retry(
+        wait=tenacity.wait_fixed(5),
+        stop=tenacity.stop_after_attempt(5),
+        retry=tenacity.retry_if_result(lambda result: result is False),
+        reraise=True,
+    )
+    def is_sentinel_discovered(self) -> bool:
```
reneradoi: General design comment: I think we should have a separate `SentinelManager` class to later be able to integrate a non-sentinel mode easily. Otherwise we will have a thousand conditions to differentiate between the modes.

src/managers/cluster.py
```diff
@@ -30,22 +35,174 @@ def __init__(self, state: ClusterState, workload: WorkloadBase):
...
+                return primary_ip
+            except (IndexError, ValkeyWorkloadCommandError):
+                logger.error("Could not get primary IP from sentinel output.")
+
+        logger.error(
+            "Could not determine primary IP from sentinels. Number of started servers: %d.",
+            len(started_servers),
+        )
+
+    def _exec_cli_command(
```
reneradoi: I think we could provide this as part of the `Client` class.

From any manager classes' point of view it should be transparent how the client implementation looks like: we should just pass params (hostname, user, password, certs) and a command and the rest is concern of the client implementation.

By that we could use the cli command method in multiple managers and have a cleaner implementation of the method itself (less if-else-switches).

src/core/models.py
```diff
@@ -36,13 +36,16 @@ class PeerAppModel(PeerModel):
     charmed_stats_password: InternalUsersSecret = Field(default="")
     charmed_sentinel_peers_password: InternalUsersSecret = Field(default="")
     charmed_sentinel_operator_password: InternalUsersSecret = Field(default="")
+    starting_member: str = Field(default="")
 
 
 class PeerUnitModel(PeerModel):
     """Model for the peer unit data."""
 
+    charmed_operator_password: InternalUsersSecret = Field(default="")
```
reneradoi: I was confused when I saw this until I remembered our discussion on the topic. We should change the name here to make clear that this is just a local copy for decoupling the workflows.

src/core/base_workload.py
```diff
@@ -26,19 +42,56 @@ def start(self) -> None:
...
         """Run a command on the workload substrate."""
         pass
 
-    def write_file(self, content: str, path: pathops.PathProtocol) -> None:
+    @abstractmethod
+    def alive(self) -> bool:
+        """Check if the Valkey service is running."""
+        pass
+
+    def get_private_ip(self) -> str:
```
reneradoi: We should be careful with that, given that we target K8s, we can't just `subprocess.run`. I've opted for using the ops property `Ingress_address` instead (https://documentation.ubuntu.com/ops/latest/reference/ops/#ops.Network.ingress_address), which is in line with using `bind_address` for the configuration.

src/core/cluster_state.py
```diff
@@ -112,18 +112,19 @@ def bind_address(self) -> str:
...
             model (Model): Model object.
             secret_id (str): The id of the secret.
+            refresh (bool): Whether to refresh the secret content from the controller. Defaults to False.
 
         Returns:
             dict: The content of the secret.
         """
         try:
-            secret_content = self.charm.model.get_secret(id=secret_id).get_content(refresh=True)
+            secret_content = self.charm.model.get_secret(id=secret_id).get_content(refresh=refresh)
```
reneradoi: What is the reason for not refreshing the secret? I think we should always do, like previously, Otherwise we will get into situations where we don't have deterministic behavior.

src/managers/config.py
```diff
@@ -70,17 +74,33 @@ def config_properties(self) -> dict[str, str]:
...
+        )
+        # replicaof
+        if primary_ip != self.state.unit_server.model.private_ip:
+            # set replicaof
+            logger.debug("Setting replicaof to primary %s", primary_ip)
+            config_properties["replicaof"] = f"{primary_ip} {CLIENT_PORT}"
+        config_properties["primaryuser"] = CharmUsers.VALKEY_REPLICA.value
+        config_properties["primaryauth"] = self.state.cluster.internal_users_credentials.get(
+            CharmUsers.VALKEY_REPLICA.value, ""
+        )
```
reneradoi: Let's move this part out of `get_config_properties` in order to have clean and readable code. Here's a proposal from the TLS branch: https://github.com/canonical/valkey-operator/blob/add-peer-TLS/src/managers/config.py#L78-L80

src/events/base_events.py
```diff
@@ -62,24 +78,202 @@ def _on_start(self, event: ops.StartEvent) -> None:
...
-            logger.warning("Scaling not implemented yet, services not started")
+            if (
+                not self.charm.state.cluster.internal_users_credentials
+                or not self.charm.cluster_manager.number_units_started
+            ):
+                logger.info(
+                    "Non-leader unit waiting for leader to set primary and internal user credentials"
+                )
+                self.charm.status.set_running_status(
+                    ClusterStatuses.WAITING_FOR_PRIMARY_START.value,
```
reneradoi: General comment: We should avoid running statuses wherever possible. In this event handler we have a few instances where a status is set based on information from the state (as in this particular example). This should be moved to the cluster manager's `get_statuses` method to clean up the event handler code here.

In particular this status is confusing because it is also displayed when a unit is added to an already running cluster (when the unit does not have a peer relation model yet).

src/events/base_events.py
```diff
@@ -62,24 +78,202 @@ def _on_start(self, event: ops.StartEvent) -> None:
...
+        )
+
         logger.info("Services started")
         self.charm.state.unit_server.update({"started": True})
 
     def _on_peer_relation_joined(self, event: ops.RelationJoinedEvent) -> None:
         """Handle event received by all units when a new unit joins the cluster relation."""
-        if self.charm.unit.is_leader():
-            logger.info("Unit %s has joined the relation", event.unit.name)
+        if not self.charm.unit.is_leader() or not event.unit:
```
reneradoi: I'm confused, when should that happen?

src/events/base_events.py
```diff
@@ -62,24 +78,202 @@ def _on_start(self, event: ops.StartEvent) -> None:
             return
 
         if not self.charm.unit.is_leader():
```
reneradoi: The `_on_start` method will have to grow with future logic to be added, we should aim at reducing its current size. For example, we could simplify and have a block at the beginning `if leader:` and group all the leader logic here. This should enable us to remove a lot. e.g. the leader doesn't need to check if there is a `primary_ip`, we can directly put the `started` flag, and so on. Then we can return, and all the subsequent non-leader code will be a lot cleaner.

src/events/base_events.py
```diff
@@ -62,24 +78,202 @@ def _on_start(self, event: ops.StartEvent) -> None:
...
+            ValkeyServiceStatuses.SERVICE_NOT_RUNNING.value,
+            scope="unit",
+            component=self.charm.cluster_manager.name,
+        )
+        if self.charm.unit.is_leader():
+            logger.info("Services started")
+            self.charm.state.unit_server.update({"started": True})
+            return
+
+        self.unit_fully_started.emit()
```
reneradoi: I'm not a big fan of emitting and deferring custom events. Instead I'd propose to "just execute" the sentinel/replica checks after startup with and have a retry with backoff, and as safety net for big amounts of data: synchronize via a flag/timestamp in the unit databag and thereby trigger peer-relation changed events.

src/common/client.py
```diff
@@ -77,3 +78,20 @@ def reload_acl(self) -> None:
         except ValkeyCustomCommandError as e:
             logger.error(f"Error loading ACL: {e}")
             raise ValkeyACLLoadError(f"Could not load ACL: {e}")
+
+    def set_runtime_config(self, config_properties: dict[str, str]) -> None:
```
reneradoi: Method can be removed, it's not used.

src/events/base_events.py
```diff
@@ -62,24 +78,202 @@ def _on_start(self, event: ops.StartEvent) -> None:
...
+
+    # TODO check how to trigger if deferred without update status event
+    def _on_unit_fully_started(self, event: UnitFullyStarted) -> None:
+        """Handle the unit-fully-started event."""
+        self.charm.status.set_running_status(
```
reneradoi: Same here: Let's avoid having all of these running statuses. They pollute the code and are not necessary. We can simplify and check in the `get_statuses` method of the cluster manager if a unit has the started flag and is the current "starting member", and if so, set it's unit status accordingly.

src/managers/config.py
```diff
@@ -116,6 +141,62 @@ def _get_user_acl_line(self, user: CharmUsers, passwords: dict[str, str] | None
         password_hash = hashlib.sha256(password.encode("utf-8")).hexdigest()
         return f"user {user.value} on #{password_hash} {CHARM_USERS_ROLE_MAP[user]}\n"
 
+    def set_sentinel_config_properties(self, primary_ip: str) -> None:
+        """Write sentinel configuration file."""
+        logger.debug("Writing Sentinel configuration")
+
+        sentinel_config = f"port {SENTINEL_PORT}\n"
```
reneradoi: Let's aim at having the sentinel config in a data structure that we can process instead of a string, otherwise we will have a very hard time in the future. 

We can load the config template in the same manner we load it for the `valkey.conf`, which also supports multi-argument configs by splitting a line only after the first whitespace and considering everything that follows as argument for that config parameter. This should also work for the sentinel config, no?

tests/integration/k8s/ha/test_scaling.py
```diff
@@ -0,0 +1,97 @@
...
+
+def test_scale_up(juju: jubilant.Juju, c_writes, c_writes_runner) -> None:
+    """Make sure new units are added to the valkey downtime."""
+    init_units_count = len(juju.status().apps[APP_NAME].units)
+
+    # scale up
+    juju.add_unit(APP_NAME, num_units=2)
+    juju.wait(
+        lambda status: are_apps_active_and_agents_idle(
+            status, APP_NAME, idle_period=10, unit_count=init_units_count + 2
```
reneradoi: We can make use of the `NUM_UNITS` literal from here on.

tests/unit/test_charm.py
```diff
@@ -66,39 +81,45 @@ def test_start_leader_unit(cloud_spec):
...
-        == pebble.ServiceStatus.ACTIVE
-    )
-    assert (
-        state_out.get_container(container.name).service_statuses[SERVICE_METRIC_EXPORTER]
-        == pebble.ServiceStatus.ACTIVE
-    )
-    assert state_out.unit_status == ActiveStatus()
-    assert status_is(state_out, CharmStatuses.SCALING_NOT_IMPLEMENTED.value, is_app=True)
+    with (
+        patch("workload_k8s.ValkeyK8sWorkload.write_file"),
```
reneradoi: We do have an autouse-fixture for mocking the `write_config_file` method and could do the same for `write_file` to avoid patching it all the time.

tests/unit/test_charm.py
```diff
@@ -66,39 +81,45 @@ def test_start_leader_unit(cloud_spec):
...
 
-    state_out = ctx.run(ctx.on.start(), state_in)
-    assert status_is(state_out, CharmStatuses.SERVICE_NOT_STARTED.value)
-    assert status_is(state_out, CharmStatuses.SERVICE_NOT_STARTED.value, is_app=True)
+        state_out = ctx.run(ctx.on.start(), state_in)
+        assert status_is(state_out, CharmStatuses.SERVICE_NOT_STARTED.value)
+        assert status_is(state_out, CharmStatuses.SERVICE_NOT_STARTED.value, is_app=True)
 
 
 def test_start_non_leader_unit(cloud_spec):
```
reneradoi: The test coverage for the start event looks good, but we should also add coverage for the peer-relation changed/joined cases, this is currently missing.

src/events/base_events.py
```diff
@@ -62,24 +78,202 @@ def _on_start(self, event: ops.StartEvent) -> None:
...
+        if (
+            self.charm.state.cluster.model.starting_member
+            and starting_unit
+            and starting_unit.is_started
+        ):
+            logger.debug(
+                "Starting member %s has started. Clearing starting member field.",
+                self.charm.state.cluster.model.starting_member,
+            )
+            self.charm.state.cluster.update({"starting_member": ""})
```
reneradoi, line 252: We might need to safeguard this whole block so that only the leader can update this information as it is the one orchestrating the flow.

src/managers/cluster.py
```diff
@@ -30,22 +35,174 @@ def __init__(self, state: ClusterState, workload: WorkloadBase):
...
+            hostname,
+            "-p",
+            str(port),
+            "--user",
+            user,
+            "--pass",
+            password,
+        ] + command
+        output, error = self.workload.exec(cli_command)
+        logger.debug("Executed command: %s, got output: %s", " ".join(command[0]), output)
```
reneradoi: Logging the output of every command is quite polluting the log, even though I understand it might be helpful for debugging. But especially the returned output from sentinel commands is heavy and I think we should keep this out of the logs.

src/managers/cluster.py
```diff
@@ -30,22 +35,174 @@ def __init__(self, state: ClusterState, workload: WorkloadBase):
...
+        if error:
+            logger.error("Error output from command '%s': %s", " ".join(command[0]), error)
```
reneradoi: This logging keeps on giving, we have massive amounts of sentinel errors in the logs saying: `Error output from command 's e n t i n e l': Warning: Using a password with '-a
' or '-u' option on the command line interface may not be safe.` Maybe we should only log error output when the cli command actually fails?

reneradoi (commented):
Thanks for the big leap here Smail! It already works well when deploying multiple units or adding units to a deployment, that's a nice result! I think we can improve on the code quality for readability and cleanness, and also the orchestration of the startup workflow could be simplified a bit more. But in general looks good!

src/managers/cluster.py
skourta: I agree. I did not want to just put it in the state root but it does make more sense. We can access it directly via `self.charm.state`. I cannot put it in the cluster because we need access to the remote unit data.

src/core/base_workload.py
skourta: Can you please explain why this wouldn't work on k8s? We are running a juju hook cmd.

src/events/base_events.py
skourta: That was my reaction when the hook failed. It seems we get a a first relation joined event without any units then we get consecutive relation joined events for the units joined.

src/core/models.py
skourta: I have added a property to use it `valkey_admin_password` at the unit_server level. I wanted to keep the key used in both unit and app data bags similar. I am open to changing this. LMWYT.

src/core/cluster_state.py
skourta: I just added the argument for cases we do not want to refresh the value to avoid losing the old value of the secret. This is already a sideeffect at the data interfaces v1 library. Currently we always refresh when we use the function.

src/managers/config.py
skourta: No that would not work. Sentinel has commands that have subcommands. I thought about doing it like that but it is very hard to parse the file correctly. It woule be a lot of conditions and not very maintainable in my opinion.

reneradoi: From the `sentinel.conf` file and the [docs](https://valkey.io/topics/sentinel/) I get that the syntax is always either `<parameter> <value(s)>` or `"sentinel" <parameter> <value(s)>`, with `sentinel` being the prefix in case of a subcommand.

Might be that I miss something, but I think this is worth a shot:
```
for line in config:
    if not line or line.startswith("#"):
        # ignore comments and empty lines
        continue
    if line.startswith("sentinel"):
        command, key, value = line.split(" ", 2)
    else:
        key, value = line.split(" ", 1)  
```

src/core/base_workload.py
reneradoi: It's not that it doesn't work, but we shouldn't do it for these reasons:
- we are bypassing `ops` when executing a hook command ourselves (I understand that we did it in previous cases)
- using `bind-address` and `ingress-address` is the recommended approach from Juju side ([link](https://documentation.ubuntu.com/juju/3.6/reference/hook-command/list-of-hook-commands/network-get/#details))
- I see [signs](https://documentation.ubuntu.com/juju/3.6/howto/manage-your-juju-deployment/upgrade-your-juju-deployment-from-36-to-40/#private-address-is-no-longer-auto-maintained-in-relation-data) that `private-address` might be disappearing soon and I think we should not bet on it anymore

src/core/models.py
reneradoi: That's understandable. What's your view on this @Mehdi-Bendriss?

src/events/base_events.py
reneradoi: Interesting! From the [docs](https://documentation.ubuntu.com/ops/latest/reference/ops/#ops.RelationJoinedEvent) `unit` might be `None`, didn't know that.

src/events/base_events.py
skourta: If we do not use a custom event we will have to either defer the `start` event which is what I wanted to avoid or find another place to put the checks. If the start event finishes and the checks are not satisfied yet we need to continue chekcing until they do. If we do decide to go with relation_changed/timestamp approach then we can potentially move the checks there. I still prefer the custom event because in my opinion it groups the checks and their goal nicely.

src/common/client.py
skourta: I cleaned all the file since we were not using the client.

src/managers/config.py
skourta: I do not see any added benefit since we only set the values we want and the rest will be the default values and we do it in a reconciliation approach. Not setting everything also makes it easier to read the file on the containers and reason about. I do not have a hard opinion on this though and can move to this approach if prefered.

src/core/base_workload.py
reneradoi: I've reconfirmed this with the Juju team: We should stay away from `unit-get` and use the results of `network-get` instead (which `ops` provides for us with `bind-address` and `ingress-address`). `unit-get` does not have any knowledge of spaces and cross-model relations and is a candidate for being deprecated on the next major release of Juju.

skourta: Switched to `bind-address` and removed the `get_private_ip` from workloads.

src/common/client.py
```diff
@@ -30,50 +19,44 @@ def __init__(
...
-            logger.debug(f"ACL load result: {result}")
-        except ValkeyCustomCommandError as e:
-            logger.error(f"Error loading ACL: {e}")
-            raise ValkeyACLLoadError(f"Could not load ACL: {e}")
+        Raises:
+            ValkeyWorkloadCommandError: If the CLI command fails to execute.
+        """
+        port = CLIENT_PORT if self.connect_to == "valkey" else SENTINEL_PORT
+        user = self.username
+        password = self.password
```
Mehdi-Bendriss: I don't think these 2 variables are necessary to declare

src/core/base_workload.py
```diff
@@ -4,16 +4,30 @@
...
 
+logger = logging.getLogger(__name__)
+
 
 class WorkloadBase(ABC):
     """Base interface for common workload operations."""
 
+    def __init__(self) -> None:
+        """Initialize the WorkloadBase."""
+        self.root: pathops.PathProtocol
```
Mehdi-Bendriss: nit, for clarity, could we have `root_dir`

src/core/base_workload.py
```diff
@@ -4,16 +4,30 @@
...
+        self.config_file: pathops.PathProtocol
+        self.sentinel_config: pathops.PathProtocol
```
Mehdi-Bendriss: for consistency with the rest of the variables: `sentinel_config_file`

src/core/base_workload.py
```diff
@@ -4,16 +4,30 @@
...
+        self.acl_file: pathops.PathProtocol
+        self.sentinel_acl_file: pathops.PathProtocol
+        self.working_dir: pathops.PathProtocol
+        self.cli: str
```
Mehdi-Bendriss: Could you explain what `cli` refer to?

src/core/models.py
Mehdi-Bendriss: I also agree with the need to clarify that this is a local copy, I suspect this will be abused by mistake by future contributors (or at least something they'll scratch their head over for some time)
What do you think of: `charmed_operator_password_replica`? or `charmed_operator_password_unit_copy`? I'm struggling with a good name on this one

src/events/base_events.py
```diff
@@ -10,38 +10,52 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+class UnitFullyStarted(ops.EventBase):
+    """Event that signals that the unit's has fully started.
+
+    This event will be deferred until:
+        The Sentinel service is running and was discovered by other units.
+        The Valkey service is running and the replica has finished syncing data.
```
Mehdi-Bendriss: Perhaps:
```suggestion
        The Valkey service is running and the current node is in sync with the primary (if a replica).
```

src/events/base_events.py
```diff
@@ -60,26 +74,151 @@ def _on_start(self, event: ops.StartEvent) -> None:
             logger.warning("Workload not ready yet")
             event.defer()
             return
+        self.charm.state.unit_server.update({"start_state": StartState.NOT_STARTED.value})
 
-        if not self.charm.unit.is_leader():
-            logger.warning("Scaling not implemented yet, services not started")
+        if self.charm.unit.is_leader():
+            self._start_services(event, primary_ip=self.charm.state.bind_address)
```
Mehdi-Bendriss: Why do we assume that the primary is the juju leader? this would only be guaranteed in an initial healthy deployment where we control the flow (and maybe on a host reboot). But it won't be the case in a scale up where the new unit suddenly assumes juju leadership, right? 

We should test on the existence of a primary address

src/events/base_events.py
```diff
@@ -60,26 +74,151 @@ def _on_start(self, event: ops.StartEvent) -> None:
...
+        ):
+            logger.info(
+                "Non-leader unit waiting for leader to set primary and internal user credentials"
+            )
+            event.defer()
             return
 
+        self.charm.state.unit_server.update({"request_start_lock": True})
+
+        if self.charm.state.cluster.model.starting_member != self.charm.unit.name:
```
Mehdi-Bendriss, line 117: I will leave a note here, so we start thinking about it from now, the vanilla unit.name will not scale in the case of sharding, as we do not have the guarantee of "app name uniqueness" cross models. 
There needs to be a combination of "unit name + model uuid" (ref opensearch `unit_name`)

src/events/base_events.py
```diff
@@ -60,26 +74,151 @@ def _on_start(self, event: ops.StartEvent) -> None:
...
+            logger.info("Non-leader unit waiting for leader to choose it as starting member")
+            event.defer()
+            return
+
+        if not (primary_ip := (self.charm.sentinel_manager.get_primary_ip())):
```
Mehdi-Bendriss:
```suggestion
        if not (primary_ip := self.charm.sentinel_manager.get_primary_ip()):
```

Mehdi-Bendriss: why is this check not happening earlier in the  block:
```
if (
    not self.charm.state.cluster.internal_users_credentials
    or not self.charm.state.number_units_started
):
```

src/events/base_events.py
```diff
@@ -60,26 +74,151 @@ def _on_start(self, event: ops.StartEvent) -> None:
...
+            return
+
+        self._start_services(event, primary_ip=primary_ip)
+        self.unit_fully_started.emit()
+
+    def _start_services(self, event: ops.StartEvent, primary_ip: str) -> None:
+        """Start Valkey and Sentinel services."""
         try:
-            self.charm.config_manager.set_config_properties()
+            self.charm.config_manager.update_local_valkey_admin()
```
Mehdi-Bendriss:
```suggestion
            self.charm.config_manager.update_local_valkey_admin_password()
```

Mehdi-Bendriss: Can you explain the flow when the local valkey admin password doesn't happen due to the fact that the app password is missing?

src/events/base_events.py
```diff
@@ -60,26 +74,151 @@ def _on_start(self, event: ops.StartEvent) -> None:
...
+            event.defer()
+            return
+
+        if not (primary_ip := (self.charm.sentinel_manager.get_primary_ip())):
+            logger.error("Primary IP not found. Deferring start event.")
+            event.defer()
+            return
+
+        self._start_services(event, primary_ip=primary_ip)
+        self.unit_fully_started.emit()
```
Mehdi-Bendriss: This is missing a condition / check on when a `CONFIGURATION_ERROR` happens

src/events/base_events.py
```diff
@@ -60,26 +74,151 @@ def _on_start(self, event: ops.StartEvent) -> None:
...
+        else:
+            logger.error("Workload failed to start.")
+            self.charm.status.set_running_status(
+                ValkeyServiceStatuses.SERVICE_NOT_RUNNING.value,
+                scope="unit",
+                component_name=self.charm.cluster_manager.name,
+                statuses_state=self.charm.state.statuses,
+            )
+
+        self.charm.state.statuses.delete(
```
Mehdi-Bendriss: this will just override the previous running status set on the `else` since there is no `return` in that block.
Let's invert the previous conditions and leave in the case `not self.charm.workload.alive()`

src/events/base_events.py
```diff
@@ -60,26 +74,151 @@ def _on_start(self, event: ops.StartEvent) -> None:
...
+        )
+        self.charm.status.set_running_status(
+            ValkeyServiceStatuses.SERVICE_STARTING.value,
+            scope="unit",
+            component_name=self.charm.cluster_manager.name,
+            statuses_state=self.charm.state.statuses,
+        )
 
         self.charm.workload.start()
+        if self.charm.workload.alive():
```
Mehdi-Bendriss: I think checking on the `active` of the service is not enough as it only checks service health, not underline application health. This will create flakiness in the start process. We need to have a couple of more additions:
- a retry mechanism on `workload.start()` 
- a check on the reachability of valkey and sentinel

src/events/base_events.py
```diff
@@ -60,26 +74,151 @@ def _on_start(self, event: ops.StartEvent) -> None:
...
+        ]
+        starting_unit = next(
+            (
+                unit
+                for unit in self.charm.state.servers
+                if unit.unit_name == self.charm.state.cluster.model.starting_member
+            ),
+            None,
+        )
+        if (
```
Mehdi-Bendriss, line 233: please follow the work by Patricia on the more robust rolling ops, we may be able to get rid of quite some code

src/events/base_events.py
```diff
@@ -122,6 +268,8 @@ def _on_leader_elected(self, event: ops.LeaderElectedEvent) -> None:
                 for user in CharmUsers
             }
         )
+        # update local unit admin password
+        self.charm.config_manager.update_local_valkey_admin()
```
Mehdi-Bendriss: shouldn't this run in all units? the current only runs in the leader

src/managers/config-template/sentinel.conf
```diff
@@ -0,0 +1,361 @@
...
+
+# When primary-reboot-down-after-period is set to 0, Sentinel does not fail over
+# when receiving a -LOADING response from a primary. This was the only supported
+# behavior before Redis OSS 7.0.
+#
+# Otherwise, Sentinel will use this value as the time (in ms) it is willing to
+# accept a -LOADING response after a primary has been rebooted, before failing
+# over.
+
+SENTINEL primary-reboot-down-after-period mymaster 0
```
Mehdi-Bendriss:
```suggestion
SENTINEL primary-reboot-down-after-period mymaster 0

```

src/managers/cluster.py
```diff
@@ -30,22 +35,77 @@ def __init__(self, state: ClusterState, workload: WorkloadBase):
...
-        except ValkeyACLLoadError:
-            raise
+            logger.info("Updated primaryauth runtime configuration on Valkey server")
+        except ValkeyWorkloadCommandError:
+            raise ValkeyConfigSetError("Could not set primaryauth on Valkey server.")
+
+    @tenacity.retry(
+        wait=tenacity.wait_fixed(5),
+        stop=tenacity.stop_after_attempt(5),
+        retry=tenacity.retry_if_result(lambda result: result is False),
```
Mehdi-Bendriss: these can be directly imported, so we don't keep referencing `tenacity.`

src/managers/cluster.py
```diff
@@ -56,11 +116,34 @@ def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObje
         if not self.workload.can_connect:
             status_list.append(CharmStatuses.SERVICE_NOT_STARTED.value)
 
-        if not self.state.unit_server.is_started:
-            status_list.append(CharmStatuses.SCALING_NOT_IMPLEMENTED.value)
+        # Peer relation not established yet, or model not built yet for unit or app
+        if not self.state.cluster.model or not self.state.unit_server.model:
+            return status_list if status_list else [CharmStatuses.ACTIVE_IDLE.value]
```
Mehdi-Bendriss: nit: (applies in the next occurrence as well)
```suggestion
            return status_list or [CharmStatuses.ACTIVE_IDLE.value]
```

src/managers/cluster.py
```diff
@@ -56,11 +116,34 @@ def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObje
...
 
-        if scope == "app":
-            # todo: remove when scaling is implemented
-            status_list.append(CharmStatuses.SCALING_NOT_IMPLEMENTED.value)
+        # non leader statuses
+        match self.state.unit_server.model.start_state:
+            case StartState.NOT_STARTED.value:
+                if (
+                    not self.state.cluster.internal_users_credentials
+                    or not self.state.number_units_started
```
Mehdi-Bendriss: similar comment/question as previously, on having a check on the primary ip

src/managers/cluster.py
```diff
@@ -56,11 +116,34 @@ def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObje
...
+            case StartState.STARTING_WAITING_SENTINEL.value:
+                status_list.append(
+                    ClusterStatuses.WAITING_FOR_SENTINEL_DISCOVERY.value,
+                )
+            case StartState.STARTING_WAITING_REPLICA_SYNC.value:
+                status_list.append(
+                    ClusterStatuses.WAITING_FOR_REPLICA_SYNC.value,
+                )
 
         return status_list if status_list else [CharmStatuses.ACTIVE_IDLE.value]
```
Mehdi-Bendriss: similar nit on using `or`

src/core/base_workload.py
```diff
@@ -4,16 +4,30 @@
...
+        self.user: str
```
Mehdi-Bendriss: with this, we unfortunately get a lot of linting warnings and inability to jump on the implementation using IDEs. We can either:
- define those as class body annotations (outside of the __init__)
- define the WorkloadBase as a `Protocol`interface
- define those as `abstract` properties

The latter may be a good compromise and help the linter+ide

src/managers/config.py
```diff
@@ -70,17 +74,36 @@ def config_properties(self) -> dict[str, str]:
...
             config_properties["bind"] = "0.0.0.0 -::1"
 
+        # replica related config
+        replica_config = self.generate_replica_config(primary_ip=primary_ip)
+        config_properties.update(replica_config)
+
         return config_properties
 
-    def set_config_properties(self) -> None:
+    def generate_replica_config(self, primary_ip):
```
Mehdi-Bendriss: should be a protected method and missing arg+return type hints

src/managers/config.py
```diff
@@ -116,6 +144,62 @@ def _get_user_acl_line(self, user: CharmUsers, passwords: dict[str, str] | None
...
         return f"user {user.value} on #{password_hash} {CHARM_USERS_ROLE_MAP[user]}\n"
 
+    def set_sentinel_config_properties(self, primary_ip: str) -> None:
+        """Write sentinel configuration file."""
+        logger.debug("Writing Sentinel configuration")
+
+        sentinel_config = f"port {SENTINEL_PORT}\n"
+
+        sentinel_config += f"aclfile {self.workload.sentinel_acl_file.as_posix()}\n"
+        # TODO consider adding quorum calculation based on number of units
```
Mehdi-Bendriss: we should work specifically with `planned_units()` here + handling of odd/even of said units

src/managers/config.py
```diff
@@ -124,6 +208,19 @@ def generate_password(self) -> str:
...
         return "".join([secrets.choice(string.ascii_letters + string.digits) for _ in range(32)])
 
+    def update_local_valkey_admin(self) -> None:
+        """Update the local unit's valkey admin password in the state."""
+        if not (
+            app_password := self.state.cluster.internal_users_credentials.get(
+                CharmUsers.VALKEY_ADMIN.value
+            )
+        ):
+            logger.warning("No valkey admin password found to update local unit state")
```
Mehdi-Bendriss: this is related to a previous comment on how do we react in this case from the caller code.

src/managers/sentinel.py
```diff
@@ -0,0 +1,116 @@
...
+        self.admin_user = CharmUsers.SENTINEL_CHARM_ADMIN.value
+
+    @property
+    def admin_password(self) -> str:
+        """Get the password of the admin user for the sentinel service."""
+        return self.state.cluster.internal_users_credentials.get(
+            CharmUsers.SENTINEL_CHARM_ADMIN.value, ""
+        )
+
+    @tenacity.retry(
```
Mehdi-Bendriss: similar comment on importing those decorators directly to shorten it

src/managers/sentinel.py
```diff
@@ -0,0 +1,116 @@
...
+                output, _ = client.exec_cli_command(
+                    command=["sentinel", "sentinels", PRIMARY_NAME],
+                    hostname=sentinel_ip,
+                )
+                if self.state.unit_server.model.private_ip not in output:
+                    logger.info(f"Sentinel at {sentinel_ip} has not discovered this sentinel")
+                    return False
+            except ValkeyWorkloadCommandError:
+                logger.warning(f"Could not query sentinel at {sentinel_ip} for primary discovery.")
+                continue
```
Mehdi-Bendriss: This defeats the purpose of iterating over all sentinels to check for discovery, no?

src/managers/sentinel.py
```diff
@@ -0,0 +1,116 @@
...
+            try:
+                output = client.exec_cli_command(
+                    command=["sentinel", "get-master-addr-by-name", PRIMARY_NAME],
+                    hostname=unit.model.private_ip,
+                )[0]
+                primary_ip = output.strip().split()[0]
+                logger.info(f"Primary IP address is {primary_ip}")
+                return primary_ip
+            except (IndexError, ValkeyWorkloadCommandError):
+                logger.error("Could not get primary IP from sentinel output.")
```
Mehdi-Bendriss: can we also log the error?

src/managers/sentinel.py
```diff
@@ -0,0 +1,116 @@
...
+            len(started_servers),
+        )
+
+    def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObject]:
+        """Compute the sentinel manager's statuses."""
+        status_list: list[StatusObject] = self.state.statuses.get(
+            scope=scope, component=self.name, running_status_only=True, running_status_type="async"
+        ).root
+
+        return status_list if status_list else [CharmStatuses.ACTIVE_IDLE.value]
```
Mehdi-Bendriss: consider using `or`

src/statuses.py
```diff
@@ -14,22 +14,65 @@
...
     )
-    SERVICE_NOT_STARTED = StatusObject(status="blocked", message="Service not started")
     SECRET_ACCESS_ERROR = StatusObject(
         status="blocked",
         message="Cannot access configured secret, check permissions",
         running="async",
     )
+    WAITING_TO_START = StatusObject(
+        status="maintenance",
+        message="Waiting for leader to authorize service start",
```
Mehdi-Bendriss:
```suggestion
        message="Waiting for leader to allow service start",
```

src/statuses.py
```diff
@@ -14,22 +14,65 @@
...
+    )
+
+    WAITING_FOR_REPLICA_SYNC = StatusObject(
+        status="maintenance",
+        message="Waiting for replica to sync with primary...",
+    )
+
+    WAITING_FOR_PRIMARY_START = StatusObject(
+        status="maintenance",
+        message="Waiting for primary to start and become active...",
```
Mehdi-Bendriss:
```suggestion
        message="Waiting for the primary unit to start...",
```

src/statuses.py
```diff
@@ -14,22 +14,65 @@
...
+    )
+
+
+class ValkeyServiceStatuses(Enum):
+    """Collection of possible Valkey service related statuses."""
+
+    SERVICE_STARTING = StatusObject(
+        status="maintenance",
+        message="waiting for valkey to start...",
```
Mehdi-Bendriss: Capital letter at the beginning of all statuses

Mehdi-Bendriss (commented):
Thanks Smail! Great work! I have left comments

src/common/client.py
skourta: We do not have access to the state inside the client

src/core/base_workload.py
skourta: The binary name differs between the rock and snap. This variable has the correct binary name to call the valkey cli

src/events/base_events.py
skourta: I added a TODO comment

src/events/base_events.py
skourta: We have a check earlier for `internal_users_credentials` for that.

src/managers/config.py
skourta: That should never happen as we check for `internal_users_credentials` on start and the leader unit sets them on `leader_elected`

src/events/base_events.py
skourta: Only the leader sets the passwords. The rest of the units react on the secret changed and update their local copies

src/events/base_events.py
skourta: Yes, I have reached out to her before I started the implementation and we agreed we would try to switch to her lib when available.

src/managers/cluster.py
skourta: For the get_statuses we only want to check the state we do not want to do any lengthy operation because this is run after every event. If one unit started then we have a primary that is the logic I followed.

src/managers/config.py
skourta: Updated the TODO comment. We will implement these when implementing the performance profiles / advanced configuration of the charm

src/common/client.py
Mehdi-Bendriss: i mean local variables, you can directly use `self.username`

src/core/models.py
skourta: Updated to `charmed_operator_password_local_unit_copy`.

src/events/base_events.py
skourta: Added both.

src/core/cluster_state.py
reneradoi: We do now have a changed functionality though because of this change: In the `leader_elected` event, we call this method without setting the `refresh` parameter, meaing we won't get up-to-date information on a leader-elected hook for a previously tracked secret. Leader-elected does not just happen at start up, so this could lead to inconsistencies. I would rather remove it again, or if you want to keep it, we should set the default to `True`.

src/managers/config.py
reneradoi: Yes, I would prefer to handle the sentinel config as a dict with the approach of loading the config template and setting values dict-like. Reasons being:
- standardized approach
- better readability in the code
- replacing/overwriting values in a dict is much easier that in a concatenated string
- it makes all config explicit, without relying on "default" values that might change between releases

src/core/base_workload.py
reneradoi: We also need to add a super class call to the workload classes now that this has been added to the base class.

src/literals.py
```diff
@@ -60,3 +69,13 @@ class Substrate(StrEnum):
 
     VM = "vm"
     K8S = "k8s"
+
+
+class StartState(StrEnum):
```
reneradoi, line 74: Nice!

src/events/base_events.py
```diff
@@ -55,31 +81,170 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
             return
+        self.charm.state.unit_server.update({"start_state": StartState.NOT_STARTED.value})
 
-        if not self.charm.unit.is_leader():
-            logger.warning("Scaling not implemented yet, services not started")
+        primary_ip = self.charm.sentinel_manager.get_primary_ip()
+        if self.charm.unit.is_leader() and not primary_ip:
+            if not self._start_services(event, primary_ip=self.charm.state.bind_address):
+                return
+            self.unit_fully_started.emit(is_primary=True)
```
reneradoi: What is the concept here in case starting fails? I see in case of errors for setting the config, we defer. But what about cases where the service doesn't come up (e.g. because of invalid config)?

src/core/cluster_state.py
skourta: Removed.

src/managers/config.py
skourta: Implemented.

src/events/base_events.py
```diff
@@ -55,31 +81,170 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
+            )
             event.defer()
-            return
+            return False
+        self.charm.state.statuses.delete(
+            CharmStatuses.CONFIGURATION_ERROR.value,
+            scope="unit",
+            component=self.charm.cluster_manager.name,
+        )
+        self.charm.status.set_running_status(
```
reneradoi, line 152: As we now have the `StartState`, we could use it for status computation in the manager and can avoid setting the running statuses here (except for configuration errors). This would remove ~20 lines of code and unclutter this method a lot.

src/events/base_events.py
```diff
@@ -55,31 +81,170 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
+        if self.charm.state.cluster.model.starting_member != self.charm.unit.name:
+            logger.info("Non-leader unit waiting for leader to choose it as starting member")
+            event.defer()
+            return
+
+        if not self._start_services(event, primary_ip=primary_ip):
+            return
+        self.unit_fully_started.emit(is_primary=False)
+
+    def _start_services(self, event: ops.StartEvent, primary_ip: str) -> bool:
```
reneradoi: I don't see the added value of the returned bool here. Given that we have health checks and deferral in the `unit_fully_started` event handling, what is the point of checking here already and returning early without doing anything? I think we should just start the services, and leave the responsibility for checking if started to the health checks.

src/events/base_events.py
```diff
@@ -55,31 +81,170 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
+        )
+        self.charm.status.set_running_status(
+            ValkeyServiceStatuses.SERVICE_STARTING.value,
+            scope="unit",
+            component_name=self.charm.cluster_manager.name,
+            statuses_state=self.charm.state.statuses,
+        )
 
         self.charm.workload.start()
+        if not self.charm.workload.alive():
```
reneradoi: Same question here: Why have two places to check the health?

src/events/base_events.py
```diff
@@ -55,31 +81,170 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
+            # if the starting member has not started yet, we want to wait for it to start instead of choosing another unit that requested start
+            self.charm.state.cluster.model.starting_member
+            and starting_unit
+            and not starting_unit.is_started
+        ):
+            logger.debug(
+                "Starting member %s has not started yet. Units requesting start: %s. ",
+                self.charm.state.cluster.model.starting_member,
+                units_requesting_start,
+            )
```
reneradoi, line 243: I think we should return here to avoid overwriting the starting member if the set of units returns a different order (might be due to edge cases with unit numbers while adding units during the start process).

src/events/base_events.py
```diff
@@ -55,31 +81,170 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
-            logger.info("Unit %s has joined the relation", event.unit.name)
+    def _on_peer_relation_changed(self, event: ops.RelationChangedEvent) -> None:
+        """Handle event received by all units when a unit's relation data changes."""
+        if not self.charm.unit.is_leader():
+            return
+
+        units_requesting_start = [
+            unit.unit_name
+            for unit in self.charm.state.servers
+            if unit.model and unit.model.request_start_lock
```
reneradoi, line 223: I like the idea of units requesting a start lock!

src/events/base_events.py
```diff
@@ -92,6 +257,13 @@ def _on_leader_elected(self, event: ops.LeaderElectedEvent) -> None:
             event.defer()
             return
 
+        self.charm.state.unit_server.update(
+            {
+                "hostname": socket.gethostname(),
+                "private_ip": self.charm.state.bind_address,
```
reneradoi, line 268: Why duplicating this address into `private_ip` and not just taking the information from the state directly?

src/managers/sentinel.py
```diff
@@ -0,0 +1,134 @@
...
+                    logger.info(f"Sentinel at {sentinel_ip} has not discovered this sentinel")
+                    return False
+            except ValkeyWorkloadCommandError:
+                logger.warning(f"Could not query sentinel at {sentinel_ip} for primary discovery.")
+                return False
+        return True
+
+    def get_primary_ip(self) -> str | None:
+        """Get the IP address of the primary node in the cluster."""
+        started_servers = [unit for unit in self.state.servers if unit.model and unit.is_started]
```
reneradoi: I believe we can avoid the `if unit.model` check here and in the `is_sentinel_discovered` method because this is already included in the `unit.is_started` flag and will return `False` if a peer relation doesn't exist yet.

src/events/base_events.py
```diff
@@ -55,31 +81,170 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
+            return
+
+        self.charm.state.unit_server.update({"request_start_lock": True})
+
+        # TODO unit.name would not work across models we need to switch to using `model.unit.name + model_uuid`
+        if self.charm.state.cluster.model.starting_member != self.charm.unit.name:
+            logger.info("Non-leader unit waiting for leader to choose it as starting member")
+            event.defer()
+            return
+
```
reneradoi, line 121: What would happen in the following edge case?
- a primary exists
- a new unit is added but not yet started
- the new unit is appointed Juju leader (eg. because of a network disruption)

This new unit will put `request_start_lock` into the databag, but `starting_member` is empty and it will always defer here, because no other unit can appoint the newly added leader unit as the `starting_member`.

src/managers/cluster.py
```diff
@@ -30,22 +34,85 @@ def __init__(self, state: ClusterState, workload: WorkloadBase):
...
+            persistence_info := client.get_persistence_info(hostname=self.state.bind_address)
+        ) and persistence_info.get("loading", "") != "0":
+            logger.warning("Health check failed: Valkey server is still loading data.")
+            return False
+        if is_primary and not client.set_value(
+            hostname=self.state.bind_address, key="healthcheck", value="ok"
+        ):
+            logger.warning("Health check failed: Could not set test key on Valkey server.")
+            return False
+
```
reneradoi: I am not a big fan of writing data to the database as part of the health check from an operator. We should only ever do this as part of a smoke test for deployments, but not as part of the operator code itself.

From my point of view, the `PING` command Valkey provides is sufficient for a health check, as it does also provide actual workload checks instead of just the network ping. According to the [docs](https://valkey.io/commands/ping/), it is `verifying the server's ability to serve data - an error is returned when this isn't the case (e.g., during load from persistence or accessing a stale replica).`.

src/managers/cluster.py
```diff
@@ -56,11 +123,35 @@ def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObje
         if not self.workload.can_connect:
             status_list.append(CharmStatuses.SERVICE_NOT_STARTED.value)
```
reneradoi: With the invention of the `StartState` that returns `not_started` by default, we can now remove this what was a workaround before we had actual startup logic.

src/core/cluster_state.py
reneradoi: Thank you!

src/managers/config.py
reneradoi: Thank you!

reneradoi (commented):
I've completed the 2nd review now. A few questions regarding the workflow are still to be clarified, and a few comments regarding simplifying/cleaning up here and there.

src/events/base_events.py
skourta: I've removed all statuses I can. For configuration and service starting we cannot remove it. The configuration does not a state to detect it. The service starting is to update the status without waiting for the hook to finish.

src/events/base_events.py
skourta: @Mehdi-Bendriss  requested I add a liveness check inside the start method of the workload.

src/events/base_events.py
skourta: But we are filtering based on the unit name. The order of the returned units does not really matter.

src/events/base_events.py
skourta: We need the IPs of all units to perform some checks like `is_sentinel_discovered` where we loop through the started units and check if they have the newly started sentinel IP.

src/managers/sentinel.py
skourta: It was an addiotional check to silent the linters

src/managers/cluster.py
skourta: Removed

src/events/base_events.py
skourta: That is true. I think it makes sense to defer and block on the health checks and only defer with maintenance status on the waiting for sentinel/ replica sync. WDYT?

src/events/base_events.py
skourta: Mostly to return early and not start the services is there was an error writing the configuration files.

src/events/base_events.py
reneradoi: For configuration errors I agree. For displaying statuses during the start process, I think we can indeed avoid the running status and simplify a bit more:
- the cases `STARTING_WAITING_VALKEY`, `STARTING_WAITING_SENTINEL` and `STARTING_WAITING_REPLICA_SYNC` can stay as they are
- if we have the `request_start_lock` in the databag: `waiting to start`
- otherwise: not started

Would be enough from my point of view, I don't think there's a lot of value in having an additional status saying "the service is starting".

src/events/base_events.py
reneradoi: Then we could boil this method down to just calling the methods for setting config and starting workload, and add the error handling in the `_on_start` method, where we try and in case of error, set the running status for the config error and defer the event. Otherwise we just move on to the health checks, where the single point of responsibility/truth is allocated for checking if all our services came up cleanly after the start procedure.

src/events/base_events.py
reneradoi: The comment some lines above says: `if the starting member has not started yet, we want to wait for it to start instead of choosing another unit that requested start`.

We only filter the `starting_unit`, but overwrite this information based on the order of `units_requesting_start`.

src/events/base_events.py
reneradoi: That makes sense, I've overlooked that.

src/events/base_events.py
skourta: Yes, this is an edge case. I will work on fixing it.

src/events/base_events.py
skourta: I like that. I will try to simplify more.

src/events/base_events.py
skourta: Yes, I'm adding a couple of States for the StartSate and simplifying the logic a bit more. I went a rabbit hole when I found a descrepency between VM and K8s on how starting the services behave.

src/events/base_events.py
skourta: Oh yes. Now I see what you mean. I will add the return.

skourta: Added

src/events/base_events.py
skourta: I've reworked all statuses and we do not have any running statuses now. The downside is that we go from waiting for leader to authorize start to directly the unit fully started because the statuses are only recomputed atthe end of the hook. I think we should add at least one running statuses to display that we are starting the services.

src/core/base_workload.py
skourta: Moved to class body annotations

src/events/base_events.py
skourta: Updated the start workflow. Now we detect config problems on `workload.start`. I have added the same check that is included in the k8s `restart_services` function to our `workload_vm`'s start function.

src/events/base_events.py
skourta: I've simplified the start workflow and now consider the leader to be a normal unit. It asks for the lock as well.

src/events/base_events.py
skourta: Under the new start workflow this shouldn't be an edge case anymore. There are only two paths for the workflow now: one for the primary node and one for the replicas. The first unit that gets the lock starts the primary node the rest are considered replicas.

src/managers/config.py
Mehdi-Bendriss: This should be part of HA, could be in another PR but still part of the HA epic. It is still unclear what performance profiles will look like / when they'll be added

skourta: Understood. I will schedule it right after scale down since we will need to update the quorom then as well.

src/managers/sentinel.py
```diff
@@ -0,0 +1,132 @@
...
+            password=self.admin_password,
+            workload=self.workload,
+            connect_to="sentinel",
+        )
+
+        for unit in started_servers:
+            if primary_ip := client.sentinel_get_primary_ip(hostname=unit.model.private_ip):
+                logger.info(f"Primary IP address is {primary_ip}")
+                return primary_ip
+        logger.error(
```
reneradoi: Minor comment: Let's make this a warning.

src/events/base_events.py
```diff
@@ -55,31 +88,180 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
+                self.charm.state.cluster.model.starting_member,
+                units_requesting_start,
+            )
+            return
+
+        self.charm.state.cluster.update(
+            {"starting_member": units_requesting_start[0] if units_requesting_start else ""}
+        )
+        logger.debug(
+            f"Updated starting member to {units_requesting_start[0] if units_requesting_start else ''}"
```
reneradoi: This results in `Updated starting member to ` if there is no more starting member.

src/events/base_events.py
```diff
@@ -55,31 +88,180 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
+            and not starting_unit.is_started
+        ):
+            logger.debug(
+                "Starting member %s has not started yet. Units requesting start: %s. ",
+                self.charm.state.cluster.model.starting_member,
+                units_requesting_start,
+            )
+            return
+
+        self.charm.state.cluster.update(
```
reneradoi: I've found this in the log: `Non existent field starting-member was attempted to be removed from the databag`. We could avoid this by checking if it exists before updating (also a bit easier to read):

```python
if units_requesting_start:
    state.cluster.update({"starting_member": units_requesting_start[0]})
    logger.debug(...)
    return

if self.charm.state.cluster.model.starting_member:
    state.cluster.update({"starting_member": ""})
```

src/events/base_events.py
```diff
@@ -55,31 +88,180 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
+        self.charm.status.set_running_status(
+            StartStatuses.SERVICE_STARTING.value,
+            scope="unit",
+            statuses_state=self.charm.state.statuses,
+            component_name=self.charm.cluster_manager.name,
+        )
+
+        self.unit_fully_started.emit(is_primary=primary_ip == self.charm.state.bind_address)
+
+    def _configure_services(self, primary_ip: str) -> None:
```
reneradoi: All of the calls in this method go to the config manager; we could move this method to the config manager too, which makes it more clear regarding responsibility.

src/workload_vm.py
```diff
@@ -83,22 +102,71 @@ def install(self, revision: str | None = None, retry_and_raise: bool = True) ->
...
-            self.valkey.start(services=[SNAP_SERVICE])
+            self.valkey.start(services=[SNAP_SERVICE, SNAP_SENTINEL_SERVICE])
         except snap.SnapError as e:
             logger.exception(str(e))
+            raise ValkeyServicesFailedToStartError(f"Failed to start Valkey services: {e}") from e
+
+        # The service might start but fail to load and die immediately
+        # On k8s starting the services will wait (poll) for them to be started.
+        # We do the same here to make sure the services are alive after start.
+        if not self.wait_for_services_to_be_alive(duration=3):
```
reneradoi, line 113: We have three different wait/retry mechanisms on this one:
- the loop in the method `wait_for_services_to_be_alive`
- the retry on the method `wait_for_services_to_be_alive`
- the retry on the `alive` method

Wouldn't it be sufficient to have a retry with backoff on the alive method?

src/events/base_events.py
```diff
@@ -55,31 +88,180 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
+            return
+        except (ValkeyServicesFailedToStartError, ValkeyServiceNotAliveError) as e:
+            logger.error(e)
+            self.charm.state.unit_server.update(
+                {"start_state": StartState.ERROR_ON_START.value, "request_start_lock": False}
+            )
+            event.defer()
+            return
+
+        self.charm.status.set_running_status(
```
reneradoi, line 152: Where do we reset this running status? Also I've found that this is being computed in the cluster manager's `get_statuses` method - I believe it should only be one of that, no?

src/events/base_events.py
reneradoi: I have to come back to this one: The `private-ip` that we add/maintain in the unit databag is a duplicate of the `private-address` that Juju populates into the unit databag. I think we should use the `private-address` instead for the purpose of getting other unit's addresses and remove the `private-ip` from our code.

reneradoi (commented):
Hey Smail! I've tested multiple constellations of deploying and scaling up, the workflow works really well now and also the UX regarding statuses is pretty nice! I've only some minor comments on the code that are left, but the overall concept is now good to go from my point of view.

src/events/base_events.py
reneradoi: I like it now! Having one running status for UX purposes is fine, it's much cleaner now and most of the other information can be computed based on the nice `StartState` you've created.

src/events/base_events.py
skourta: I missed that juju provides one in the unit databag. I will update the model attribute name from `private_ip` to `private_address` so we can read the juju populated address using the unit model.

src/events/base_events.py
skourta: It gets cleared when the status event handler is ran because always reset the statuses of the cluster statuses manager to an empty list. We do not need to clean it explicitely.

skourta: I only added it explicitely here so we get a nice UX for users and they see the status change before the hook fininshes running.

src/workload_vm.py
skourta: Th reason we are not just using the `alive` method is when the check is done after starting the workload we do it too fast it return true. After a small amount of time the service might crash if the config is invalid for example. Adding this behaviour makes it behave similar to what happens on k8s.
We can potentially remove the retries on the `wait_for_services_to_be_alive`.

src/events/base_events.py
skourta: It seems the address juju populates in the unit databag is not the container's address on k8s so we cannot use it.
image

reneradoi: Alright, then let's leave it at this, sorry for the noise.

src/events/base_events.py
reneradoi: Nice, thank you!

src/events/base_events.py
skourta: I got excited for a second as well. If juju sets it automatically we don't have to maintain it.

reneradoi (approved):
Great result, very nice improvements!

src/events/base_events.py
```diff
@@ -69,111 +88,107 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
+        # TODO unit.name would not work across models we need to switch to using `model.unit.name + model_uuid`
+        if self.charm.state.cluster.model.starting_member != self.charm.unit.name:
+            logger.info("Waiting for lock to start")
             event.defer()
             return
 
-        self._start_services(event, primary_ip=primary_ip)
-        self.unit_fully_started.emit()
+        primary_ip = self.charm.sentinel_manager.get_primary_ip()
+        if not primary_ip:
```
Mehdi-Bendriss: nit: walrus, we should try to have this as a coding practice in general

src/managers/config.py
```diff
@@ -144,35 +146,95 @@ def _get_user_acl_line(self, user: CharmUsers, passwords: dict[str, str] | None
...
+            # keyword argument1 argument2 ... argumentN
+            # sentinel keyword argument1 argument2 ... argumentN
+            for line in config:
+                line = line.strip().lower()
+                if not line or line.startswith("#"):
+                    # ignore comments and empty lines
+                    continue
+                elif line.startswith("sentinel "):
+                    try:
+                        key, value = line.split(" ", 2)[1:]
```
Mehdi-Bendriss, line 175: can you add a comment about this case? this is hard to interpret just from reading

src/managers/config.py
```diff
@@ -144,35 +146,95 @@ def _get_user_acl_line(self, user: CharmUsers, passwords: dict[str, str] | None
...
+                    except ValueError:
+                        key = line.strip().split(" ", 1)[1]
+                        value = ""
+                    sentinel_properties[key.strip()] = value.strip().replace(
+                        "mymaster", PRIMARY_NAME
+                    )
+                else:
+                    try:
+                        key, value = line.split(" ", 1)
+                    except ValueError:
```
Mehdi-Bendriss, line 188: similar comment on leaving a comment here

Mehdi-Bendriss (commented):
Thanks Smail! Great work! Only small nits remaining for me

Mehdi-Bendriss (approved):

PR #12, opened 2026-02-11
[DPE-9314] Add support for peer and client TLS

reneradoi:
This PR adds client TLS to Valkey, based on the `tls-certificates` interface in version `v4`. If no client TLS is enabled, communication between Valkey peers and between Sentinel and Valkey is still encrypted by default, using self-signed certificates.

Main features:
- enabling or disabling TLS happens without restarting the Valkey server, thus no need for `rolling ops` and we can avoid using an additional peer relation for coordination
- TLS mode will only be enabled once all units have stored the respective certificates (to avoid communication downtime)
- supports a directory for CA certificates, thus enabling using multiple CA certs for verification at the same time
- enhances the Valkey client to use TLS certificates when client TLS is enabled.

Please note: In order to not break the workflow for cluster initialization, the starting procedure is halted in case client TLS is enabled but a unit does not yet have the client TLS certificates installed. Once available, Valkey will start directly with client TLS.


src/common/client.py
```diff
@@ -31,20 +35,43 @@ def __init__(
...
+        tls_ca_cert: bytes | None,
     ):
         self.hosts = hosts
         self.user = username
         self.password = password
+        self.tls_cert = tls_cert
+        self.tls_key = tls_key
+        self.tls_ca_cert = tls_ca_cert
 
     async def create_client(self) -> GlideClient:
```
skourta: I think we should continue using the cli inside the charm and avoid using glide. For a single unit this is not an issue but for multiple units you will have to reload them on every single instance individually.

src/common/client.py
```diff
@@ -31,20 +35,43 @@ def __init__(
...
+
+            client_config = GlideClientConfiguration(
+                addresses,
+                use_tls=True,
+                credentials=credentials,
+                request_timeout=1000,  # in milliseconds
+                advanced_config=AdvancedGlideClientConfiguration(tls_config=tls_config),
+            )
+        else:
+            client_config = GlideClientConfiguration(
```
skourta: nit:
You can simplify this and remove the if else if you use ternaries for the tls_config and the advance_config.

src/common/client.py
```diff
@@ -77,3 +103,25 @@ def reload_acl(self) -> None:
         except ValkeyCustomCommandError as e:
             logger.error(f"Error loading ACL: {e}")
             raise ValkeyACLLoadError(f"Could not load ACL: {e}")
+
+    def reload_tls(self, tls_config: dict[str, str]) -> None:
+        """Trigger Valkey to load the TLS settings."""
+        cmd = ["CONFIG", "SET"]
+
+        # avoid "bind: Address already in use" by advancing the disabled port
+        if tls_config["tls-port"] == "0":
```
skourta: Can you please explain this further (the pop at the end)? The comment was not enough for me to get the full context.

src/events/tls.py
```diff
@@ -0,0 +1,191 @@
...
+        logger.info("Storing %s certificate", cert_type.value)
+        try:
+            self.charm.tls_manager.write_certificate(cert_type, cert, private_key)
+            self.charm.tls_manager.set_cert_state(cert_type, is_ready=True)
+        except ValkeyWorkloadCommandError as e:
+            logger.error("Failed to store certificate: %s", e)
+            event.defer()
+            return
+
+        if tls_state == TLSState.TO_TLS:
```
skourta, line 124: In case we are rotating the certificate we need to reload them in valkey. I do not see that here. The state would be `TLSState.TLS`

tests/integration/tls/test_tls.py
```diff
@@ -0,0 +1,159 @@
...
+        hostnames=hostnames,
+        username=CharmUsers.VALKEY_ADMIN.value,
+        password=password,
+        tls_enabled=True,
+        key=TEST_KEY,
+        value=TEST_VALUE,
+    )
+    assert result == "OK", "Failed to write data with TLS enabled"
+
+    assert await get_key(
```
skourta, line 65: we should also test writing without TLS and making sure it fails

tests/integration/tls/test_tls.py
```diff
@@ -0,0 +1,159 @@
...
+        hostnames=hostnames,
+        username=CharmUsers.VALKEY_ADMIN.value,
+        password=password,
+        tls_enabled=True,
+        key=TEST_KEY,
+    ) == bytes(TEST_VALUE, "utf-8"), "Failed to read data with TLS enabled"
```
skourta, line 152: Likewise I would like to add a check for a client not using TLS

metadata.yaml
```diff
@@ -27,10 +27,20 @@ resources:
   valkey-image:
     type: oci-image
     description: OCI Image for Valkey
-    upstream-source: ghcr.io/canonical/valkey@sha256:3f884d584eac51f3794d3538861f84e5f9e866b890ae0869deb7e4df6fc8eb21
+    upstream-source: ghcr.io/canonical/valkey:9.0.1-26.04-edge
```
skourta, line 30: I found myself dealing with cache issues and had to use the hash locally to make it work.

tests/unit/test_tls.py
```diff
@@ -0,0 +1,445 @@
...
+    STATUS_PEERS_RELATION,
+)
+from src.statuses import TLSStatuses
+
+from .helpers import status_is
+
+CONTAINER = "valkey"
+
+
+def test_peer_tls_relation_created_no_client_tls(cloud_spec):
```
skourta: Very nice test coverage! Thank you

skourta (commented):
Looks really neat. Thank you @reneradoi . I have one main concern.

src/common/client.py
reneradoi: Yes I agree. The plan is to keep it as is until the scale-up PR is merged (I want to avoid back-and-forth with the other PR). Then I will update here or follow up with a small PR (also adjusting the unit number in integration tests).

src/common/client.py
reneradoi: The issue here is we cannot assign a port if it is still in use. When we have:
- port=0, tls-port=6379 -> we cannot assign 6379 to port
- port=6379, tls-port=0 -> we cannot assign 0 to port

The order in `ConfigManager.generate_tls_config` is: port, tls-port. We either have to adjust this there, based on whether we switch to TLS or back. Or we can do it here in the client, based on the value. I opted for here because it is required by how the client applies commands.

The third option would be to use different port numbers, but I ruled that out as not desired.

metadata.yaml
reneradoi: I found myself dealing with an old image when I expected a package to be available and therefore updated it :)

src/events/tls.py
reneradoi: That is one of the few changes I expect to do when implementing cert rotation. I've deliberately left it for then.

src/common/client.py
reneradoi: Updated. I had it with if-else because I thought it would be clearer, but it's fine for me.

tests/integration/tls/test_tls.py
reneradoi: Good point - updated.

src/managers/cluster.py
```diff
@@ -33,20 +33,44 @@ def __init__(self, state: ClusterState, workload: WorkloadBase):
...
         except ValkeyACLLoadError:
             raise
 
+    def reload_tls_settings(self, tls_config: dict[str, str]) -> None:
+        """Update TLS by loading the TLS settings."""
+        try:
+            client = self._get_valkey_client()
+            client.reload_tls(tls_config)
+        except ValkeyTLSLoadError:
+            raise
```
Mehdi-Bendriss: why catch if we're simply re-raising?

tests/integration/k8s/test_charm.py
```diff
@@ -112,7 +118,12 @@ async def test_update_admin_password(juju: jubilant.Juju) -> None:
     juju.config(app=APP_NAME, reset=[INTERNAL_USERS_PASSWORD_CONFIG])
 
     # wait for config-changed hook to finish executing
-    juju.wait(lambda status: jubilant.all_agents_idle(status, APP_NAME), timeout=1200)
+    juju.wait(
+        lambda status: jubilant.all_agents_idle(status, APP_NAME),
+        timeout=1200,
+        delay=5,
+        successes=3,
```
Mehdi-Bendriss: can you explain this?

Mehdi-Bendriss (commented):
Thank you René, great work and really clean!
 
I was really scratching my head on this one for a while, until I realized after a few hours of research that there is no way to really **enforce** TLS on clients while not doing so for peers. 

1. Can you confirm my understanding is correct:

| Scenario	           | `port`   | `tls-port` | `tls-replication` | behavior                                                                                   |
|---------------------|--------|----------|-----------------|-----------------------------------------------------------------------------------------------------|
| 1. Both Disabled    | 6379   | 0        | "no"              | Standard plaintext mode. No encryption active.                                                      |
| 2. Client TLS Only  | 6379   | 6380     | "no"              | Clients **NOT prevented** from connecting to 6379 without TLS, and replicas/Sentinels use 6379.     |
| 3. Peer TLS Only    | 6379   | 6380     | "yes"             | Replicas/Sentinels **forced to use TLS** (6380). Clients **not prevented** from using 6379 (Plain). |
| 4. Both Enabled	    | 0      | 6380     | "yes"             | All traffic (Client and Peer) **must use TLS.** Plaintext port is closed.                           |

2. If the above is correct, it doesn't make sense to handle `2.` but it does make sense to handle `1, 3, 4`, and it doesn't seem that we're handling 3, which can also be enriched an ACL (where we block any "non replication" user from psync/replconf/resync) for safeguarding against an attempted "fake" replica joining.

src/common/client.py
Mehdi-Bendriss: I think It would make sense to add this as a comment or tweak the logic to be less obscure

src/managers/cluster.py
reneradoi: True!

tests/integration/k8s/test_charm.py
reneradoi: I've noticed that the existing tests (from previous PR's) are not 100% consistent in terms of waiting for the units to finish execution, which results in authentication checks before something was enabled/changed. To have a minimal invasive stabilization, I opted for increasing the wait time on the `jubilant.wait()` by this, to not create too many conflicts with Smail's [changes](https://github.com/canonical/valkey-operator/pull/8/changes#diff-642cf7dd7aa6642e06e3a0c51708ea494904fde750d3eea6a10a938850081d88R76) in the scale-up PR that are addressing this topic.

reneradoi:
> Thank you René, great work and really clean!
> 
> I was really scratching my head on this one for a while, until I realized after a few hours of research that there is no way to really **enforce** TLS on clients while not doing so for peers.
> 
> 1. Can you confirm my understanding is correct:
> 
> Scenario	`port`	`tls-port`	`tls-replication`	behavior
> 1. Both Disabled	6379	0	"no"	Standard plaintext mode. No encryption active.
> 2. Client TLS Only	6379	6380	"no"	Clients **NOT prevented** from connecting to 6379 without TLS, and replicas/Sentinels use 6379.
> 3. Peer TLS Only	6379	6380	"yes"	Replicas/Sentinels **forced to use TLS** (6380). Clients **not prevented** from using 6379 (Plain).
> 4. Both Enabled	0	6380	"yes"	All traffic (Client and Peer) **must use TLS.** Plaintext port is closed.
> 2. If the above is correct, it doesn't make sense to handle `2.` but it does make sense to handle `1, 3, 4`, and it doesn't seem that we're handling 3, which can also be enriched an ACL (where we block any "non replication" user from psync/replconf/resync) for safeguarding against an attempted "fake" replica joining.

To summarize our offline discussion on this topic: The main concern here is that Valkey doesn't really have the concept of `peer TLS`, as other servers (replicas) are always connecting to a Valkey server as a client. The only reason why Valkey supports having dedicated TLS certificates for connections to other Valkey servers is to account for the case where a TLS certificate is issued with attributes that designate it as "client-only" or "server-only certificates" (which we usually don't do in our TLS providers).

For this reason, I would like to avoid peer TLS at all and only have one TLS interface ("client TLS") for the charm. This would be adhering to upstream's understanding and concepts, while also relieving users from the burden to always have to enable both TLS interfaces. The main benefit that we usually have from having separate TLS interfaces for internal (peer) vs. external (client) communication would not be available for us anyway, because Valkey does not distinguish between external clients and replicas, and it would always use the same "client TLS certificate", also for internal (peer) connections.

As a side note: I would really like to avoid the downsides of supporting different ports depending on the current state of TLS relations. This would introduce a lot of complexity into the charm and would also have negative effects on clients that would have to switch ports when peer TLS is enabled/disabled (which should be transparent to clients).

src/common/client.py
```diff
@@ -97,11 +112,7 @@ def get_persistence_info(self, hostname: str) -> dict[str, str] | None:
                 continue
             values_parts = line.split(":", 1)
             if len(values_parts) != 2:
-                logger.error(
-                    "Unexpected output format when getting persistence info from Valkey server at %s: %s",
-                    hostname,
-                    output,
-                )
```
reneradoi, line 104: I have removed a lot of debugging log output in case of errors in this class. When testing, I noticed that these logs print passwords in clear text, and this happens all over the place here. We really have to be careful with that, unfortunately it slipped through in the reviews. Please be careful.

src/common/client.py
reneradoi: This is no longer necessary after refactoring.

tests/integration/k8s/test_charm.py
reneradoi: Removed after merging the changes from `9/edge`.

src/common/client.py
```diff
@@ -199,16 +203,31 @@ def load_acl(self, hostname: str) -> bool:
...
             return False
         except ValkeyWorkloadCommandError as e:
             logger.error(f"Failed to load ACL file on Valkey server at {hostname}: {e}")
             return False
 
+    def reload_tls(self, tls_config: dict[str, str], hostname: str) -> None:
+        """Trigger to load the TLS settings."""
+        cmd = ["CONFIG", "SET"]
+
+        for key, value in tls_config.items():
```
skourta, line 219: I would rather we do this at the manager level and then call `config_set`. This would keep the client tightly coupled to the CLI commands and if we move to a different client it would be easy enough.

src/core/base_workload.py
```diff
@@ -14,6 +14,33 @@
 logger = logging.getLogger(__name__)
 
 
+class TLSPaths:
+    """Object to store the TLS paths."""
+
+    def __init__(self, tls_root: pathops.LocalPath or pathops.ContainerPath):
```
skourta: For typing we need to use the `|` operator.
```suggestion
    def __init__(self, tls_root: pathops.LocalPath | pathops.ContainerPath):
```

skourta: Same thing applies for the return types of the other properties.

src/core/models.py
```diff
@@ -115,6 +132,11 @@ def unit_name(self) -> str:
         """The unit's name."""
         return self.unit.name
 
+    @property
+    def hostname(self) -> str:
```
skourta: Why have the property if this is not computed? We can just access it through the model no?

src/core/models.py
```diff
@@ -127,6 +149,19 @@ def valkey_admin_password(self) -> str:
...
+    @property
+    def tls_client_state(self) -> TLSState:
+        """The current TLS state of the Valkey server for client TLS."""
+        if not self.model:
+            return TLSState.NO_TLS
+
+        return TLSState(self.model.tls_client_state or TLSState.NO_TLS.value)
+
+    @property
+    def client_cert_ready(self) -> bool:
```
skourta: Same here

src/events/base_events.py
```diff
@@ -258,7 +262,8 @@ def _on_update_status(self, event: ops.UpdateStatusEvent) -> None:
 
     def _on_leader_elected(self, event: ops.LeaderElectedEvent) -> None:
         """Handle the leader-elected event."""
-        if not self.charm.state.peer_relation:
+        if not (self.charm.state.peer_relation and self.charm.workload.can_connect):
```
skourta, line 273: Great catch. I think this solves some weird behaviour I was investigating yesterday.

pyproject.toml
```diff
@@ -51,7 +52,8 @@ data-platform-helpers = ">=0.1.7"
 jubilant = "^1.6.0"
 python-dateutil = "*"
 tenacity = "^9.1.2"
-valkey-glide = "^2.2.7"
+# https://github.com/valkey-io/valkey-glide/pull/5124 not yet released
```
skourta, line 55: We should stick to the original version. We do not build from source for the integration tests. We just do it for the charm.

src/managers/cluster.py
```diff
@@ -89,14 +88,12 @@ def is_replica_synced(self) -> bool:
...
-        client = ValkeyClient(
-            username=self.admin_user,
-            password=self.admin_password,
-            workload=self.workload,
-        )
+        client = self._get_valkey_client()
 
         if not client.ping(hostname=self.state.bind_address):
             logger.warning("Health check failed: Valkey server did not respond to ping.")
+            # restart to pick up updated TLS certs if available
```
skourta: We should also restart sentinel to pick up on the changes no?

src/managers/cluster.py
```diff
@@ -40,23 +40,26 @@ def admin_password(self) -> str:
...
-    def reload_acl_file(self) -> None:
-        """Reload the ACL file into the cluster."""
-        client = ValkeyClient(
+    def _get_valkey_client(self) -> ValkeyClient:
+        """Get a client connection to Valkey."""
+        return ValkeyClient(
             username=self.admin_user,
             password=self.admin_password,
+            tls=True
+            if self.state.unit_server.tls_client_state in [TLSState.TLS, TLSState.TO_NO_TLS]
```
skourta, line 49: If we reuse this a lot we might want to make this a property

src/managers/config.py
```diff
@@ -99,14 +104,55 @@ def _generate_replica_config(self, primary_ip: str) -> dict[str, str]:
         if primary_ip != self.state.unit_server.model.private_ip:
             # set replicaof
             logger.debug("Setting replicaof to primary %s", primary_ip)
-            replica_config["replicaof"] = f"{primary_ip} {CLIENT_PORT}"
+            replica_config["replicaof"] = f"{primary_ip} {TLS_PORT}"
```
skourta, line 108: Can we leave a comment for future maintainers to remember this is because we always use TLS for peers?

src/managers/sentinel.py
```diff
@@ -106,15 +108,12 @@ def get_primary_ip(self) -> str | None:
...
-            username=self.admin_user,
-            password=self.admin_password,
-            workload=self.workload,
-            connect_to="sentinel",
-        )
+        client = self._get_valkey_client()
 
         if not client.ping(hostname=self.state.bind_address):
             logger.warning("Health check failed: Sentinel did not respond to ping.")
+            # restart to pick up updated TLS certs if available
```
skourta: Likewise if one service is not healthy and the other is then we will restart one and not restart the other.

src/events/tls.py
```diff
@@ -0,0 +1,222 @@
...
+        # in case a non-leader unit gets the event before the leader unit has processed it
+        if not self.charm.state.cluster.internal_ca_certificate:
+            logger.warning("Self-signed CA certificate not yet available")
+            event.defer()
+            return
+
+        try:
+            self.charm.tls_manager.create_and_store_self_signed_certificate()
+        except ValkeyWorkloadCommandError as e:
+            logger.error("Failed to create certificate for peer-TLS, startup will fail: %s", e)
```
skourta, line 90: Shouldn't we display a status to notify the user?

src/events/tls.py
```diff
@@ -0,0 +1,222 @@
...
+            except (
+                ValkeyWorkloadCommandError,
+                ValkeyServicesFailedToStartError,
+                ValkeyTLSLoadError,
+                ValueError,
+            ):
+                logger.error("Failed to enable client TLS")
+                event.defer()
+                return
+            except ValkeyCertificatesNotReadyError:
```
skourta, line 138: Is it necessary to defer here? We are have any logic after this exception block.

src/events/tls.py
```diff
@@ -0,0 +1,222 @@
...
+                logger.warning("Not all units have stored the client certificate")
+                event.defer()
+                return
+
+    def _on_tls_relation_broken(self, event: ops.RelationBrokenEvent) -> None:
+        """Handle the `relation-broken` event."""
+        if self.charm.app.planned_units() == 0:
+            return
+
+        if not self.charm.state.cluster.internal_ca_certificate:
```
skourta, line 148: Can you please clarify when this scenario would happen?

src/events/tls.py
```diff
@@ -0,0 +1,222 @@
...
+        if self.charm.state.unit_server.tls_client_state in [TLSState.TLS, TLSState.TO_NO_TLS]:
+            logger.info("Disabling client TLS")
+            self.charm.tls_manager.set_tls_state(TLSState.TO_NO_TLS)
+            try:
+                primary_ip = self.charm.sentinel_manager.get_primary_ip()
+                self.charm.config_manager.set_config_properties(primary_ip=primary_ip)
+                tls_config = self.charm.config_manager.generate_tls_config()
+                self.charm.cluster_manager.reload_tls_settings(tls_config)
+                self.charm.config_manager.set_sentinel_config_properties(primary_ip=primary_ip)
+                self.charm.sentinel_manager.restart_service()
```
skourta, line 165: Do we need to restart sentinel to pick up on the changes? Also, this need to be a rolling op.

src/events/tls.py
```diff
@@ -0,0 +1,222 @@
...
+                ValkeyTLSLoadError,
+                ValueError,
+            ):
+                logger.error("Failed to disable client TLS")
+                event.defer()
+                return
+
+            self.charm.tls_manager.set_cert_state(is_ready=False)
+            self.charm.tls_manager.set_tls_state(TLSState.NO_TLS)
+            self.charm.unit.open_port("tcp", CLIENT_PORT)
```
skourta, line 178: Let's use `set_ports` instead so we close the `TLS_PORT`. See [here](https://documentation.ubuntu.com/ops/latest/reference/ops/#ops.Unit.set_ports)

skourta, line 178: I think you are missing a return in this block.

src/events/tls.py
```diff
@@ -0,0 +1,222 @@
...
+        if tls_state == TLSState.TO_TLS:
+            try:
+                self._enable_client_tls()
+            except (
+                ValkeyWorkloadCommandError,
+                ValkeyServicesFailedToStartError,
+                ValkeyTLSLoadError,
+                ValueError,
+            ):
+                logger.error("Failed to enable client TLS")
```
skourta, line 135: Let's add a status here.

src/managers/tls.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def _generate_private_key(self) -> PrivateKey:
+        """Generate a private key for use in peer TLS."""
+        return PrivateKey.generate()
+
+    def generate_ca_certificate(self) -> None:
+        """Generate a CA certificate for use in peer TLS and store it to peer relation data."""
+        private_key = self._generate_private_key()
+        ca_attributes = CertificateRequestAttributes(
+            common_name="Valkey Operator",
```
skourta: can we something that looks like a DN and avoid spaces?

tests/integration/k8s/tls/test_tls.py
```diff
@@ -0,0 +1,164 @@
...
+
+async def test_tls_enabled(juju: jubilant.Juju) -> None:
+    """Check if the TLS has been enabled on app startup."""
+    logger.info("Downloading TLS certificates from deployed app.")
+    download_client_certificate_from_unit(juju, APP_NAME)
+
+    hostnames = get_cluster_hostnames(juju, APP_NAME)
+    secret = get_secret_by_label(juju, label=INTERNAL_USERS_SECRET_LABEL)
+    password = secret.get(f"{CharmUsers.VALKEY_ADMIN.value}-password")
+    assert password is not None, "Admin password secret not found"
```
skourta: We have a helper to abstract all of these. `get_password`

tests/integration/k8s/tls/test_tls.py
```diff
@@ -0,0 +1,164 @@
...
+
+NUM_UNITS = 3
+TEST_KEY = "test_key"
+TEST_VALUE = "test_value"
+
+
+def test_build_and_deploy(charm: str, juju: jubilant.Juju) -> None:
+    """Deploy the charm under test and a TLS provider."""
+    juju.deploy(charm, resources=IMAGE_RESOURCE, num_units=NUM_UNITS, trust=True)
+    juju.deploy(TLS_NAME, channel="1/edge")
```
skourta: Let's make the channel a global variable. We have seen changes in the past and it would make it easier to update and maintain.

skourta (commented):
Great work, René. I left some questions.

src/common/client.py
reneradoi: When I implemented this, `config_set` was not a method on the client class and this was the only case where I needed it. Now that we have it, I can indeed switch.

reneradoi: Update: I remember why I chose not to do it: `config_set` is for one parameter and one value, and I like it being this concise and didn't want to make this a list of parameters. For reloading TLS, we need to have all of them set in one command.

src/core/base_workload.py
reneradoi: updated

src/core/models.py
reneradoi: Mostly because I don't like to read `server.model.hostname` if I can read `server.hostname` instead, but that's a personal preference and I have removed the properties now.

src/core/models.py
reneradoi: updated.

pyproject.toml
reneradoi: The released version doesn't have the feature included that we need for using "custom" certificates for Glide TLS connections, therefore we cannot use it. Once a new release is out that has the PR mentioned in it, I will switch this over.

src/managers/cluster.py
reneradoi: Only if necessary, that's why we restart the valkey service here if the health check for valkey fails and the sentinel service in the sentinel manager if the sentinel health check fails.

src/managers/config.py
reneradoi: Sure.

src/managers/sentinel.py
reneradoi: As mentioned, only if required because the health check has failed, not preemptively.

src/events/tls.py
reneradoi: The charm is in the process of starting when this happens, and the users will get displayed a `CONFIGURATION_ERROR`, which I find rather accurate for this.

src/events/tls.py
reneradoi: Yes, we have to come back. This is effectively the synchronisation logic that ensures no unit switches to client TLS before they all are ready for it. Deferring the certificate available event helps us to avoid listening to peer-relation changed events and deferring them.

src/events/tls.py
reneradoi: It's a safeguard if for whatever reason the secret is gone. This is here because the internal secret might not have been updated for a very long time (depending on how long client TLS was enabled), and if for whatever reason unused secrets get cleaned up, we don't want to fail.

src/common/client.py
skourta: I'm ok with both keeping this as is or modifying the config set to take a list.

pyproject.toml
skourta: Understood. We can also install from their git if we want but I'm fine with keeping this as is.

src/events/tls.py
reneradoi: Yes, it is not possible to reload TLS settings for sentinel. This can only be done for a couple of settings, see https://valkey.io/topics/sentinel/#reconfiguring-sentinel-at-runtime.

As for the rolling operation: I don't think we should do this, for two reasons:
1) If we reload Valkey and don't immediately reload Sentinel, it will not be able to connect to Valkey, thus making it error-prone for false-positives on "Valkey down" detection (this is a critical one and effectively forbids delaying the restart)
2) less critical, but obvious: Other sentinels will consider this sentinel "down" only after 30 seconds (or other configurable value), the restart will therefor not impact the failover mechanisms.

src/events/tls.py
skourta: I see your point. Please keep this in mind when implementing cert/CA rotation as we will use the same workflow.

src/events/tls.py
reneradoi: No we can't use `set_ports`, it will close all other ports than the one we define. This would also close the sentinel port, for example. And also, we don't want the `TLS_PORT` to be closed, it has to stay open because we still use it for internal TLS.

src/events/tls.py
skourta: I still do not get it. How will the deferal help us with the sync? We use TLS state for that and it is set before deferring. Deferring here or not does not alter that behaviour. I feel like I am missing something.

src/events/tls.py
reneradoi: No, we don't want to return. We immediately enable internal peer TLS after client TLS was disabled.

skourta: True.

src/events/tls.py
reneradoi: We already have a maintenance status "enabling client TLS", why having another one? If it should be a blocked status: What is the user supposed to do? The idea here is that it is self-healing, and I don't think we should confuse users with flickering statuses in the meantime.

src/events/tls.py
skourta: I see, yes. Maybe we can avoid restarting twice then and only restart when we configure everything.

tests/integration/k8s/tls/test_tls.py
reneradoi: Oh I didn't know that, thank you!

tests/integration/k8s/tls/test_tls.py
reneradoi: Agree, updated.

src/managers/tls.py
reneradoi: I've replaced the space with a dash. Typically we use unit names for the common name in our certs, but as this is not the "regular way" of using TLS, I wanted to distinguish from that.

The common name is packed into a DN by the TLS lib methods then, we don't need to do that on our own.

src/common/client.py
reneradoi: Thank you. I prefer to keep it for now. We can still modify if we feel the need for it.

src/events/tls.py
reneradoi: In `_enable_tls` we check that all units have the `client_cert_ready`, otherwise we raise the `ValkeyCertificatesNotReadyError`. The flag `client_cert_ready` is only being set after a unit has stored the provided certs on its disk. By deferring, we wait until all units have received and stored the certs, and once that is the case, all units reload TLS settings and switch to client TLS at "the same time".

src/events/tls.py
skourta: What you are saying makes sense to me. If the problem does not self heal though we will end up having a charm blocked in a maintenance status with the user not knowing what is happening.

src/events/tls.py
reneradoi: No, unfortunately not. The flow is the following:
- client certs are stored, the operator uses them
- TLS relation broken, operator has to disable TLS using the "old" client certs
- new, internal certs are stored with the same file names
- TLS is being reloaded for Valkey/Sentinel to pick up the new cert files

If we do this in one go, the operator would use the new, internal certs to connect to Valkey, but Valkey would expect the client certs (which are still active at that time). This would fail, therefore we need to decouple the steps but still do it right after another to ensure internal TLS is enabled immediately.

src/events/tls.py
skourta: Oh I see. The deferral is not for the block coming afterwards but for reexecuting `_enable_tls`.

src/events/tls.py
skourta: ACK

src/managers/cluster.py
reneradoi: Currently not needed from my point of view, but thanks for the hint, will keep it in mind for the upcoming PR's.

src/common/client.py
Mehdi-Bendriss: We don't need to lose value of the debug log here, we have to adopt the same approach we had adopted in opensearch for the `exec_cli_command` to have support for attributes that may be secrets - where the logged command has masked attributes

reneradoi: I've already discussed this with Smail, who has refactored the client class and its logging in the scale down PR. Therefore I'd like to keep it as it is here to not duplicate efforts.

src/events/tls.py
```diff
@@ -0,0 +1,213 @@
...
+            self.client_certificate.on.certificate_available, self._on_certificate_available
+        )
+        self.framework.observe(
+            self.charm.on[PEER_RELATION].relation_created, self._on_peer_relation_created
+        )
+
+    def _on_peer_relation_created(self, event: ops.RelationCreatedEvent) -> None:
+        """Set up self-signed certificates for peer TLS by default."""
+        if self.charm.unit.is_leader():
+            self.charm.tls_manager.generate_ca_certificate()
```
Mehdi-Bendriss, line 75: 1. shouldn't we test on whether we have a peer tls_relation or not, to see if it is relevant to generate a self signed CA?
2. why do we do it based on `peer_rel_created`, why not on_start since we have the tls provider check there?

src/events/tls.py
```diff
@@ -0,0 +1,213 @@
...
+            private_key=None,
+            refresh_events=[self.refresh_tls_certificates_event],
+        )
+
+        # --- EVENTS TO OBSERVE ---
+        self.framework.observe(
+            self.charm.on[CLIENT_TLS_RELATION_NAME].relation_created, self._on_tls_relation_created
+        )
+        self.framework.observe(
+            self.charm.on[CLIENT_TLS_RELATION_NAME].relation_broken, self._on_tls_relation_broken
```
Mehdi-Bendriss, line 63: Why are we only handling the events of the peer TLS?

Mehdi-Bendriss (commented):
Thanks René! I have a couple of questions

src/common/client.py
Mehdi-Bendriss: That's okay, my point is this is the approach that needs to be adopted. It can be in a followup PR.

src/events/tls.py
reneradoi: If you mean only handling the events of client TLS: Because we only have a relation interface for client TLS. Peer TLS is internal only and there is no TLS relation for it, as outlined in the spec.

src/events/tls.py
reneradoi: 1) There is no peer TLS relation, peer TLS is internal only. Before generating the self signed certs for this unit, we check whether there are client TLS certs already, in order not to overwrite.
2) We generate the self-signed certificate on `valkey-peers-relation-created` so that it is available when starting a unit, without the need to defer the start event if the Juju leader has not added the CA cert/key to peer relation data yet.

src/events/tls.py
Mehdi-Bendriss: I missed this in the spec, but this flow was not what I had described in our call 😬 what I described was:

- self-managed self-signed-certificate **IF NO** tls peer relation established, expiration infinite
- peer tls certificates from relation if relation established

Hence my insistence on having a unified handling of certificates expiration / rotation and no difference between both self-managed and provider-issued in the code in a comment.
I will confirm with Mykola

Mehdi-Bendriss (approved):

src/events/tls.py
reneradoi: I will keep this in mind as a potential future optimization, when we have gained some experience with integration test runs over time to see if this somehow unstable and needs better status display.

skourta (approved):
Thank you @reneradoi

PR #14, opened 2026-02-26
[DPE-9324]: Scale down

skourta:
This pull request introduces the scale down procedure for valkey on both VM and K8s.It includes refactoring to the Valkey client codebase, including a separation of client logic for Valkey and Sentinel, enhanced error handling, and the addition of new client methods. It also updates the storage configuration for the Valkey container. 

**ScaleDown workflow in storage_detaching:**
- Introduced the scale down workflow in the `storage_detaching` event handler. 

**Client Refactoring and Enhancements:**

- Refactored `ValkeyClient` into a base `CliClient` class, and introduced separate `ValkeyClient` and `SentinelClient` classes for clearer separation of concerns and easier extension. (`src/common/client.py`) [[1]](diffhunk://#diff-f5c8dd5ff5ebd3473ee519d51e59ddb12e43e04365489c1c429256d3d20c3bc4L16-R57) [[2]](diffhunk://#diff-f5c8dd5ff5ebd3473ee519d51e59ddb12e43e04365489c1c429256d3d20c3bc4L59-R103)
- Added new methods to both clients for more granular operations, such as `get`, `set`, `delifeq`, `role`, `acl_load` in `ValkeyClient`, and `get_primary_addr_by_name`, `primary`, `failover_primary_coordinated`, `is_failover_in_progress`, `reset`, `replicas_primary`, and `sentinels_primary` in `SentinelClient`. (`src/common/client.py`) [[1]](diffhunk://#diff-f5c8dd5ff5ebd3473ee519d51e59ddb12e43e04365489c1c429256d3d20c3bc4L101-R214) [[2]](diffhunk://#diff-f5c8dd5ff5ebd3473ee519d51e59ddb12e43e04365489c1c429256d3d20c3bc4R226-L275)
- Improved command execution logic to support JSON output parsing and better error handling, including raising `ValkeyWorkloadCommandError` on failures. (`src/common/client.py`)

**Error Handling Improvements:**

- Introduced several new custom exception classes for more precise error reporting, such as `ValkeyCannotGetPrimaryIPError`, `SentinelFailoverError`, `ValkeyServicesCouldNotBeStoppedError`, `CannotSeeAllActiveSentinelsError`, `SentinelIncorrectReplicaCountError`, and `RequestingLockTimedOutError`. (`src/common/exceptions.py`) [[1]](diffhunk://#diff-e793bad618b16c64102391f92909ac368210513be36cc7ff3ea633e458e54feeR23-R26) [[2]](diffhunk://#diff-e793bad618b16c64102391f92909ac368210513be36cc7ff3ea633e458e54feeR41-R60)

**Configuration and Storage Updates:**

- Added a `storage` section to the `metadata.yaml` file for the Valkey container, specifying a persistent data mount and minimum storage requirements. (`metadata.yaml`) [[1]](diffhunk://#diff-d186204ea7abb85bbadd9d94ab29d5bcefcc43d6a0ed06015e98a71b2f099423R25-R27) [[2]](diffhunk://#diff-d186204ea7abb85bbadd9d94ab29d5bcefcc43d6a0ed06015e98a71b2f099423R40-R46)

These changes collectively improve the structure, reliability, and maintainability of the Valkey charm's client and configuration logic.


src/common/locks.py
```diff
@@ -0,0 +1,270 @@
...
+            return
+
+        if self.is_lock_free_to_give:
+            next_unit = self.next_unit_to_give_lock
+            self.state.cluster.update({self.member_with_lock_atr_name: next_unit})
+            logger.debug(f"Gave {self.name} lock to {next_unit}")
+        logger.debug(
+            f"{self.name} lock is currently held by {getattr(self.state.cluster.model, self.member_with_lock_atr_name)}"
+        )
+
```
reneradoi: We should only log these messages if there is actually a unit that holds the lock, otherwise we get this:
```
unit-valkey-1: 12:19:54 DEBUG unit.valkey/1.juju-log valkey-peers:1: Gave startlock lock to None
unit-valkey-1: 12:19:54 DEBUG unit.valkey/1.juju-log valkey-peers:1: startlock lock is currently held by 
```

src/common/locks.py
```diff
@@ -0,0 +1,270 @@
...
+    def request_lock(self) -> bool:
+        """Request the lock for the local unit."""
+        self.state.unit_server.update(
+            {
+                self.unit_request_lock_atr_name: True,
+            }
+        )
+        if self.state.unit_server.unit.is_leader():
+            logger.info(
+                f"Leader unit requesting {self.name} lock. Triggering lock request processing."
```
reneradoi: The `lock` is duplicate in this log message. Please also check other log messages for `{self.name} lock`, which would have the same result.

src/common/locks.py
```diff
@@ -0,0 +1,270 @@
...
+
+        Args:
+            timeout (int | None): The maximum time to keep trying to acquire the lock, in seconds. If None, it will keep trying indefinitely.
+
+        Returns:
+            bool: True if the lock was acquired, False if the timeout was reached before acquiring the lock.
+        """
+        logger.debug(f"{self.charm.state.unit_server.unit_name} is requesting {self.name} lock.")
+        retry_until = time.time() + timeout if timeout else None
+        primary_ip = self.charm.sentinel_manager.get_primary_ip()
```
reneradoi: If we fail to get a primary ip, this will raise and the hook will fail. The removal will then be stuck. I was able to provoke this by having three units, first removing one and then removing the remaining two. It might also happen if users want to remove units that are down.

This is basically the downside of acquiring a database lock: We can only get it when the database is online. That means we need to add safeguards, or rather: what should happen if the database is not available for the lock? In case of Opensearch, the fallback is to use the peer relation for locking. I wouldn't do that for Valkey, I think if we cannot get a lock for scaling down because the database is offline, we might just continue.

If the database lock will ever be used for something else, this might be different though depending on the use case.

src/events/base_events.py
```diff
@@ -89,7 +97,14 @@ def _on_install(self, event: ops.InstallEvent) -> None:
 
     def _on_start(self, event: ops.StartEvent) -> None:
         """Handle the on start event."""
-        self.charm.state.unit_server.update({"start_state": StartState.NOT_STARTED.value})
+        self.charm.state.unit_server.update(
+            {
+                "start_state": StartState.NOT_STARTED.value,
+                "hostname": socket.gethostname(),
+                "private_ip": self.charm.state.bind_address,
+            }
```
reneradoi: Why updating these? They are set by the previous `config-changed` event.

src/events/base_events.py
```diff
@@ -103,23 +118,16 @@ def _on_start(self, event: ops.StartEvent) -> None:
...
             event.defer()
             return
-
-        if not (primary_ip := self.charm.sentinel_manager.get_primary_ip()):
+        try:
+            primary_ip = self.charm.sentinel_manager.get_primary_ip()
+        except ValkeyCannotGetPrimaryIPError:
             if self.charm.state.number_units_started == 0 and self.charm.unit.is_leader():
                 primary_ip = self.charm.state.bind_address
             else:
```
reneradoi, line 144: Should we also release the lock in this case?

src/events/base_events.py
```diff
@@ -210,46 +214,8 @@ def _on_peer_relation_changed(self, event: ops.RelationChangedEvent) -> None:
...
-            return
-
-        if units_requesting_start:
-            self.charm.state.cluster.update({"starting_member": units_requesting_start[0]})
-            logger.debug(f"Updated starting member to {units_requesting_start[0]}")
-            return
-
-        if self.charm.state.cluster.model.starting_member:
-            self.charm.state.cluster.update({"starting_member": ""})
+        for lock in [StartLock(self.charm.state)]:
```
reneradoi, line 233: Which other locks are expected to be added here?

src/events/base_events.py
```diff
@@ -301,11 +267,6 @@ def _on_leader_elected(self, event: ops.LeaderElectedEvent) -> None:
         )
         # update local unit admin password
         self.charm.config_manager.update_local_valkey_admin_password()
-        try:
-            self.charm.config_manager.set_acl_file()
-        except ValkeyWorkloadCommandError:
-            logger.error("Failed to write acl file")
-            raise
```
reneradoi, line 287: Why was setting the acl file removed?

src/literals.py
```diff
@@ -83,3 +85,15 @@ class StartState(StrEnum):
...
     STARTED = "started"
+
+
+class ScaleDownState(StrEnum):
+    """Scale down states for the service."""
+
+    NO_SCALE_DOWN = ""
+    WAIT_FOR_LOCK = "wait_for_lock"
+    WAIT_TO_FAILOVER = "wait_to_failover"
+    STOP_SERVICES = "stopped_services"
```
reneradoi: This should be "stopping" because we set it before stopping.

src/events/base_events.py
```diff
@@ -451,3 +416,71 @@ def _update_internal_users_password(self, secret_id: str) -> None:
...
+        # TODO consider quorum when removing unit
+
+        self.charm.status.set_running_status(
+            ScaleDownStatuses.SCALING_DOWN.value,
+            scope="unit",
+            component_name=self.charm.cluster_manager.name,
+            statuses_state=self.charm.state.statuses,
+        )
+        # if unit has primary then failover
+        primary_ip = self.charm.sentinel_manager.get_primary_ip()
```
reneradoi: Same here: If we can't get the primary ip for whatever reason, this would raise and removal would be stuck here. We currently cannot get to this code in this case, because it already raises on `request_lock`, but we need safeguards here, too.

src/common/client.py
```diff
@@ -13,52 +16,91 @@
 logger = logging.getLogger(__name__)
 
 
-class ValkeyClient:
+class CliClient:
     """Handle valkey client connections."""
 
+    port: int = CLIENT_PORT
```
reneradoi, line 22: Heads up: this will need adjustments, as with the TLS PR, we distinguish between `CLIENT_PORT` and `TLS_PORT` depending on the TLS state.

src/common/client.py
```diff
@@ -166,110 +227,173 @@ def config_set(self, hostname: str, parameter: str, value: str) -> bool:
...
+    def __init__(
+        self,
+        username: str,
+        password: str,
+        workload: WorkloadBase,
+    ):
+        super().__init__(username, password, workload)
+
+    def ping(self, hostname: str) -> bool:
+        """Ping the Valkey server to check if it's responsive.
```
reneradoi: "Sentinel" instead of "Valkey".

src/common/locks.py
```diff
@@ -0,0 +1,270 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Collection of lock names for cluster operations."""
```
reneradoi: This is an awesome addition to the code. I particularly like how clean and easily readable it is, great work!

src/common/locks.py
```diff
@@ -0,0 +1,270 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def release_lock(self) -> bool:
+        """Release the lock from the local unit."""
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def do_i_hold_lock(self) -> bool:
```
reneradoi: Can we rename this to something like `held_by_this_unit`?

src/common/locks.py
```diff
@@ -0,0 +1,270 @@
...
+        )
+
+
+class ScaleDownLock(Lockable):
+    """Lock for scale down operations.
+
+    This will use valkey to store the lock state and will check if the unit with the lock has completed its scale down operation
+    """
+
+    lock_key = "scale_down_lock"
```
reneradoi: I think we should include the application name in the key for the lock. Reasons:
- it reduces the chance that a client will attempt to write this key
- it allows for future Valkey clusters to independently set their lock per application (though this is not 100% accurate as hash slot distribution has to be factored in, but we should at least prepare for that)

src/workload_k8s.py
```diff
@@ -127,3 +128,13 @@ def exec(self, command: list[str]) -> tuple[str, str | None]:
...
             logger.error("Command failed with %s, %s", e.exit_code, e.stdout)
             raise ValkeyWorkloadCommandError(e)
+
+    @override
+    def stop(self) -> None:
+        try:
+            self.container.stop(self.valkey_service, self.sentinel_service, self.metric_service)
+        except ops.pebble.ChangeError as e:
+            logger.error("Failed to stop Valkey services: %s", e)
+            raise ValkeyServicesCouldNotBeStoppedError(
```
reneradoi, line 157: Maybe also consider the other errors that may be raised by Pebble: `TimeoutError, ConnectionError, APIError`.

src/core/models.py
```diff
@@ -75,20 +77,21 @@ def update(self, items: dict[str, Any]) -> None:
...
 
-        self.data_interface.write_model(self.relation.id, model)
+        self.data_interface.write_model(self.relation.id, self.model)
 
 
 @final
 class ValkeyServer(RelationState):
     """State/Relation data collection for a unit."""
 
+    model: PeerUnitModel
```
reneradoi, line 116: Is this the change that has improved the speed of relation data lookups?

src/core/models.py
```diff
@@ -120,6 +118,20 @@ def is_started(self) -> bool:
         """Check if the unit has started."""
         return self.model.start_state == StartState.STARTED.value if self.model else False
 
+    @property
+    def is_being_removed(self) -> bool:
+        """Check if the unit is being removed from the cluster."""
+        return self.model.scale_down_state not in {
```
reneradoi: To make it easier to understand: Let's make this a whitelist of `ScaleDownStates` that mean a unit is being removed.

src/managers/cluster.py
```diff
@@ -79,7 +79,8 @@ def is_replica_synced(self) -> bool:
             password=self.admin_password,
             workload=self.workload,
         )
-        return client.is_replica_synced(hostname=self.state.bind_address)
+        role_info = client.role(hostname=self.state.bind_address)
+        return role_info[0] == "slave" and role_info[3] == "connected"
```
reneradoi: Could this raise a KeyError if executed on a primary?

src/managers/sentinel.py
```diff
@@ -43,86 +47,264 @@ def admin_password(self) -> str:
...
-        if not client.sentinel_get_master_info(hostname=self.state.bind_address):
+        try:
+            client.primary(hostname=self.state.bind_address)
+        except ValkeyWorkloadCommandError:
             logger.warning("Health check failed: Could not query sentinel for master information.")
             return False
 
         return True
 
+    def failover(self) -> None:
```
reneradoi, line 148: General comment on the new methods added to the sentinel manager, not specific to this one: they need some unit test coverage. This file currently only has 40% coverage. This is probably due to patching out all the sentinel manager methods, but it would also be possible to mock the client's return results, esp. as these are cli calls. It can be done for multiple calls to the same method with something like this:

```
with patch (..., side_effect = [None, "something else"]):
```

which will give `None` on the first call of a method and "something else" on the 2nd call.

tests/integration/ha/test_scaling.py
```diff
@@ -89,3 +93,222 @@ async def test_scale_up(juju: jubilant.Juju, c_writes, c_writes_runner) -> None:
         username=CharmUsers.VALKEY_ADMIN.value,
         password=get_password(juju, user=CharmUsers.VALKEY_ADMIN),
     )
+
+
+async def test_scale_down_one_unit(juju: jubilant.Juju, substrate: Substrate, c_writes) -> None:
+    """Make sure scale down operations complete successfully."""
+    number_of_slaves = await get_number_connected_slaves(
```
reneradoi: Let's rename `slave` to `replica` on any occurrence.

tests/integration/ha/test_scaling.py
```diff
@@ -89,3 +93,222 @@ async def test_scale_up(juju: jubilant.Juju, c_writes, c_writes_runner) -> None:
...
+    juju.remove_unit(primary_unit)
+    juju.wait(
+        lambda status: are_apps_active_and_agents_idle(
+            status, APP_NAME, unit_count=NUM_UNITS - 1, idle_period=10
+        )
+    )
+    c_writes.update()
+    new_primary_ip = get_primary_ip(juju, APP_NAME)
+    assert new_primary_ip != primary_ip, "Primary IP did not change after removing primary unit."
+    logger.debug(f"New primary IP after scale down is {new_primary_ip}.")
```
reneradoi: Please always log with log level `info` in integration tests, so that we have all data to debug failed runs without re-running with different log level.

tests/integration/ha/test_scaling.py
```diff
@@ -89,3 +93,222 @@ async def test_scale_up(juju: jubilant.Juju, c_writes, c_writes_runner) -> None:
...
+    logger.debug(
+        f"Identified primary unit {primary_unit} with IP {primary_ip} for scale down test."
+    )
+    juju.remove_unit(primary_unit)
+    juju.wait(
+        lambda status: are_apps_active_and_agents_idle(
+            status, APP_NAME, unit_count=NUM_UNITS - 1, idle_period=10
+        )
+    )
+    c_writes.update()
```
reneradoi, line 333: Why do we have to `update` here if `update` is already done in the loop of continuous writes itself?

tests/unit/test_scaledown.py
```diff
@@ -0,0 +1,136 @@
...
+        storages={data_storage},
+    )
+
+    with (
+        patch("common.locks.ScaleDownLock.request_lock", return_value=False),
+    ):
+        # expect raised exception due to lock not being acquired
+        with pytest.raises(testing.errors.UncaughtCharmError) as exc_info:
+            ctx.run(ctx.on.storage_detaching(data_storage), state_in)
+        assert "RequestingLockTimedOutError" in str(exc_info.value)
```
reneradoi, line 64: Very nice!

reneradoi (commented):
Hey Smail! This PR is already in pretty good shape. I like how the code is structured, and it looks also very clean. The main concern that I have is about how the scale down will look like if the unit's workloads are not online anymore. I've also left a few questions for my understanding.

src/common/locks.py
skourta: We cannot fall back on the peer relation since we cannot defer the `storage_detaching` event. We risk losing data in case the database is offline and we just conitnue and remove the units. Let's discuss this in the office hours in depth.

src/events/base_events.py
skourta: This might not be necessary when we move to hostnames for k8s (next PR). When pods are restarted the IPs are changed in a lot of cases. This essentially picks up the new ip whenever we get a start event.

src/events/base_events.py
skourta: Yes, we can. The outcome will essentially be the same with some additional `relation_changed` events. I will add it.

src/events/base_events.py
skourta: Any other databag locks we might add in the future. This will probably be removed when we move to advanced rolling ops.

src/events/base_events.py
skourta: We do not need it. We always write it again on the start event.

src/common/locks.py
skourta: Renamed to `is_held_by_this_unit`

src/core/models.py
skourta: Yes, this is it. We only build the model once now and rebuild it after every write in the update method.

src/core/models.py
skourta: Switched logic.

src/managers/cluster.py
skourta: Yes, but we only call it on replicas.

tests/integration/ha/test_scaling.py
skourta: We trigger a config recompute on the continuous writes runner using this update. We only update the config if we get some data in the queue.

src/common/client.py
skourta: I added a TODO comment so I do not forget about it

src/common/locks.py
skourta: Added. Thank you for detailing the reasons I had not thought of this.

src/workload_k8s.py
skourta: Added. I reaed the doctsring and had an initial look at the code and only added `ChangeError`. But you are right, upon closer inspection of the code they do raise other errors.

src/events/base_events.py
```diff
@@ -464,3 +431,106 @@ def _update_internal_users_password(self, secret_id: str) -> None:
...
```
reneradoi: We did already get the primary ip earlier, why trying again (especially since the failover is just about to happen a few lines below)?

src/events/base_events.py
```diff
@@ -464,3 +431,106 @@ def _update_internal_users_password(self, secret_id: str) -> None:
...
+                {
+                    "internal_ca_certificate": None,
+                    "internal_ca_private_key": None,
+                }
+            )
+
+        self.charm.state.unit_server.update({"scale_down_state": ScaleDownState.GOING_AWAY})
+
+    @tenacity.retry(wait=tenacity.wait_fixed(5), stop=tenacity.stop_after_attempt(8), reraise=True)
+    def _get_primary_ip_for_scale_down(self) -> str:
```
reneradoi: Let's move this to the sentinel manager.

Reasoning: We might have a situation in the future where we don't use Sentinel but operate Valkey in Cluster mode. My plan for then (currently only living in my head...) is to replace the hardcoded use of Sentinel Manager with a HA Manager, a base class which will either be instantiated as Sentinel Manager or "Cluster mode" Manager. Therefor everything related to Sentinel should be living in the Sentinel Manager exclusively, to avoid coupling Sentinel too tight so that we cannot replace it then.

src/managers/sentinel.py
reneradoi: Awesome unit test coverage now, that helps a lot, thank you!

reneradoi (commented):
Only minor comments remaining, otherwise good to go!

src/events/tls.py
```diff
@@ -153,7 +153,7 @@ def _on_tls_relation_broken(self, event: ops.RelationBrokenEvent) -> None:
                 event.defer()
                 return
 
-        if self.charm.state.unit_server.tls_client_state in [TLSState.TLS, TLSState.TO_NO_TLS]:
+        if self.charm.state.unit_server.is_tls_enabled:
```
reneradoi, line 225: I found an issue with removing units when client TLS is enabled. We have to add a check + early return in this hook to ensure we don't fail and error the hook when disabling client TLS. This can be added to the check for planned units some lines above, which I already added in anticipation of this.

```
if self.charm.app.planned_units() == 0 or self.charm.state.unit_server.is_being_removed:
    return
```

src/events/base_events.py
skourta: In case the lock to request the lock takes long and a failover happens in between.

src/events/base_events.py
skourta: Agree. I should add the check on all get primary IPs.

src/events/base_events.py
skourta: Will do. Thank you for anticipating that.

src/events/base_events.py
reneradoi: Ok makes sense, thank you!

src/events/tls.py
skourta: Added.

reneradoi (approved):
Great work, thank you Smail!

src/common/locks.py
```diff
@@ -0,0 +1,272 @@
...
+                ):
+                    logger.debug(
+                        f"{self.charm.state.unit_server.unit_name} acquired {self.name} lock."
+                    )
+                    return True
+            except ValkeyWorkloadCommandError:
+                logger.warning(
+                    f"{self.charm.state.unit_server.unit_name} failed to acquire {self.name} lock due to a workload command error. Retrying..."
+                )
+            if retry_until and time.time() > retry_until:
```
Mehdi-Bendriss: can you replace the while True with this condition?

src/common/locks.py
```diff
@@ -0,0 +1,272 @@
...
+                logger.warning(
+                    f"{self.charm.state.unit_server.unit_name} failed to acquire {self.name} lock within timeout. Giving up."
+                )
+                return False
+            logger.info(
+                f"{self.charm.state.unit_server.unit_name} failed to acquire {self.name} lock. Retrying in 5 seconds."
+            )
+            time.sleep(5)
```
Mehdi-Bendriss: why not handle this whole retry logic with a tenacity wait_fixed(5)?

src/common/locks.py
```diff
@@ -0,0 +1,272 @@
...
+        primary_ip = primary_ip or self.charm.sentinel_manager.get_primary_ip()
+        if (
+            self.client.delifeq(
+                hostname=primary_ip,
+                key=self.lock_key,
+                value=self.charm.state.unit_server.unit_name,
+            )
+            == "1"
+        ):
+            logger.debug(f"{self.charm.state.unit_server.unit_name} released {self.name} lock.")
```
Mehdi-Bendriss: across the board, we need to use lazy formatting in log lines, this is best practice in python, but also increases the TIOBE score

src/common/locks.py
```diff
@@ -0,0 +1,272 @@
...
+            return True
+        else:
+            logger.warning(
+                f"{self.charm.state.unit_server.unit_name} failed to release {self.name} lock. It may not have held the lock or it may have already been released."
+            )
+            return False
```
Mehdi-Bendriss:
```suggestion
        
        logger.warning(
            f"{self.charm.state.unit_server.unit_name} failed to release {self.name} lock. It may not have held the lock or it may have already been released."
        )
        return False
```

src/events/base_events.py
```diff
@@ -464,3 +433,105 @@ def _update_internal_users_password(self, secret_id: str) -> None:
             scope="app",
             component=self.charm.cluster_manager.name,
         )
+
+    def _on_storage_detaching(self, event: ops.StorageDetachingEvent) -> None:
+        """Handle removal of the data storage mount, e.g. when removing a unit."""
+        # get scale down lock
+        scale_down_lock = ScaleDownLock(self.charm)
+
+        self.charm.state.unit_server.update({"scale_down_state": ScaleDownState.WAIT_FOR_LOCK})
```
Mehdi-Bendriss: i'm not sure I understand the use for this, it will not be committed on the peer rel databag until this hook completes, and by the time it completes, it won't be in a `WAIT_FOR_LOCK` state, no?
(The scaledown lock is synchronous)

src/events/base_events.py
```diff
@@ -464,3 +433,105 @@ def _update_internal_users_password(self, secret_id: str) -> None:
...
+        except ValkeyCannotGetPrimaryIPError as e:
+            logger.error(e)
+            self._set_state_for_going_away()
+            return
+
+        # blocks until the lock is acquired
+        if not scale_down_lock.request_lock(primary_ip=primary_ip):
+            raise RequestingLockTimedOutError("Failed to acquire scale down lock within timeout")
+
+        self.charm.state.statuses.delete(
```
Mehdi-Bendriss, line 460: similar comment

src/events/base_events.py
```diff
@@ -464,3 +433,105 @@ def _update_internal_users_password(self, secret_id: str) -> None:
...
+        active_sentinels = self.charm.sentinel_manager.get_active_sentinel_ips(primary_ip)
+        if (
+            primary_ip == self.charm.state.unit_server.get_endpoint(self.charm.state.substrate)
+            and len(active_sentinels) > 1
+        ):
+            self.charm.state.unit_server.update(
+                {"scale_down_state": ScaleDownState.WAIT_TO_FAILOVER}
+            )
+            logger.debug("Triggering sentinel failover on primary IP %s", primary_ip)
+            self.charm.sentinel_manager.failover()
```
Mehdi-Bendriss, line 487: exception handling?

src/events/base_events.py
```diff
@@ -464,3 +433,105 @@ def _update_internal_users_password(self, secret_id: str) -> None:
...
+            primary_ip = self.charm.sentinel_manager.get_primary_ip()
+            logger.debug(
+                "Failover completed, new primary ip %s",
+                primary_ip,
+            )
+
+        # stop valkey and sentinel processes
+        self.charm.state.unit_server.update({"scale_down_state": ScaleDownState.STOP_SERVICES})
```
Mehdi-Bendriss: similar question on state committing

src/managers/cluster.py
```diff
@@ -119,38 +122,45 @@ def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObje
...
-                    StartStatuses.ERROR_ON_START.value,
-                )
+                return StartStatuses.ERROR_ON_START.value
 
-        return status_list or [CharmStatuses.ACTIVE_IDLE.value]
+        return None
+
+    def _get_scale_down_status(self) -> StatusObject | None:
+        """Get the current scale down status of the unit."""
+        match self.state.unit_server.model.scale_down_state:
```
Mehdi-Bendriss: better replace with an `if`

src/managers/sentinel.py
```diff
@@ -41,85 +46,239 @@ def admin_password(self) -> str:
...
-            else False,
+            tls=self.state.unit_server.is_tls_enabled,
             workload=self.workload,
-            connect_to="sentinel",
         )
 
     @retry(
         wait=wait_fixed(5),
-        stop=stop_after_attempt(5),
+        stop=stop_after_attempt(6),
```
Mehdi-Bendriss, line 60: why specifically 6?

src/managers/sentinel.py
```diff
@@ -41,85 +46,239 @@ def admin_password(self) -> str:
...
         retry=retry_if_result(lambda result: result is False),
         retry_error_callback=lambda _: False,
     )
     def is_sentinel_discovered(self) -> bool:
         """Check if the sentinel of the local unit was discovered by the other sentinels in the cluster."""
-        # list of active sentinels: units with started flag true
+        # list of active sentinels: units with started flag true and not being removed
```
Mehdi-Bendriss, line 66: can you explain the use of "not being removed" in this particular context?

src/managers/sentinel.py
```diff
@@ -41,85 +46,239 @@ def admin_password(self) -> str:
...
         for sentinel_ip in active_sentinels:
             try:
-                output, _ = client.exec_cli_command(
-                    command=["sentinel", "sentinels", PRIMARY_NAME],
-                    hostname=sentinel_ip,
-                )
-                if self.state.unit_server.model.private_ip not in output:
-                    logger.info(f"Sentinel at {sentinel_ip} has not discovered this sentinel")
+                discovered_sentinels = {
+                    sentinel["ip"] for sentinel in client.sentinels_primary(hostname=sentinel_ip)
```
Mehdi-Bendriss: what does `sentinel["ip"]` in k8s?

src/managers/sentinel.py
```diff
@@ -41,85 +46,239 @@ def admin_password(self) -> str:
...
+            unit.get_endpoint(self.state.substrate)
             for unit in self.state.servers
-            if unit.is_started and unit.model.private_ip != self.state.unit_server.model.private_ip
+            if unit.is_active and unit.get_endpoint(self.state.substrate) != self.state.endpoint
         ]
 
-        client = self._get_valkey_client()
+        client = self._get_sentinel_client()
 
         for sentinel_ip in active_sentinels:
```
Mehdi-Bendriss:
```suggestion
        for sentinel_host in active_sentinels:
```

src/managers/sentinel.py
```diff
@@ -41,85 +46,239 @@ def admin_password(self) -> str:
...
+
             except ValkeyWorkloadCommandError:
                 logger.warning(f"Could not query sentinel at {sentinel_ip} for primary discovery.")
                 return False
         return True
 
-    def get_primary_ip(self) -> str | None:
-        """Get the IP address of the primary node in the cluster."""
-        started_servers = [unit for unit in self.state.servers if unit.is_started]
+    def get_primary_ip(self) -> str:
```
Mehdi-Bendriss, line 95: why are we relying on the IP for both both vm and k8s? (same for `get_primary_ip_for_scale_down`)

src/managers/sentinel.py
```diff
@@ -41,85 +46,239 @@ def admin_password(self) -> str:
...
+
+        This method triggers a failover through the sentinel client and then checks if the failover is in progress.
+
+        Raises:
+            SentinelFailoverError: If triggering failover fails or if failover does not start after triggering.
+        """
+        client = self._get_sentinel_client()
+        try:
+            client.failover_primary_coordinated(self.state.endpoint)
+            client.is_failover_in_progress(self.state.endpoint)
```
Mehdi-Bendriss: in the case of `NOT "failover_in_progress" in self.primary(hostname=hostname).get("flags", "")"` and no `ValkeyWorkloadCommandError` raised, this will just pass smoothly

src/managers/sentinel.py
```diff
@@ -41,85 +46,239 @@ def admin_password(self) -> str:
...
+        except ValkeyWorkloadCommandError as e:
+            logger.error(f"Failed to trigger failover: {e}")
+            raise SentinelFailoverError from e
+
+    def reset_sentinel_states(self, sentinel_ips: list[str]) -> None:
```
Mehdi-Bendriss, line 165: reference to IPs

src/managers/sentinel.py
```diff
@@ -41,85 +46,239 @@ def admin_password(self) -> str:
...
+            expected_replicas,
+            sentinel_ips,
+        )
+
+        for sentinel_ip in sentinel_ips:
+            if expected_replicas != (
+                number_replicas := len(client.replicas_primary(hostname=sentinel_ip))
+            ):
+                logger.warning(
+                    f"Sentinel at {sentinel_ip} sees {number_replicas} replicas, expected {expected_replicas}."
```
Mehdi-Bendriss: similar command, move to printf style syntax for logging

src/workload_vm.py
```diff
@@ -184,3 +185,19 @@ def wait_for_services_to_be_alive(self, duration: float = 30, delay: float = 0.1
...
+            self.valkey.stop(services=[SNAP_SERVICE, SNAP_SENTINEL_SERVICE])
+        except snap.SnapError as e:
+            logger.error("Failed to stop Valkey services: %s", e)
+            raise ValkeyServicesCouldNotBeStoppedError(
+                f"Failed to stop Valkey services: {e}"
+            ) from e
+
+        if self.alive():
+            logger.error("Valkey services are still alive after stop.")
+            raise ValkeyServicesCouldNotBeStoppedError(
```
Mehdi-Bendriss, line 201: is this caught somewhere? maybe we add a retry before giving up on it?

tests/integration/ha/test_scaling.py
```diff
@@ -68,13 +74,13 @@ async def test_scale_up(juju: jubilant.Juju, c_writes, c_writes_runner) -> None:
...
         username=CharmUsers.VALKEY_ADMIN.value,
         password=get_password(juju, user=CharmUsers.VALKEY_ADMIN),
     )
-    assert connected_slaves == NUM_UNITS - 1, (
-        f"Expected {NUM_UNITS - 1} connected slaves, got {connected_slaves}."
+    assert connected_replicas == NUM_UNITS - 1, (
+        f"Expected {NUM_UNITS - 1} connected replicas, got {connected_replicas}."
     )
 
     await assert_continuous_writes_increasing(
```
Mehdi-Bendriss, line 93: we can check on the consistency in non disaster tests

tests/integration/ha/test_scaling.py
Mehdi-Bendriss, line 1: these tests need to account for a passed `app` like in opensearch. So they can be ran against an existing cluster. 
check the safeguards we also add when scaling down (check on pre-existing number of units etc.)

tests/integration/ha/test_scaling.py
```diff
@@ -89,3 +95,225 @@ async def test_scale_up(juju: jubilant.Juju, c_writes, c_writes_runner) -> None:
...
+
+    assert_continuous_writes_consistent(
+        hostnames=get_cluster_hostnames(juju, APP_NAME),
+        username=CharmUsers.VALKEY_ADMIN.value,
+        password=get_password(juju, user=CharmUsers.VALKEY_ADMIN),
+    )
+    await c_writes.async_clear()
+
+
+async def test_scale_down_to_zero_and_back(
```
Mehdi-Bendriss:
```suggestion
async def test_scale_down_to_zero_and_back_up(
```

tests/integration/ha/test_scaling.py
```diff
@@ -89,3 +95,225 @@ async def test_scale_up(juju: jubilant.Juju, c_writes, c_writes_runner) -> None:
...
+        f"Identified primary unit {primary_unit} with IP {primary_ip} for scale down test."
+    )
+    juju.remove_unit(primary_unit)
+    juju.wait(
+        lambda status: are_apps_active_and_agents_idle(
+            status, APP_NAME, unit_count=NUM_UNITS - 1, idle_period=10
+        )
+    )
+    c_writes.update()
+    new_primary_ip = get_primary_ip(juju, APP_NAME)
```
Mehdi-Bendriss: can we use `_host` or `_address` instead of `_ip`

tests/integration/helpers.py
```diff
@@ -367,19 +369,27 @@ def download_client_certificate_from_unit(juju: jubilant.Juju, app_name: str = A
...
+        username=CharmUsers.VALKEY_ADMIN.value,
+        password=get_password(juju),
+        command="info replication",
+    ).stdout
+    # if master then we return the hostname
+    if "role:master" in replication_info:
+        return hostnames[0]
+    # extract ip
+    match = re.search(r"master_host:([^\s]+)", replication_info)
+    if not match:
```
Mehdi-Bendriss: walrus

Mehdi-Bendriss (commented):
Thanks Smail! Great work! I left some comments
Can you revisit the occurrences of ip vs hostnames in tests and across the charm?

src/common/locks.py
skourta: Because we want the timeout configurable.

src/events/base_events.py
skourta: Yes, correct. We do not write it in the databag. I will clean gthe states we do not need.

src/events/base_events.py
skourta: Can you clarify? This is a status not a state. We set and clear running statuses because of the sync nature of this hook

src/events/base_events.py
skourta: What would be the strategy in case it happens? If it raises the hook will be retried and it can self heal. If we catch the exception then we will have to unsafely delete the unit.

src/common/locks.py
skourta: Switched the logic to use tenacity programmatically

src/managers/sentinel.py
skourta: To adhere to the 30s from the docs.

src/managers/sentinel.py
skourta: We do not want to consider sentinels that are being removed. They will be stoped at that time.

src/managers/sentinel.py
skourta: The hostname because we use `announce-ip` and set it to the hostname so they only use hostnames.

src/managers/sentinel.py
skourta: It is not the IP in case of k8s it will be the hostname. See the previous comment.

src/managers/sentinel.py
skourta: Can you please clarify?

src/workload_vm.py
skourta: The `alive` function does retry 3 times

tests/integration/ha/test_scaling.py
skourta: We do check right below.

tests/integration/ha/test_scaling.py
```diff
@@ -89,3 +95,225 @@ async def test_scale_up(juju: jubilant.Juju, c_writes, c_writes_runner) -> None:
...
+        hostnames=hostnames,
+        username=CharmUsers.VALKEY_ADMIN.value,
+        password=get_password(juju, user=CharmUsers.VALKEY_ADMIN),
+    )
+    await c_writes.async_clear()
+
+
+def test_scale_down_remove_application(juju: jubilant.Juju) -> None:
+    """Make sure the application can be removed."""
+    juju.remove_application(APP_NAME)
```
skourta, line 357: This might delete the existing app which can be risky.

src/events/base_events.py
Mehdi-Bendriss: i misread state* instead of status*

src/managers/sentinel.py
Mehdi-Bendriss: i'm referring to the argument of the method: `sentinel_ips`

tests/integration/ha/test_scaling.py
Mehdi-Bendriss: yeah only non-destructive tests should run against an `app`, destructive ones should always only apply to APP_NAME

src/common/locks.py
```diff
@@ -225,24 +244,9 @@ def request_lock(self, timeout: int | None = None, primary_ip: str | None = None
...
-                logger.warning(
-                    f"{self.charm.state.unit_server.unit_name} failed to acquire {self.name} lock within timeout. Giving up."
-                )
-                return False
-            logger.info(
-                f"{self.charm.state.unit_server.unit_name} failed to acquire {self.name} lock. Retrying in 5 seconds."
-            )
-            time.sleep(5)
-            # update the primary ip in case a failover happens when we are waiting to acquire the lock
-            primary_ip = self.charm.sentinel_manager.get_primary_ip()
```
Mehdi-Bendriss: missing return False at the end?

src/events/base_events.py
Mehdi-Bendriss: you're right, i suppose this is better than letting the unit depart without notice or user intervention

src/managers/sentinel.py
Mehdi-Bendriss: i'm referring to the name of the method

tests/integration/ha/test_scaling.py
```diff
@@ -100,52 +107,67 @@ async def test_scale_up(juju: jubilant.Juju, c_writes) -> None:
 
 async def test_scale_down_one_unit(juju: jubilant.Juju, substrate: Substrate, c_writes) -> None:
     """Make sure scale down operations complete successfully."""
+    app_name = existing_app(juju) or APP_NAME
+    init_units_count = len(juju.status().apps[app_name].units)
+
+    if init_units_count < 1:
```
Mehdi-Bendriss: since there is a substraction below, shouldn't this be `< NUM_UNITS`?

Mehdi-Bendriss (commented):
Thanks Smail, good work! Only minor comments left for me

src/managers/sentinel.py
reneradoi: I stumbled over that one, too. But considering the naming of config parameters in Sentinel and Valkey, I think it is fine to keep the name `ip` in the occurrences we currently have, even though we might use hostnames in the values instead.

src/managers/sentinel.py
reneradoi: I'll also leave it here: I stumbled over that one, too. But considering the naming of config parameters in Sentinel and Valkey, I think it is fine to keep the name ip in the occurrences we currently have, even though we might use hostnames in the values instead.

skourta: We can switch everything to `endpoint` Valkey will still use ip for everything though. For config-related functions I belive having ip instead of endpoint makes more sense.

Mehdi-Bendriss (approved):

PR #17, opened 2026-03-06
Bugfix: Use dedicated TLS port for Sentinel

skourta:
Currently, sentinel cannot communicate with other sentinels when the TLS relation is not established:
```json
[
  {
    "name": "806702609387275df560cf8948545df364162177",
    "ip": "10.1.153.187",
    "port": "26379",
    "runid": "806702609387275df560cf8948545df364162177",
    "flags": "s_down,sentinel",
    "link-pending-commands": "41",
    "link-refcount": "1",
    "last-ping-sent": "319719",
    "last-ok-ping-reply": "319719",
    "last-ping-reply": "319719",
    "s-down-time": "289658",
    "down-after-milliseconds": "30000",
    "last-hello-message": "1722",
    "voted-leader": "?",
    "voted-leader-epoch": "0"
  }
]
``` 
This pull request proposes to use different ports for TLS and non-TLS traffic in Sentinel:
- port `26379` for non-TLS traffic
- port `26380` for TLS traffic

By setting `tls-replication=yes`, communication between Sentinels is always TLS encrypted. 

This setup is in line with the TLS configuration of Valkey itself:
- Clients can connect to Sentinel and Valkey without TLS when client TLS is not enabled
- When client TLS is enabled, the default ports `6379` and `26379` will be closed and clients can only connect to Valkey and Sentinel on the TLS ports `6380` and `26380`


skourta (commented):
Thank you @reneradoi . I have tested the PR manually and can connect on both ports as expected. I also checked sentinel's statuses and performed a failover successfully.

reneradoi (approved):
Approved as per confirmation from @skourta.

Mehdi-Bendriss (approved):

PR #18, opened 2026-03-06
[DPE-9313] Add certificate and CA rotation

reneradoi:
This PR adds capabilities and tests for rotating TLS certificates to Valkey:
- rotate client TLS certificate provided by TLS provider
- rotate CA certificate provided by TLS provider
- rotate internal certificate (self-signed)
- rotate internal CA certificate (self-signed)

The rotation workflows are executed without downtime. In case of a CA rotation, the synchronization between units is orchestrated via `peer-relation-changed` events without having to defer the `certificate-available` events.


tests/integration/test_charm.py
```diff
@@ -271,7 +271,7 @@ async def test_user_secret_permissions(juju: jubilant.Juju) -> None:
     )
 
     juju.wait(
-        lambda status: are_apps_active_and_agents_idle(status, APP_NAME, idle_period=10),
+        lambda status: are_apps_active_and_agents_idle(status, APP_NAME, idle_period=60),
```
reneradoi, line 274: This is an unrelated change: I found in a couple of test runs that this doesn't wait for the actual secret change to be performed

src/events/tls.py
```diff
@@ -211,3 +306,48 @@ def _enable_client_tls(self) -> None:
...
+                    server.tls_ca_rotation_state
+                    in [TLSCARotationState.NEW_CA_ADDED, TLSCARotationState.CA_UPDATED]
+                    for server in self.charm.state.servers
+                ):
+                    raise ValkeyCertificatesNotReadyError
+
+                logger.info("Reload TLS certificates after all units have added the new CA")
+                tls_config = self.charm.config_manager.generate_tls_config()
+                self.charm.cluster_manager.reload_tls_settings(tls_config)
+                self.charm.sentinel_manager.restart_service()
```
skourta, line 320: Is it ok to restart the service without a rolling op? We are still mandated to have 0 downtime right?

src/events/tls.py
```diff
@@ -211,3 +306,48 @@ def _enable_client_tls(self) -> None:
...
+                self.charm.tls_manager.set_ca_rotation_state(TLSCARotationState.CA_UPDATED)
+
+                if len(self.charm.state.servers) == 1:
+                    self.charm.on[PEER_RELATION].relation_changed.emit(
```
skourta, line 324: Nice touch

src/events/tls.py
```diff
@@ -211,3 +306,48 @@ def _enable_client_tls(self) -> None:
...
+                    raise ValkeyCertificatesNotReadyError
+
+                logger.info("Remove old CA after all units have updated the TLS certificate")
+                self.charm.workload.tls_paths.client_ca.with_name("old_client_ca.pem").unlink(
+                    missing_ok=True
+                )
+                self.charm.tls_manager.rehash_ca_certificates()
+                tls_config = self.charm.config_manager.generate_tls_config()
+                self.charm.cluster_manager.reload_tls_settings(tls_config)
+                self.charm.sentinel_manager.restart_service()
```
skourta, line 342: Same comment

src/events/tls.py
```diff
@@ -211,3 +306,48 @@ def _enable_client_tls(self) -> None:
         tls_config = self.charm.config_manager.generate_tls_config()
         self.charm.cluster_manager.reload_tls_settings(tls_config)
         self.charm.sentinel_manager.restart_service()
+
+    def _orchestrate_ca_rotation(self) -> None:
+        """Orchestrate the workflow when a TLS CA rotation has been initiated."""
+        match self.charm.state.unit_server.tls_ca_rotation_state:
```
skourta, line 302: I love the match case. Very clean.

src/events/tls.py
```diff
@@ -89,6 +99,50 @@ def _on_peer_relation_created(self, event: ops.RelationCreatedEvent) -> None:
...
+                logger.debug("Not all units ready")
+            except (ValkeyServicesFailedToStartError, ValkeyTLSLoadError):
+                logger.error("Failed to reload TLS certificates")
+                event.defer()
+            finally:
+                return
+
+        try:
+            # will raise if cert expires soon
+            self.charm.tls_manager.check_certificate_validity()
```
skourta: I think it would be cleaner and easier to read to extract this as a helper function that returns a boolean. It's easy to miss the return after the if

src/events/tls.py
skourta: You could also just call the `_orchestrate_ca_rotation` which is what we want no?

tests/integration/tls/test_certificate_rotation.py
```diff
@@ -0,0 +1,185 @@
...
+TEST_KEY = "test_key"
+TEST_VALUE = "test_value"
+CERTIFICATE_EXPIRY_TIME = 320
+
+
+def test_build_and_deploy(charm: str, juju: jubilant.Juju, substrate: Substrate) -> None:
+    """Deploy the charm under test and a TLS provider."""
+    juju.deploy(
+        charm,
+        resources=IMAGE_RESOURCE if substrate == Substrate.K8S else None,
```
skourta, line 55: Starting to think we should have this as a fixture and asbtract it away as well.

tests/integration/tls/test_certificate_rotation.py
```diff
@@ -0,0 +1,185 @@
...
+        lambda status: does_status_match(
+            status,
+            expected_unit_statuses={APP_NAME: [TLSStatuses.CERTIFICATE_EXPIRING.value]},
+            num_units={APP_NAME: NUM_UNITS},
+        ),
+        timeout=100,
+    )
+
+
+async def test_ca_rotation_by_config_change(juju: jubilant.Juju) -> None:
```
skourta, line 155: Can we also have a ca rotation by expiration like in etcd?

tests/integration/tls/test_certificate_rotation.py
```diff
@@ -0,0 +1,185 @@
...
+    """Test the CA rotation.
+
+    The CA certificate should be rotated and the cluster should still be accessible.
+    The rotation is triggered by updating the config for `ca-common-name` on the TLS provider side.
+    """
+    # Rotate the CA certificate
+    logger.info("Getting the current CA certificates")
+    download_client_certificate_from_unit(juju, APP_NAME)
+    with open(TLS_CA_FILE, "r") as file:
```
skourta: nit: let's stay consistent through all charm code ans use pathlib here

tests/integration/tls/test_certificate_rotation.py
```diff
@@ -0,0 +1,185 @@
...
+        lambda status: are_agents_idle(status, APP_NAME, idle_period=30, unit_count=NUM_UNITS),
+        timeout=600,
+    )
+
+    logger.info("Checking if the CA certificates are rotated")
+    download_client_certificate_from_unit(juju, APP_NAME)
+    with open(TLS_CA_FILE, "r") as file:
+        new_ca_certificate = file.read()
+    assert new_ca_certificate, "Failed to get updated ca certificate"
+    assert old_ca_certificate != new_ca_certificate, "CA certificate was not updated"
```
skourta, line 298: We can also have checks on the certificates themselves. THey should be updated as well

skourta (commented):
Thank you René. One question and a few minor comments.

src/events/tls.py
reneradoi: I opted for triggering the event instead in order to have the same workflow, independent if 1 or multiple units.

src/events/tls.py
reneradoi: Yes, the database itself has 0 downtime because it can reload without restarting.

Sentinel is not impacting the availablity of Valkey. It also takes 30 seconds until other sentinels consider this sentinel down, so they won't even realize it. And we should reload TLS files in sentinel shortly after Valkey to make sure it doesn't fall behind on the changes (e.g. uses the old CA when it is already removed in Valkey).

tests/integration/tls/test_certificate_rotation.py
reneradoi: This is on the test runner machine, not on any of the units. It should also work on any machine where this integration test is run, therefor I would like to keep pathops (which requires us to define a path for the file, starting from root) out of here.

tests/integration/tls/test_certificate_rotation.py
reneradoi: Added.

src/events/tls.py
reneradoi: I've refactored it to return a bool. Please let me know if you have an idea for a better name.

tests/integration/tls/test_certificate_rotation.py
reneradoi: I've tested it manually to ensure it works. 

@Mehdi-Bendriss What do you think: Should we apply [the same hack as in etcd](https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/tests/integration/tls/test_ca_rotation.py#L214) and automate the test for CA rotation on expiration?

src/events/tls.py
skourta: I'm ok with this name. Other alternatives might be `are_certificates_about_to_expire` or `are_certificates_expiring`.

tests/integration/tls/test_certificate_rotation.py
Mehdi-Bendriss: yes, that's sensible

reneradoi: thx. for confirmation, I have updated it.

src/events/tls.py
```diff
@@ -89,6 +99,44 @@ def _on_peer_relation_created(self, event: ops.RelationCreatedEvent) -> None:
         except ValkeyWorkloadCommandError as e:
             logger.error("Failed to create certificate for peer-TLS, startup will fail: %s", e)
 
+    def _on_peer_relation_changed(self, event: ops.RelationChangedEvent) -> None:
+        """Handle TLS related changes to the peer relation."""
+        if self.charm.state.unit_server.tls_ca_rotation_state != TLSCARotationState.NO_ROTATION:
+            try:
+                self._orchestrate_ca_rotation()
+            except ValkeyCertificatesNotReadyError:
+                logger.debug("Not all units ready")
```
Mehdi-Bendriss, line 108: Do we just abandon the event here?

src/events/tls.py
```diff
@@ -114,32 +162,51 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
...
-                ValkeyServicesFailedToStartError,
-                ValkeyTLSLoadError,
-                ValueError,
-            ):
-                logger.error("Failed to enable client TLS")
-                event.defer()
-                return
+                if rotate_ca:
+                    self.charm.tls_manager.set_ca_rotation_state(TLSCARotationState.NEW_CA_ADDED)
+                    self._orchestrate_ca_rotation()
```
Mehdi-Bendriss, line 180: I'm a bit confused by this flow, from one side we rely on_peer_rel_changed to `_orchestrate_ca_rotation` on another hand we rely directly on `_certificates_available_event`?

src/events/tls.py
```diff
@@ -195,6 +262,22 @@ def _on_tls_relation_broken(self, event: ops.RelationBrokenEvent) -> None:
...
+        """Handle TLS related parts of update_status event."""
+        if not self.charm.tls_manager.will_certificate_expire():
+            return
+
+        if self.charm.state.client_tls_relation:
+            return
+
+        if self.charm.unit.is_leader():
+            # need to renew CA first (same validity), this triggers relation-changed event
+            self.charm.tls_manager.generate_ca_certificate()
```
Mehdi-Bendriss, line 277: this is the case for self-managed self-signed CA, right?

src/managers/tls.py
```diff
@@ -169,10 +179,83 @@ def create_and_store_self_signed_certificate(self) -> None:
...
+        if ca_cert.raw == current_ca_cert:
+            logger.debug("No CA rotation, CA cert is up-to-date")
+            return False
+
+        logger.info("New CA certificate detected")
+        self.workload.write_file(
+            current_ca_cert, self.workload.tls_paths.client_ca.with_name("old_client_ca.pem")
+        )
+
+        self.set_ca_rotation_state(TLSCARotationState.NEW_CA_DETECTED)
```
Mehdi-Bendriss: should we take this one our of the manager? we are setting the CA rotation state between the manager and event handler, we should perhaps stay consistent

src/events/tls.py
```diff
@@ -114,32 +162,51 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
...
-                event.defer()
-                return
+                if rotate_ca:
+                    self.charm.tls_manager.set_ca_rotation_state(TLSCARotationState.NEW_CA_ADDED)
+                    self._orchestrate_ca_rotation()
+                    return
+
+                tls_config = self.charm.config_manager.generate_tls_config()
+                self.charm.cluster_manager.reload_tls_settings(tls_config)
+                self.charm.sentinel_manager.restart_service()
```
Mehdi-Bendriss, line 185: I have 2 points here:
1. why is sentinel restart triggered? my understanding was that "Sentinel inherits its networking configuration from the common Valkey configuration, so all of the above applies to Sentinel as well." [source](https://valkey.io/topics/tls/#sentinel)
2. why not rolling restart the sentinels? 
  - clients using sentine-aware libs will go down
  - if the master dies during that period, no sentinel may be available to promote one (clients may face down time), 
  - and when they're back, a failover can't start until enough sentinels are back, and we'll end up with a confused topology until they're all back

src/events/tls.py
Mehdi-Bendriss: I think there is a nuance here, the issue is not about Valkey itself when everything is alright, the problem that will occur is when the primary is down while sentinels are restarting - related comment https://github.com/canonical/valkey-operator/pull/18/changes#r2934298273

src/events/tls.py
Mehdi-Bendriss: related comment https://github.com/canonical/valkey-operator/pull/18/changes#r2934298273

Mehdi-Bendriss (commented):
Thanks René! Good work! I left a few questions and comments

src/events/tls.py
reneradoi: Yes, if not all units ready. Reason being: When a unit becomes ready, it updates its peer relation data and thereby triggers another `peer relation changed` event. So no need to defer in this case.

src/events/tls.py
reneradoi: Yes, because this is where the operator can detect a CA rotation and initiate the workflow.

The main reason for relying on `peer relation changed` for the workflow synchronization is that it allows us to handle both TLS use cases the same way with the same code: external certs from TLS provider and internal, self-managed certs. Only the way it is triggered differs:
- certificate available with new CA in case of TLS provider
- certificate validity check in case of self-managed certs

From the trigger point onwards, both have the same workflows, orchestrated by updates to the peer unit data and handled by the `peer relation changed` event.

src/events/tls.py
reneradoi: Yes, correct. If we have a client TLS relation, we leave early.

src/managers/tls.py
reneradoi: Updated. We need a `noqa: C901` now :grimacing:

src/events/tls.py
reneradoi: We (currently*) need to restart Sentinel because it can not be triggered with `config set` to reload the TLS files (different than Valkey itself). In Sentinel, only a few "global" configuration parameters can be reloaded:
- resolve-hostnames, announce-hostnames
- announce-ip, announce-port
- sentinel-user, sentinel-pass
(https://valkey.io/topics/sentinel/#reconfiguring-sentinel-at-runtime)

/* Currently because a PR ([#3023](https://github.com/valkey-io/valkey/pull/3020)) has been merged that allows auto-reload of TLS files. Once that is available (`9.1`), we can move away from restarting Sentinel and reloading Valkey for certificate and CA rotation. `9.1` is expected to be released at the end of April ([release schedule](https://github.com/valkey-io/valkey/wiki/Upcoming-release-schedule)).

On the question of rolling restarts: I don't think we should add rolling restarts here. First reason is the above mentioned feature that will make it obsolete to use restarts on Sentinel. Furthermore, Sentinel's responsibilities are not impacted by the restart (even for multiple units at the same time):
- A `primary down` (`ODOWN`) will only be detected after `sentinel down-after-milliseconds`, in our current configuration that is 30 seconds, while the restart is very quick.
- The restart is bound to succeed because the TLS configuration has been verified by the `config set` command on Valkey just before, it guarantees that TLS configuration for Sentinel is valid.
- Clients only need to connect to Sentinel in case of disrupted Valkey connection, not on every connection. So clients will not go down if a Sentinel is not available
- Sentinel endpoints are not required to be available all the time by clients that are aware of Sentinel: `The client should iterate the list of Sentinel addresses. For every address it should try to connect to the Sentinel, using a short timeout (in the order of a few hundreds of milliseconds). On errors or timeouts the next Sentinel address should be tried.` (https://valkey.io/topics/sentinel-clients/)

src/events/tls.py
reneradoi: Please see the discussion [here](https://github.com/canonical/valkey-operator/pull/18#discussion_r2939145091).

src/events/tls.py
reneradoi: Please see the discussion [here](https://github.com/canonical/valkey-operator/pull/18#discussion_r2939145091).

src/events/tls.py
Mehdi-Bendriss: I did some reading and research after this morning's discussion, and Valkey 9.1 aside (certificates/CA renewal). On the case of No-TLS to TLS or the opposite, as I was fearing, we still have the risk of Sentinel taking +30 sec to restart due to the health and condition of the environment runtime. 

We cannot assume guarantees of "fast" restart in a few cases such as:
- CPU starvation / scheduler delays - entering [TILT mode](https://valkey.io/topics/sentinel/#tilt-mode)
- Slow or stalled disk I/O - (sentinel relying on the sentinel.conf on disk)
- Service stop delay on an overloaded host - on a small instance under pressure, the old process may not shut down quickly, and the service manager can spend a lot of time waiting before force killing it AKA "slow restart" (note: even if sentinel start is not the issue)
- Slow DNS/network caused by host pressure (for reconnects) - `The name resolution configuration on your Valkey and Sentinel nodes must be reliable and be able to resolve addresses quickly. Unexpected delays in address resolution may have a negative impact on Sentinel.` [reference](https://valkey.io/topics/sentinel/#ip-addresses-and-dns-names) 

With this, I'm afraid we need a rolling restart and cannot make an assumption on a quick restart :(

reneradoi: Thank you for the detailed answer. I agree with the conclusion. This will be addressed in a separate PR, once TLS and HA features have been merged into `9/edge`. I have created an issue (#25) to keep track of it.

Mehdi-Bendriss (approved):

skourta (approved):
Thank you @reneradoi

PR #19, opened 2026-03-10
[DPE-9373]: Use hostnames instead of IPs for k8s

skourta:
This pull request introduces a significant refactor to how the codebase handles primary endpoint addressing, shifting from using only IP addresses to supporting both IP addresses and hostnames, depending on the deployment substrate. It updates configuration management and event handling to use the new `primary_endpoint` concept, adds hostname resolution to Sentinel management, and ensures better compatibility with non-VM substrates. Additionally, it introduces a utility function for IP validation and improves Sentinel configuration for hostname-based communication.

Key changes include:

**Primary Endpoint Refactor and Configuration Management:**

- Refactored all configuration and service management logic in `config.py` and related event handlers to use a new `primary_endpoint` parameter (which can be a hostname or IP) instead of just `primary_ip`. This includes updating method signatures, internal logic, and how endpoints are determined based on the substrate (VM vs. Kubernetes). [[1]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL106-R106) [[2]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL140-R147) [[3]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL156-R160) [[4]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cL50-R50) [[5]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cL85-R88) [[6]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cL97-R122) [[7]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cL196-R205) [[8]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cL244-R253) [[9]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cL253-R268) [[10]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cR284-R292) [[11]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cL338-R361) [[12]](diffhunk://#diff-3dfb4a104b407f7d4b2e07795a63745a09cde228a8dbe86602e7b72ea600fbe4L161-R166) [[13]](diffhunk://#diff-3dfb4a104b407f7d4b2e07795a63745a09cde228a8dbe86602e7b72ea600fbe4L209-R212)

- Updated event-driven updates to always use FQDN hostnames instead of short hostnames, improving consistency and DNS compatibility. [[1]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL106-R106) [[2]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL248-R256) [[3]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL290-R298)

**Sentinel Management Improvements:**

- Enhanced Sentinel management to resolve hostnames to IPs when necessary, using the new `is_valid_ip` helper, and improved logic for retrieving active Sentinel IPs. [[1]](diffhunk://#diff-b226109a257f8cc9cb6b0a4a1eb4c1c730d2cc9620744b6855e4f3a96ca3041dR22) [[2]](diffhunk://#diff-b226109a257f8cc9cb6b0a4a1eb4c1c730d2cc9620744b6855e4f3a96ca3041dL272-R285)

- Added `resolve-hostnames` and `announce-hostnames` options to the generated Sentinel configuration to enable hostname-based communication within the cluster.

**Utility Enhancements:**

- Introduced a new `is_valid_ip` helper function in `common/helpers.py` to reliably check if a string is a valid IP address, supporting the above refactors and Sentinel logic.

These changes collectively improve the charm's flexibility and reliability in heterogeneous environments, especially for Kubernetes and other non-VM substrates.


gboutry:
Hello, drive by comment, but I'm very interested in this charm, and I would love to see:
- ensure if we give an fqdn from k8s in the relation that it's appended with `.` (full fqdn notation) (https://github.com/canonical/mysql-router-k8s-operator/pull/445)
- And if you could take a look at: https://github.com/canonical/mysql-operators/issues/137

For the first point:
We've noticed in deployments from our experience with MySQL that when not having a trailing . clients will launch repeated queries with up to the 5 domain search defined in regular k8s cluster, and the FQDN is always last, and with many clients to your service, this could lead to increased usage of coredns. (it's even more apparent for CMR, since you're client is most likely to go through the 5 domain search before finding the suitable match)

For the second point:
We've noticed issues from the mysql charm when trying to resolve a suitable endpoints style fqdn, especially when there's multiple PTR records. More info on the issue linked.

skourta:
Thank you for your interest @gboutry and for bringing up these concerns. I have ported the fixes implemented in MySQL to this PR.

src/common/helpers.py
```diff
@@ -0,0 +1,32 @@
...
+    """Append a dot to a string if it does not already end with one."""
+    if not string.endswith("."):
+        string += "."
+    return string
+
+
+def get_k8s_fqdn(name: str) -> str:
+    """Resolve the canonical FQDN for a Kubernetes service or pod name."""
+    try:
+        info = socket.getaddrinfo(
```
reneradoi: Why not using `socket.getfqdn()`?

```
$ juju ssh valkey/1
# python3
Python 3.12.3 (main, Nov  6 2024, 18:32:19) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import socket
>>> socket.getfqdn()
'valkey-1.valkey-endpoints.valkey.svc.cluster.local'
```

We could simplify and just get that in the state, and then append a dot if it doesn't have one.

reneradoi (commented):
Thanks for bringing this change, Smail! I've tested it and it looks good. Regarding the implementation I have only one comment/question.

src/common/helpers.py
skourta: That is how I initially implemented it. This is what @gboutry raised on his 2nd bullet point. See the details [here](https://github.com/canonical/mysql-operators/issues/137)

reneradoi: Ok got it, thank you!

reneradoi (approved):

src/common/helpers.py
```diff
@@ -2,20 +2,31 @@
...
-        return False
+        info = socket.getaddrinfo(
+            name,
+            None,
+            family=socket.AF_UNSPEC,
+            flags=socket.AI_CANONNAME,
+            type=socket.SOCK_STREAM,
+        )
+    except socket.gaierror as e:
+        raise RuntimeError(f"Failed to resolve canonical name for {name}") from e
```
Mehdi-Bendriss: please no RuntimeError in the code, catch the `socket.gaierror` in the client code instead or raise a custom `K8sDNSResolutionError`  or better just return the name

src/common/helpers.py
```diff
@@ -2,20 +2,31 @@
...
+
+    for entry in info:
+        if canonname := entry[3]:
+            return canonname
+
+    raise RuntimeError(f"Could not determine canonical name for {name}")
```
Mehdi-Bendriss: raise a `socket.gaierror` here or raise a custom `K8sDNSResolutionError` and as fallback potentially return the arg name

src/core/cluster_state.py
```diff
@@ -134,14 +134,20 @@ def ingress_address(self) -> str | None:
...
+
     @property
     def endpoint(self) -> str:
         """The endpoint to be used by other units to connect to this unit.
 
         On VM-based substrates, this should be the bind address.
         On Kubernetes, this should be the fully qualified domain name of the unit.
         """
-        return self.bind_address if self.substrate == Substrate.VM else socket.getfqdn()
+        return self.bind_address if self.substrate == Substrate.VM else self.fqdn
```
Mehdi-Bendriss: 1. Why do we resolve the FQDN here using `getaddrinfo()` since it can be generated deterministically?
f"{unit}.{self.app.name}-endpoints.{model}.svc.cluster.local"

2. On k8s, the unit identity is also deterministic (`<pod>.<headless-service`> == `f"{unit}.{self.app.name}-endpoints`). Meaning, what `get_unit_hostname` returns is plenty for internal cluster communication (replication, sentinel, peer discovery, etc).
Performing DNS canonicalization introduces an unnecessary dependency on DNS and adds complexity. The deterministic pod identity (`get_unit_hostname`) can/should be used directly here. 
The fqdn is useful for SANs (on top of get_unit_hostname) or client-facing endpoints (client relations for ex), but it is not required for internal cluster wiring.

3. If I understand the problem correctly on the trailing dot, I'm not so sure about the solution. I think this can break some TLS hostname verification libraries (based on pure san hostnames equality) - so I'm not sure it is worth the fewer (1?) DNS lookups we want to end up with, especially if we were already returning the FQDN on the relation.

(side note: for external endpoints we may also create primary+replica services on `keyspace_requested` event)

src/events/base_events.py
```diff
@@ -103,7 +102,7 @@ def _on_start(self, event: ops.StartEvent) -> None:
         self.charm.state.unit_server.update(
             {
                 "start_state": StartState.NOT_STARTED.value,
-                "hostname": socket.getfqdn(),
+                "hostname": self.charm.state.fqdn,
```
Mehdi-Bendriss: similar comment

Mehdi-Bendriss (commented):
Thanks Smail! I have a couple of comments

src/core/cluster_state.py
skourta: I will move the FQDN to be statically generated based on the template. I ported the MySQL changes as they are on this one.

skourta: Any suggestions on the trailing dot?

skourta: Confirmed 2: `{unit.name}.{app.name}-endpoints` is enough for the workload to correctly resolve the IP and communicate.

Mehdi-Bendriss: I believe we can park the dot one for the time being, it would be considered an optimization anyway, let's just create a new issue so we don't lose track of it

Mehdi-Bendriss (approved):

src/core/cluster_state.py
skourta: [Ticket created](https://github.com/canonical/valkey-operator/issues/22)

PR #20, opened 2026-03-12
[DPE-9315] add user-provided private key

reneradoi:
This PR adds functionality + tests for having a private key for client certificates provided by users. The private key is added as a user secret and the secret URI configured to the charm. If the provided private key is invalid or the charm has no access, the previously used private key will be maintained.


skourta (approved):
Amazing work @reneradoi. No concerns from my part. The implementation is identical to the one we have and hardened on etcd.

src/core/models.py
```diff
@@ -235,3 +237,12 @@ def internal_ca_private_key(self) -> PrivateKey | None:
             return None
 
         return PrivateKey.from_string(self.model.internal_ca_private_key)
+
+    @property
+    def tls_client_private_key(self) -> PrivateKey | None:
+        """Retrieve the private key for client TLS."""
+        if self.model and (private_key := self.model.tls_client_private_key):
+            private_key = PrivateKey(raw=private_key)
+            return private_key
```
Mehdi-Bendriss:
```suggestion
            return PrivateKey(raw=private_key)
```

Mehdi-Bendriss (approved):
Thanks René!

PR #23, opened 2026-03-16
[DPE-9325]: Network HA tests

skourta:
This pull request introduces several improvements focused on handling TLS certificate updates when a unit's IP address changes, enhances lock management logic, and adds support for advanced TLS configuration in integration tests. It also introduces new HA testing infrastructure for Kubernetes environments.

TLS certificate and IP change handling:

* Added logic to detect IP address changes and trigger certificate regeneration or refresh, including a new `_on_ip_change` method in `base_events.py` and a custom exception `TLSCertificatesRequireRefreshError`. Certificates are now checked for SAN changes and updated if needed, with events emitted for client relations. (`src/events/base_events.py`, `src/managers/tls.py`, `src/common/exceptions.py`, [[1]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eR291-R301) [[2]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eR550-R576) [[3]](diffhunk://#diff-a2c036f44ae9bbab595aad84b23bc3106a342c5ab12f7b0cf3d26268b51cdc83R172-R218) [[4]](diffhunk://#diff-e793bad618b16c64102391f92909ac368210513be36cc7ff3ea633e458e54feeR69-R72)
* Updated integration tests to support TLS, including reading certificate files and configuring Glide clients with advanced TLS settings. (`tests/integration/continuous_writes.py`, [[1]](diffhunk://#diff-e992c4deba695defabfadb58b521e21a92646fb5e8a128fca2a80c5162790697R18-R24) [[2]](diffhunk://#diff-e992c4deba695defabfadb58b521e21a92646fb5e8a128fca2a80c5162790697R79-R127) [[3]](diffhunk://#diff-e992c4deba695defabfadb58b521e21a92646fb5e8a128fca2a80c5162790697L246-R312)

Lock management enhancements:

* Improved lock handling by adding a timestamp for lock requests/releases, customizing lock timestamp attribute per subclass, and refining logic for determining the next unit to receive the lock. (`src/common/locks.py`, [[1]](diffhunk://#diff-d03a955c679a606f62dada86ecc23f37ab4e4d3bb47a9647bd10c82979481d18R55-R64) [[2]](diffhunk://#diff-d03a955c679a606f62dada86ecc23f37ab4e4d3bb47a9647bd10c82979481d18R77-R78) [[3]](diffhunk://#diff-d03a955c679a606f62dada86ecc23f37ab4e4d3bb47a9647bd10c82979481d18R109-R113) [[4]](diffhunk://#diff-d03a955c679a606f62dada86ecc23f37ab4e4d3bb47a9647bd10c82979481d18R126-R130) [[5]](diffhunk://#diff-d03a955c679a606f62dada86ecc23f37ab4e4d3bb47a9647bd10c82979481d18R170)

Kubernetes HA testing infrastructure:

* Added new fixtures and helpers for deploying Chaos Mesh in Kubernetes environments to enable HA failure simulations, including a deployment script and network loss YAML. (`tests/integration/ha/conftest.py`, `tests/integration/ha/helpers/deploy_chaos_mesh.sh`, `tests/integration/ha/helpers/chaos_network_loss.yml`, [[1]](diffhunk://#diff-c36ba0d7364f531c62b997bae37483d1479aab7f85db704b017f3a4e62db92a8R1-R24) [[2]](diffhunk://#diff-ffc837f97657b488ac78ab4befea1a08dde96783c3c0f92bcebccf9224fa7bb4R1-R26) [[3]](diffhunk://#diff-76837a80d0f97900d9ca7ec7657e89e239d32304e355e65b3a034a829eb14b4fR1-R18)

Configuration and dependency updates:

* Added `kubernetes` dependency to `pyproject.toml` for K8S-related testing and operations. (`pyproject.toml`, [pyproject.tomlR57](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R57))
* Updated config management to bind to the current endpoint and propagate IP changes to Valkey runtime configuration. (`src/managers/config.py`, src/managers/cluster.py, [[1]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cL81) [[2]](diffhunk://#diff-20b77aa4083c3162fac7cd352e1b8ceef186a967a37f906277cb77346967c113R70-R78)

These changes improve the robustness of certificate management, lock coordination, and enable advanced HA testing capabilities in Kubernetes environments.


src/events/tls.py
```diff
@@ -299,6 +301,30 @@ def _enable_client_tls(self) -> None:
         self.charm.cluster_manager.reload_tls_settings(tls_config)
         self.charm.sentinel_manager.restart_service()
 
+    def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
```
reneradoi: Separating concerns is nice and I really like it! After reading the code though I think I need to revert my earlier comment and admit that it makes more sense to have it in the same event handler and method. Reason being that it seems to be overly complicated just to adhere to the structure, and now half of the logic here is not related to TLS concerns. Therefore, I'd say let's move this code to the `base_events._on_config_changed()` and have a clean implementation there.

src/events/base_events.py
```diff
@@ -524,3 +552,54 @@ def _set_state_for_going_away(self) -> None:
...
+            check_replica_sync=False
+        ):
+            self.charm.status.set_running_status(
+                ClusterStatuses.VALKEY_UNHEALTHY_RESTART.value,
+                scope="unit",
+                component_name=self.charm.cluster_manager.name,
+                statuses_state=self.charm.state.statuses,
+            )
+
+        self.charm.state.statuses.delete(
```
reneradoi: Does this reset a status that might just previously have been set? Or am I missing something here?

src/events/base_events.py
```diff
@@ -41,6 +41,29 @@
 logger = logging.getLogger(__name__)
 
 
+class RestartWorkloadEvent(ops.EventBase):
+    """Event for restarting the workload when certain events happen, e.g. IP change."""
+
+    def __init__(
+        self, handle: ops.Handle, restart_valkey: bool = True, restart_sentinel: bool = True
```
reneradoi, line 48: Nice! This will certainly be useful!

src/events/base_events.py
```diff
@@ -524,3 +552,54 @@ def _set_state_for_going_away(self) -> None:
...
+        if not restart_lock.is_held_by_this_unit:
+            logger.info("Waiting for lock to restart workload")
+            event.defer()
+            return
+
+        if event.restart_valkey:
+            self.charm.workload.restart(self.charm.workload.valkey_service)
+        if event.restart_sentinel:
+            self.charm.sentinel_manager.restart_service()
+
```
reneradoi: We need to add error handling here in case these methods raise, otherwise the event would crash.

tests/integration/ha/helpers/destroy_chaos_mesh.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+# Utility script to removing chaosmesh from the K8S cluster, to clean up test artefacts
```
reneradoi, line 3: If available, can you please add the source of these `deploy/destroy` scripts as a comment so that we can cross check in case of future failures?

tests/integration/ha/test_network_cut.py
```diff
@@ -0,0 +1,202 @@
...
+    if substrate == Substrate.K8S:
+        primary_hostname = f"{primary_hostname}.{APP_NAME}-endpoints"
+
+    logger.info("Identified container name for primary unit: %s", primary_hostname)
+    cut_network_from_unit(substrate, juju.model, machine_name, change_ip=change_ip)
+
+    for unit in juju.status().apps[APP_NAME].units:
+        if unit == primary_unit_name:
+            continue
+        assert not is_unit_reachable(
```
reneradoi, line 129: In addition to that, we should also make sure that the connectivity from the controller is lost, see here for reference: https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/tests/integration/ha/test_network_cut.py#L236

reneradoi (commented):
Very nice, especially the test coverage and the restart lock! I only have a few comments, but general concept looks good already.

src/common/locks.py
```diff
@@ -52,10 +53,16 @@ class DataBagLock(Lockable):
 
     unit_request_lock_atr_name: str
     member_with_lock_atr_name: str
+    lock_timestamp: str = "databaglock_timestamp"
```
Mehdi-Bendriss, line 56: these iterations are exactly why we wanna have a unified locking mechanism 😬

src/events/base_events.py
```diff
@@ -524,3 +552,54 @@ def _set_state_for_going_away(self) -> None:
...
+        if event.restart_sentinel and not self.charm.sentinel_manager.is_healthy():
+            self.charm.status.set_running_status(
+                ClusterStatuses.SENTINEL_UNHEALTHY_RESTART.value,
+                scope="unit",
+                component_name=self.charm.cluster_manager.name,
+                statuses_state=self.charm.state.statuses,
+            )
+
+        self.charm.state.statuses.delete(
```
Mehdi-Bendriss: why are we clearing immediately a running status we just set ?

src/events/tls.py
```diff
@@ -299,6 +301,30 @@ def _enable_client_tls(self) -> None:
...
+        """Handle the `config-changed` event."""
+        if (
+            self.charm.state.unit_server.model.private_ip
+            and self.charm.state.bind_address != self.charm.state.unit_server.model.private_ip
```
Mehdi-Bendriss: is my understanding correct that this only applies to VM? if yes, since k8s uses FQDNs why not exit the hook immediately?

src/events/tls.py
```diff
@@ -299,6 +301,30 @@ def _enable_client_tls(self) -> None:
...
+        ):
+            if self.charm.tls_manager.certificate_sans_require_update():
+                if not self.charm.state.client_tls_relation:
+                    self.charm.tls_manager.create_and_store_self_signed_certificate()
+                else:
```
Mehdi-Bendriss: can we invert the conditions and exit?

tests/integration/ha/test_network_cut.py
```diff
@@ -0,0 +1,202 @@
...
+        lambda status: are_apps_active_and_agents_idle(status, APP_NAME, idle_period=30),
+        timeout=600,
+    )
+
+    assert len(juju.status().apps[APP_NAME].units) == NUM_UNITS, (
+        f"Unexpected number of units after initial deploy: expected {NUM_UNITS}, got {len(juju.status().apps[APP_NAME].units)}"
+    )
+
+
+@pytest.mark.parametrize("tls_enabled", [False, True], ids=["tls_off", "tls_on"])
```
Mehdi-Bendriss, line 71: I like this

tests/integration/ha/test_network_cut.py
```diff
@@ -0,0 +1,202 @@
...
+@pytest.mark.parametrize("change_ip", [True, False], ids=["change_ip", "no_change_ip"])
```
Mehdi-Bendriss: And this :)

tests/integration/ha/test_network_cut.py
```diff
@@ -0,0 +1,202 @@
...
+        pytest.skip("Changing IP is not applicable for k8s substrate.")
+
+    download_client_certificate_from_unit(juju, APP_NAME)
+
+    c_writes.tls_enabled = tls_enabled
+    await c_writes.async_clear()
+    c_writes.start()
+
+    # Get the current primary unit
+    primary_ip = get_primary_ip(juju, APP_NAME, tls_enabled=tls_enabled)
```
Mehdi-Bendriss, line 94: nit, for consistency - `primary_host/endpoint` (your reasoning makes sense for the helpers themselves as discussed)

tests/integration/ha/test_network_cut.py
```diff
@@ -0,0 +1,202 @@
...
+            except ValueError as e:
+                logger.warning(f"Error getting primary IP after network cut: {e}")
+            logger.info("Waiting for new primary to be elected...")
+
+    assert new_primary_ip and new_primary_ip != primary_ip, (
+        "Primary IP did not change after cutting network to the primary unit."
+    )
+    logger.info(
+        "New primary IP after network cut: %s vs old primary IP: %s", new_primary_ip, primary_ip
+    )
```
Mehdi-Bendriss, line 165: isn't this only for the case where `change_ip == True` what about the when no ip change?

tests/integration/ha/helpers/helpers.py
```diff
@@ -0,0 +1,363 @@
...
+from tests.integration.helpers import APP_NAME
+
+logger = getLogger(__name__)
+
+
+def lxd_cut_network_from_unit_with_ip_change(machine_name: str) -> None:
+    """Cut network from a lxc container in a way the changes the IP."""
+    # apply a mask (device type `none`)
+    cut_network_command = f"lxc config device add {machine_name} eth0 none"
+    subprocess.check_call(cut_network_command.split())
```
Mehdi-Bendriss, line 31: We need to add a sleep here, refer to opensearch, this is due to LXD sometimes taking a few seconds before propagating the changes.

tests/integration/ha/test_network_cut.py
```diff
@@ -0,0 +1,202 @@
...
+    await assert_continuous_writes_increasing(
+        hostnames=hostnames,
+        username=CharmUsers.VALKEY_ADMIN.value,
+        password=get_password(juju, user=CharmUsers.VALKEY_ADMIN),
+        tls_enabled=tls_enabled,
+    )
+
+    # restore network to the original primary unit
+    logger.info("Restoring network to original primary unit at %s", primary_hostname)
+    restore_network_to_unit(substrate, juju.model, machine_name)
```
Mehdi-Bendriss: shouldn't we distinguish between with/without IP change? namely resetting the ingress/egress/priority

tests/integration/ha/test_network_cut.py
```diff
@@ -0,0 +1,202 @@
...
+    juju.wait(
+        lambda status: are_apps_active_and_agents_idle(
+            status, APP_NAME, unit_count=NUM_UNITS, idle_period=30
+        )
+    )
+    c_writes.update()
```
Mehdi-Bendriss, line 218: Is this for the case where we have an IP change?

tests/integration/ha/test_network_cut.py
```diff
@@ -0,0 +1,202 @@
...
+    # check replica number that it is down to NUM_UNITS - 2
+    number_of_replicas = await get_number_connected_replicas(
+        hostnames=hostnames,
+        username=CharmUsers.VALKEY_ADMIN.value,
+        password=get_password(juju, user=CharmUsers.VALKEY_ADMIN),
+        tls_enabled=tls_enabled,
+    )
+    assert number_of_replicas == NUM_UNITS - 2, (
+        f"Expected {NUM_UNITS - 2} connected replicas, got {number_of_replicas}."
+    )
```
Mehdi-Bendriss, line 184: Let's also add a check here that confirms the endpoints returned by Sentinel do not contain the old primary

tests/integration/ha/test_network_cut.py
```diff
@@ -0,0 +1,202 @@
...
+            juju, hostname_from_unit(juju, unit), primary_hostname, substrate
+        ), (
+            f"Unit {unit} cannot reach the original primary unit {primary_hostname} after network restoration."
+        )
+
+    download_client_certificate_from_unit(juju, APP_NAME, unit_name=primary_unit_name)
+    # read ip from cert and check if is a different ip than before if change_ip is True
+    certificate_sans = get_sans_from_certificate("./client.pem")
+    if change_ip:
+        assert primary_ip not in certificate_sans["sans_ip"], (
```
Mehdi-Bendriss: Let's also add a check here that confirms the endpoints returned by Sentinel do not contain the old primary but contain the new one

tests/integration/ha/test_network_cut.py
```diff
@@ -0,0 +1,202 @@
...
+    number_of_replicas = await get_number_connected_replicas(
+        hostnames=hostnames,
+        username=CharmUsers.VALKEY_ADMIN.value,
+        password=get_password(juju, user=CharmUsers.VALKEY_ADMIN),
+        tls_enabled=tls_enabled,
+    )
+    assert number_of_replicas == NUM_UNITS - 1, (
+        f"Expected {NUM_UNITS - 1} connected replicas after network restoration, got {number_of_replicas}."
+    )
+
```
Mehdi-Bendriss, line 269: Let's also add a check here that confirms the endpoints returned by Sentinel contain the old primary

Mehdi-Bendriss (commented):
Thanks Smail! Good work! I have some comments

src/common/locks.py
skourta: Yes, I have touched based with Patricia again last week about using the lib. I will open a PR for it when things settle down a little bit.

src/events/tls.py
skourta: Mostly, but we still need to update the certs for k8s

src/events/tls.py
skourta: I'm ok moving it back but the current code is to update the certificates on IP change.

tests/integration/ha/test_network_cut.py
skourta: No this is for failover. We want to check that a new primary has been elected.

tests/integration/ha/test_network_cut.py
skourta: Yes, correct to update the list of endpoints for the cw

tests/integration/ha/test_network_cut.py
skourta: Added

src/events/base_events.py
```diff
@@ -567,39 +567,47 @@ def _on_restart_workload(self, event: RestartWorkloadEvent) -> None:
...
 
-        if event.restart_valkey:
-            self.charm.workload.restart(self.charm.workload.valkey_service)
-        if event.restart_sentinel:
-            self.charm.sentinel_manager.restart_service()
+        try:
+            if event.restart_valkey:
+                self.charm.workload.restart(self.charm.workload.valkey_service)
+            if event.restart_sentinel:
+                self.charm.sentinel_manager.restart_service()
```
Mehdi-Bendriss, line 607: could you end the `except` below this line to reduce the nestedness and highlight which pieces of code raise the exception?

src/events/base_events.py
```diff
@@ -567,39 +567,47 @@ def _on_restart_workload(self, event: RestartWorkloadEvent) -> None:
...
 
-        if event.restart_valkey and not self.charm.cluster_manager.is_healthy(
-            check_replica_sync=False
-        ):
-            self.charm.status.set_running_status(
+            if event.restart_valkey and not self.charm.cluster_manager.is_healthy(
+                check_replica_sync=False
+            ):
+                self.charm.status.set_running_status(
```
Mehdi-Bendriss: question here, should this be a running status or rather a regular state computed status?

src/events/tls.py
Mehdi-Bendriss: I'm not clear on why? isn't the sans_dns the only relevant bit for k8s, so why refresh the certs in that case?
if it is because we always add the sans_ips regardless of the substrate, that's what we should drop instead

tests/integration/ha/test_network_cut.py
```diff
@@ -81,40 +83,58 @@ async def test_network_cut_primary(  # noqa: C901
...
+    hostnames = get_cluster_hostnames(juju, APP_NAME)
 
     c_writes.tls_enabled = tls_enabled
     await c_writes.async_clear()
     c_writes.start()
 
     # Get the current primary unit
-    primary_ip = get_primary_ip(juju, APP_NAME, tls_enabled=tls_enabled)
-    assert primary_ip, "Failed to get primary endpoint from Juju status."
+    old_primary_ip = get_primary_ip(juju, APP_NAME, tls_enabled=tls_enabled)
```
Mehdi-Bendriss: This is not yet an `old` primary. The logic should be inverted such that this stays as `primary_host` (notice the name), and the new one should be `new_primary_host`

tests/integration/ha/test_network_cut.py
```diff
@@ -81,40 +83,58 @@ async def test_network_cut_primary(  # noqa: C901
...
 
-    logger.info("Identified container name for primary unit: %s", primary_hostname)
+    logger.info("Identified container name for primary unit: %s", old_primary_hostname)
     cut_network_from_unit(substrate, juju.model, machine_name, change_ip=change_ip)
 
+    # on K8s the controller is on a different namespace
+    if substrate == Substrate.VM:
+        controller_hostname = lxd_get_controller_hostname(juju)
+        assert not is_unit_reachable(
+            juju, controller_hostname, old_primary_hostname, substrate, number_of_retries=3
```
Mehdi-Bendriss: 2 points here: 
1. please use named arguments here, to make it clear `from= / to=`
2. why aren't we performing this check on k8s as well?

tests/integration/ha/test_network_cut.py
Mehdi-Bendriss: still relevant, we should in tests use `_host` (no matter what the helpers are), but `_host` makes it easy to reason with when reading the code cross substrates

tests/integration/ha/test_network_cut.py
```diff
@@ -153,32 +197,43 @@ async def test_network_cut_primary(  # noqa: C901
...
+        "Network restored to original primary unit %s. Verifying that all units can reach the original primary unit at %s...",
+        primary_unit_name,
+        old_primary_hostname,
+    )
     for unit in juju.status().apps[APP_NAME].units:
         if unit == primary_unit_name:
             continue
         assert is_unit_reachable(
-            juju, hostname_from_unit(juju, unit), primary_hostname, substrate
+            juju, hostname_from_unit(juju, unit), old_primary_hostname, substrate
```
Mehdi-Bendriss: missing named arguments

tests/integration/ha/helpers/helpers.py
```diff
@@ -361,3 +387,45 @@ def get_sans_from_certificate(certificate_path: str) -> dict[str, set[str]]:
...
+    controller_details = yaml.safe_load(raw_controller)
+    controller_name = model_details[juju.model]["controller-name"]
+
+    return [
+        machine.get("instance-id")
+        for machine in controller_details[controller_name]["controller-machines"].values()
+    ][0]
+
+
+def endpoint_in_sentinels(
```
Mehdi-Bendriss: is_endpoint_in_sentinels

tests/integration/ha/test_network_cut.py
```diff
@@ -145,6 +166,29 @@ async def test_network_cut_primary(  # noqa: C901
     assert number_of_replicas == NUM_UNITS - 2, (
         f"Expected {NUM_UNITS - 2} connected replicas, got {number_of_replicas}."
     )
+
+    logger.info(
```
Mehdi-Bendriss, line 182: These logs are a bit cumbersome, let's be consistent across the board, and rather log _before_ an operation / check happens (1 at a time), not after

tests/integration/ha/test_network_cut.py
```diff
@@ -194,6 +249,36 @@ async def test_network_cut_primary(  # noqa: C901
         f"Expected {NUM_UNITS - 1} connected replicas after network restoration, got {number_of_replicas}."
     )
 
+    for hostname in hostnames:
+        if hostname == new_unit_ip:
+            continue
+        if change_ip:
```
Mehdi-Bendriss: across the board, let's rename to `ip_change`

Mehdi-Bendriss (commented):
Thanks Smail! I have a few comments remaining from my side, that will improve readability.
We should also be consistent in the variable namings, let's agree that in tests we only use `_host` in variables (not `_ip`) - it makes it simpler to reason in k8s

src/events/tls.py
reneradoi: I would agree if it was only about refreshing the certs. But now we have half of the logic for IP changes in the `base_events` (such as applying configuration) and the other half in the `tls_events` (such as restarting and updating state). This is something we should avoid, only one event handler should be responsible for an IP change.

src/events/base_events.py
reneradoi: I think we should avoid checking the health in the status computation. This would delay all event handlers significantly.

tests/integration/ha/test_network_cut.py
reneradoi: I agree. Logging before an operation also makes it easier to follow the test execution and even serves as documentation for future maintenance of the tests.

src/events/base_events.py
skourta: We would have to add the lock release for every return if we do that. I prefer to keep it like this if this is not a hard opinion.

src/events/base_events.py
skourta: I agree with René. We try to keep the `get_statuses` as light as possible for all managers. I prefer to have this as a running status here and avoid having to do a health check on every event.

tests/integration/ha/test_network_cut.py
skourta: On k8s the controllers is not on the same namespace. It will not be able to communicate with the hostanme of the unit.

src/events/tls.py
skourta: Dropped IPs in K8s certificates

src/events/base_events.py
Mehdi-Bendriss: This is not what a running status was designed to serve. A running status is expected to report on long running **operations**, expected to be cleared by the time the operation completes - as it takes over the priority of all remaining statuses (except approved critical ones: upgrades) until it's cleared.
For example: 
- a synchronous backup within a hook ==> running status 
- an asynchronous backup cross multiple hooks ==> running status

skourta: It is my understanding that any status that might go across hook should be decalred as `async running status` as described in the [spec here](https://docs.google.com/document/d/1SV11ct-flQkc5BOYOeXgmPeglL8bVs-mDVkGjG20K48/edit?tab=t.0#heading=h.jv3enyriqiy1). That is why we declare this as an async running status.

reneradoi: I agree with Smail on this one. Since advanced statuses came out, we have adopted the pattern of `async running status` in our charms (etcd as well) for operations where user feedback is needed over hooks and cannot be computed, such as for example [here](https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/src/statuses.py#L87).

skourta: Summary from a discussion on a call with @Mehdi-Bendriss. Running statuses are statuses that run across hooks AND are behind a running (continuous) operation AND they take priority over other statuses. It is not enough for a status to run across hooks to be a running status. I will update the status and use a databag flag to recompute it when necessary.

Mehdi-Bendriss: small clarification: async* running statuses are the ones cross hooks, non-async* running statuses only last the current hook

src/events/base_events.py
```diff
@@ -322,6 +339,22 @@ def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
...
 
+            if self.charm.tls_manager.certificate_sans_require_update():
+                if self.charm.state.client_tls_relation:
+                    self.charm.tls_events.refresh_tls_certificates_event.emit()
+                    event.defer()
+                    return
+
+                self.charm.tls_manager.create_and_store_self_signed_certificate()
+
+            self.charm.state.unit_server.update(
```
Mehdi-Bendriss, line 350: this is fine, but i wonder why don't we use property setters directly instead of working with dictionaries?

tests/integration/ha/test_network_cut.py
```diff
@@ -124,70 +129,71 @@ async def test_network_cut_primary(  # noqa: C901
...
         with attempt:
             try:
-                new_primary_ip = get_primary_ip(juju, APP_NAME, tls_enabled=tls_enabled)
+                # we exclude the old primary ip because on k8s the unit is reachable by ip
+                # from outside k8s and is forming its own cluster
+                new_primary_ip = get_primary_ip(
+                    juju,
+                    APP_NAME,
+                    tls_enabled=tls_enabled,
+                    hostnames=[ip for ip in hostnames if ip != primary_ip],
```
Mehdi-Bendriss: why use `ip` when we are iterating over hostnames?

tests/integration/ha/test_network_cut.py
```diff
@@ -124,70 +129,71 @@ async def test_network_cut_primary(  # noqa: C901
...
                 break
             except ValueError as e:
                 logger.warning(f"Error getting primary IP after network cut: {e}")
-            logger.info("Waiting for new primary to be elected...")
+                logger.info("Waiting for new primary to be elected...")
+                raise
 
-    assert new_primary_ip and new_primary_ip != old_primary_ip, (
-        "Primary IP did not change after cutting network to the primary unit."
+    assert new_primary_ip and new_primary_ip != primary_ip, (
```
Mehdi-Bendriss, line 158: on naming, this was commented/missed a few times now, let's please use `_endpoint` instead of `_ip`.

Mehdi-Bendriss (changes requested):
No concerns on the code from my side, but my concern on variable naming remains. 

The naming is not consistent in the tests. 

We made the choice to make dual-substrate inclusive tests, I would like variables to reflect that and stop using `*_ip`, it is hard to reason with it when we read the tests with k8s in mind. 

**Please use the** `*_endpoint` **suffix instead of** `*_ip`.

src/events/base_events.py
skourta: We've been using this pattern across etcd and valkey. It makes it easy to udpate multiple attributes of the model when needed.
We can migrate to setters but I would prefer doing that in a different PR

Mehdi-Bendriss: Not needed, it's a matter of taste. And agreed, a dict does make sense when dealing with multiple properties

tests/integration/ha/test_network_cut.py
skourta: This is an IP for both substrates hence the name. It would be confusing to use `_endpoint` as reading `endpont` means it will be DN for k8s. We get the primary IP and construct the `endpoint` accordingly see line 180.

tests/integration/ha/test_network_cut.py
```diff
@@ -0,0 +1,298 @@
...
+    chaos_mesh,
+    c_writes,
+    c_writes_async_clean,
+) -> None:
+    """Cut the network to the primary unit and verify that a new primary is elected."""
+    if ip_change and substrate == Substrate.K8S:
+        pytest.skip("Changing IP is not applicable for k8s substrate.")
+
+    download_client_certificate_from_unit(juju, APP_NAME)
+    hostnames = get_cluster_hostnames(juju, APP_NAME)
```
skourta: Going through the comments I think this is misleading we should rename it to `get_cluster_ips`. WDYT @reneradoi @Mehdi-Bendriss? It actually resolves the IPs from juju statuses.

skourta:
> No concerns on the code from my side, but my concern on variable naming remains.
> 
> The naming is not consistent in the tests.
> 
> We made the choice to make dual-substrate inclusive tests, I would like variables to reflect that and stop using `*_ip`, it is hard to reason with it when we read the tests with k8s in mind.
> 
> **Please use the** `*_endpoint` **suffix instead of** `*_ip`.

I have not ignored your comments. The variables are named `_ip` if it an IP on both vm and k8s. We rely on juju statuses to get the IPs and then resolve the `endpoint`. I admit `get_cluster_hostnames` is wrongly named and should be `get_cluster_ips` for clarity. I use `_endpoint` when I expect the value to be the FQDN on the k8s substrate.

tests/integration/ha/test_network_cut.py
skourta: These are actually IPs and not hostnames. This is my bad I suggested to rename the function we use to get them and update the subsequent usages of them.

tests/integration/ha/test_network_cut.py
Mehdi-Bendriss: yes, or `get_unit_addresses`? since what the helper does is essentially:
```
if model_info.type == "kubernetes":
    return [unit.address for unit in status.get_units(app_name).values()]
return [unit.public_address for unit in status.get_units(app_name).values()]
```

Mehdi-Bendriss (dismissed):

reneradoi (dismissed):
Thank you Smail!

Mehdi-Bendriss (approved):

reneradoi (approved):
Re-approval

PR #26, opened 2026-03-19
[DPE-9318] Config option for certificate extra sans

reneradoi:
This PR adds a configuration option for `certificate-extra-sans`, which allows users to configure specific Subject Alternative Names (SAN) that they want to add to the default SANs in the client certificates. These can be IP addresses or hostnames.

The configuration option will only be considered if client TLS is enabled, not for internal peer TLS (self-managed certificates).  If the configuration value is invalid, a "blocked" status will be displayed to the user.


src/events/tls.py
```diff
@@ -305,17 +305,23 @@ def _on_secret_changed(self, event: ops.SecretChangedEvent) -> None:
 
     def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
         """Handle TLS related config changes."""
-        if not (secret_id := self.charm.config.get(TLS_CLIENT_PRIVATE_KEY_CONFIG)):
-            return
-
-        if not (private_key := self.charm.tls_manager.read_and_validate_private_key(secret_id)):
-            logger.error("Invalid private key provided, cannot update TLS certificates.")
```
Mehdi-Bendriss, line 312: Is this error log going too? This was a case we said (in etcd) we wanted to alert the user of through logging

Mehdi-Bendriss (commented):
Thanks René, good work. I have a comment on a missing error log, but should be good for me.

src/events/tls.py
reneradoi: Users are still alerted through the `blocked` status and the detailed error log in `read_and_validate_private_key()` for every condition that is checked.

src/managers/tls.py
```diff
@@ -304,6 +322,96 @@ def start_ca_rotation_if_required(
...
+        Returns:
+            bool: True if config value is valid, False if invalid.
+        """
+        if not (extra_sans_config := self.state.config.get("certificate-extra-sans")):
+            return True
+
+        extra_sans = [san.strip() for san in extra_sans_config.split(",")]
+
+        for san in extra_sans:
+            if not self._is_ip_address(san):
```
skourta: use a single if instead of nested ones

skourta (dismissed):
One minor comment else LGTM. Thank you @reneradoi

src/managers/tls.py
reneradoi: Thank you, updated.

src/events/tls.py
Mehdi-Bendriss: Yes, but the log should remain for ops reasons

reneradoi: Updated

Mehdi-Bendriss (approved):

Mehdi-Bendriss:
Note on failed tests, let's please stabilize those on a next PR

reneradoi:
> Note on failed tests, let's please stabilize those on a next PR

Addressed in https://github.com/canonical/valkey-operator/pull/35

skourta (approved):
LGTM thank you @reneradoi

PR #29, opened 2026-03-20
[DPE-9663] Support error propagation in TLS relation

reneradoi:
This PR adds support for error propagation from a TLS provider in accordance with [TE202](https://docs.google.com/document/d/1Czkzq8aF-ac74yobbopWr4fSV_YfED9QHhMdHeb61BI/edit?usp=sharing).

The operator observes the newly added `certificate-denied` event and displays a `blocked` status in case client TLS is enabled and the TLS provider is sending `request_errors` over the TLS relation interface.

The PR also adds integration test coverage with Vault as TLS provider.


src/events/tls.py
```diff
@@ -212,6 +216,14 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
             event.defer()
             return
 
+    def _on_certificate_denied(self, event: CertificateDeniedEvent) -> None:
+        """Handle the `certificate-denied` event from TLS provider."""
+        if event.certificate_signing_request in [
+            csr.certificate_signing_request
+            for csr in self.client_certificate.get_csrs_from_requirer_relation_data()
+        ]:
+            logger.error("Certificate request was denied: %s", event.error.message)
```
Mehdi-Bendriss, line 225: I have 2 questions here:
1. do we ever expect to receive from this hook a CSR we didn't create ourselves? if yes, how?
2. if yes, we should also log when a CSR is not ours

tests/integration/tls/test_certificate_options.py
```diff
@@ -98,3 +120,149 @@ def test_extra_sans_config_option(juju: jubilant.Juju) -> None:
...
+
+    logger.info("Remove relation with %s", TLS_NAME)
+    juju.remove_relation(f"{APP_NAME}:client-certificates", f"{TLS_NAME}:certificates")
+    juju.wait(
+        lambda status: are_agents_idle(status, APP_NAME, idle_period=30, unit_count=NUM_UNITS),
+        timeout=600,
+    )
+
+
+def test_initialize_vault(juju: jubilant.Juju, substrate: Substrate) -> None:
```
Mehdi-Bendriss, line 135: can you add inline comments in this test?

tests/integration/tls/test_certificate_options.py
```diff
@@ -98,3 +120,149 @@ def test_extra_sans_config_option(juju: jubilant.Juju) -> None:
...
+    result = await set_key(
+        hostnames=hostnames,
+        username=CharmUsers.VALKEY_ADMIN.value,
+        password=get_password(juju, user=CharmUsers.VALKEY_ADMIN),
+        tls_enabled=False,
+        key=TEST_KEY,
+        value=TEST_VALUE,
+    )
+    assert result == "OK", "Failed to write data without TLS"
+
```
Mehdi-Bendriss, line 271: Can you add an additional test where you remove the relation with Vault and the status goes away?

Mehdi-Bendriss (approved):
Thanks René, small comments on my side

src/events/tls.py
reneradoi: In theory we should not, but as we have the same safeguard in place for the certificate available event ... I've added some error logging for this case.

tests/integration/tls/test_certificate_options.py
reneradoi: updated

tests/integration/tls/test_certificate_options.py
reneradoi: Good idea - I found a nice improvement for the `tls-relation-broken` workflow on checking that.

src/events/tls.py
```diff
@@ -212,11 +216,30 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
...
+            "Certificate denied event received for unknown signing request: %s",
+            event.certificate_signing_request,
+        )
+
     def _on_tls_relation_broken(self, event: ops.RelationBrokenEvent) -> None:
         """Handle the `relation-broken` event."""
         if self.charm.app.planned_units() == 0 or self.charm.state.unit_server.is_being_removed:
             return
 
+        if not self.charm.state.unit_server.model.client_cert_ready:
```
skourta, line 238: Great find!

tests/integration/tls/test_certificate_options.py
```diff
@@ -98,3 +120,165 @@ def test_extra_sans_config_option(juju: jubilant.Juju) -> None:
...
+            status, APP_NAME, idle_period=30, unit_count=NUM_UNITS
+        ),
+        timeout=100,
+    )
+
+
+def _install_dependencies() -> None:
+    """Install dependencies for the test."""
+    # Install TLS Certificates interface library
+    subprocess.run(
```
skourta: nit: If it's just the one liner we can just add it in the test

tests/integration/test_charm.py
```diff
@@ -220,9 +220,9 @@ async def test_user_secret_permissions(juju: jubilant.Juju) -> None:
     )
 
     logger.info("Secret access will be granted now - wait for updated password")
+    juju.grant_secret(identifier=secret_name, app=APP_NAME)
```
skourta, line 223: Is this to fix the flakiness we've been noticing on CI?

skourta (approved):
Amazing work @reneradoi

tests/integration/test_charm.py
reneradoi: Yes. I was investigating these tests locally and found that the `update-status` triggered by `fast_forward` can happen before the secret gets granted, and then the purpose of `fast_forward` gets defeated. So I moved it.

tests/integration/tls/test_certificate_options.py
reneradoi: Updated.

PR #30, opened 2026-03-20
chore: improve stability of integration tests

reneradoi:
small fixes:
- `test_charm.py`: grant the secret in time before forcing the deferred event again
- `test_certificate_rotation.py`: wait for agents to be active/idle after CA has expired before checking access


reneradoi:
Was addressed with #35

PR #34, opened 2026-03-31
[DPE-9584] Support client relations

reneradoi:
This PR adds a client interface to the Valkey operator. Clients can integrate using `data-interfaces` in v0 or v1. The PR also adds a requirer charm for testing purposes as well as integration test coverage. Furthermore it migrates the code from using the charm lib for `data-interfaces` to using the Python package.

**UX:**
- integration happens over the `valkey-client` interface
- when a client has requested a resource that was not processed yet, a new `maintenance` status will be displayed
- if creating the requested resource fails, a `blocked` status will be displayed

**Design:**
- adds `ExternalClientsEvents` and `ExternalClientsManager` classes
- the Juju leader unit is responsible for processing resource requests and managing created users
- a resource request is only processed if the requested user does not yet exist (to avoid unnecessary operations)
- the client user credentials and permissions are stored in a new secret `external_client_users`
- the secret is used to populate the ACL files with required user names and permissions for Valkey and Sentinel
- a new field `client_user_epoch` is added to app and unit peer data; it is used to trigger / track required changes to ACLs on non-leader units (which do not receive `client-relation-changed` events due to the nature of `data-interfaces`)
- the Juju leader unit is also responsible for updating the provider data of the client interface; required updates are processed in the `peer-relation-changed` event handler of the `ExternalClientsEvents` class; peer relation changes are triggered on every relevant state change (TLS switchover, scaling, etc)

**Out of scope:**
- `certificate-transfer` interface
- optional mTLS, currently mTLS is enforced
- authentication over TLS, currently username and password are required

**Testing:**
- a requirer charm has been added for testing the client interface
- it can be configured to use either v0 or v1 of `data-interfaces`; depending on the configuration, the requirer charm instantiates the respective interface
- it uses `valkey-glide` for client connections to Valkey
- it supports TLS over the same TLS provider as Valkey (the `certificate_transfer` interface will be added in a separate PR)


src/managers/sentinel.py
```diff
@@ -123,6 +130,32 @@ def get_primary_ip(self) -> str:
         )
         raise ValkeyCannotGetPrimaryIPError("Could not determine primary IP from sentinels.")
 
+    def get_primary_endpoint(self) -> str:
```
skourta, line 133: This feels hacky.
You can either use `get_primary_addr_by_name` which returns the full endpoint. Or implement the logic here and update the `get_primary_ip` to use this one and take only the first value of the instance.

src/managers/sentinel.py
```diff
@@ -123,6 +130,32 @@ def get_primary_ip(self) -> str:
...
+
+        return f"{primary_address}:{port}"
+
+    def get_replica_endpoints(self) -> str:
+        """Get the endpoints of all replica nodes, consisting of address and port."""
+        port = TLS_PORT if self.state.unit_server.is_tls_enabled else CLIENT_PORT
+        client = self._get_sentinel_client()
+
+        replica_list = client.replicas_primary(hostname=self.state.endpoint)
+        return ",".join(sorted([f"{replica['ip']}:{port}" for replica in replica_list]))
```
skourta, line 146: Why not use the port returned by sentinel?

src/common/client.py
```diff
@@ -152,6 +152,30 @@ def info_persistence(self, hostname: str) -> dict[str, str] | None:
...
+
+        Returns:
+            dict[str, str] | None: The server information retrieved from Valkey.
+
+        Raises:
+            ValkeyWorkloadCommandError: If the CLI command fails to execute.
+        """
+        output = self.exec_cli_command(["info", "server"], hostname=hostname, json_output=False)
+        values = {}
+        if not output.strip():
```
skourta, line 168: This would certainly mean an error no? Let's raise and avoid returning None.

src/core/models.py
```diff
@@ -222,6 +230,14 @@ def internal_users_credentials(self) -> dict[str, str]:
                 passwords[user.value] = password
         return passwords
 
+    @property
+    def external_users_credentials(self) -> dict[str, dict[str, str]] | None:
+        """Retrieve the user credentials for external clients from the state and return as dict."""
+        if not self.model or not (external_clients := self.model.external_client_users):
+            return None
+
+        return json.loads(external_clients)
```
skourta, line 249: nit: can we have this modeled as well? 
I was looking at the manager and found myself wondering what are the values stored inside.

src/managers/config.py
```diff
@@ -168,31 +168,56 @@ def set_acl_file(self, passwords: dict[str, str] | None = None) -> None:
...
-            str: ACL line for the user.
+            str: ACL line for the internal user.
         """
         passwords = passwords or self.state.cluster.internal_users_credentials
         if not (password := passwords.get(user.value, "")):
             raise ValueError(f"No password found for user {user}")
         password_hash = hashlib.sha256(password.encode("utf-8")).hexdigest()
         return f"user {user.value} on #{password_hash} {CHARM_USERS_ROLE_MAP[user]}\n"
 
+    def _get_client_user_acl_lines(self, for_sentinel: bool = False) -> str | None:
```
skourta:
```suggestion
    def _get_client_user_acl_lines(self, for_sentinel: bool = False) -> str:
```

src/managers/external_clients.py
```diff
@@ -0,0 +1,129 @@
...
+                del external_client_users[username]
+
+        self.state.cluster.update({"external_client_users": external_client_users})
+
+    def does_username_exist(self, username: str) -> bool:
+        """Check if a username already exists."""
+        if not (external_client_users := self.state.cluster.external_users_credentials):
+            return False
+
+        if external_client_users.get(username):
```
skourta, line 77: We can simplify a little:
```
return username in (self.state.cluster.external_users_credentials or {})
```

tests/integration/clients/requirer-charm/src/charm.py
```diff
@@ -0,0 +1,296 @@
...
+
+        key = str(event.params.get("key", ""))
+        value = str(event.params.get("value", ""))
+        user = str(event.params.get("user", ""))
+        if not key or not value or not user:
+            event.fail("Parameters key, value and user are required.")
+            event.set_results({"ok": False})
+            return
+
+        client = ValkeyClient(
```
skourta: nit: This can be a property or helper

tests/integration/clients/requirer-charm/src/charm.py
```diff
@@ -0,0 +1,296 @@
...
+
+    def _on_get_credentials_action(self, event: ops.ActionEvent) -> None:
+        """Return the credentials an action response."""
+        if not self.valkey_relation:
+            event.fail("The action can be run only after relation is created.")
+            event.set_results({"ok": False})
+            return
+
+        credentials = self.credentials
+        usernames = ",".join(list(credentials.keys()))
```
skourta, line 264: Do we only return the usernames not the passwords?

skourta (commented):
Thank you @reneradoi.
I've left some minor comments.
I will run some manual tests next.

src/managers/sentinel.py
reneradoi: In the future, we will need to return the K8s service endpoint for the primary service here instead of the primary's address. This method has this in mind already, the logic then will be something like this:

```
if substrate == K8s:
    return primary_service_endpoint
else:
    return primary_address
```

On another note: The port returned is the one for peer communication and it might be different from the port for client communication (internal peer TLS vs. no TLS for clients), therefor we cannot use the returned information.

src/managers/sentinel.py
reneradoi: The port returned is the one for peer communication and it might be different from the port for client communication (internal peer TLS vs. no TLS for clients), therefor we cannot use the returned information.

src/common/client.py
reneradoi: The implementation is in line with the `info_persistence` method some lines above. Do you think we should change both or keep them similar?

src/managers/config.py
reneradoi: Thank you, updated. This was a leftover from an earlier implementation of this method, which I refactored then.

src/managers/external_clients.py
reneradoi: We would lose the log message then, which was very helpful for debugging.

tests/integration/clients/requirer-charm/src/charm.py
reneradoi: Yes, as we don't need them for the tests. I could add them of course, but why?

tests/integration/clients/requirer-charm/src/charm.py
reneradoi: Good idea.

src/core/models.py
reneradoi: What do you mean with modeled? It's a dict structure stored into a secret as str. I've added an example in the doc string for better understanding.

reneradoi:
Note: The integration tests are failing after the latest push due to unrelated errors setting up the environment:

```
error: cannot install "charmcraft": cannot query the store for
       updates: got unexpected HTTP status code 503 via POST to
       "https://api.snapcraft.io/v2/snaps/refresh"
```

Please refer to this CI run to see the tests in action: https://github.com/canonical/valkey-operator/actions/runs/24089167054

src/common/client.py
skourta: Yes I think we should change both. It's better for types. I do not like how I implemented that one. Let's change it if possible.

tests/integration/clients/requirer-charm/src/client.py
```diff
@@ -0,0 +1,79 @@
...
+        self,
+        username: str,
+        password: str,
+        host: str,
+        port: int,
+        tls_cert: bytes | None,
+        tls_key: bytes | None,
+        tls_ca_cert: bytes | None,
+    ):
+        self.host = host
```
skourta, line 33: Shouldn't we pass all endpoints to glide? Iin case the primary changes.

src/events/external_clients.py
```diff
@@ -0,0 +1,293 @@
...
+
+        self.valkey_provides.set_responses(event.relation.id, responses)
+        self.charm.state.cluster.update({"client_user_epoch": time.time()})
+        self.charm.state.statuses.delete(
+            ExternalClientsStatuses.USER_SETUP_FAILED.value,
+            scope="unit",
+            component=self.charm.client_manager.name,
+        )
+
+    def _on_peer_relation_changed(self, event: ops.RelationChangedEvent) -> None:
```
skourta, line 161: I think this is the best we can do for `primary_endpoint` changing right?

tests/integration/clients/requirer-charm/src/client.py
reneradoi: We only need the primary endpoint at a time. Relation data is updated from provider side on changes, and the requirer charm will always pick up the updated information from relation data when creating a client connection.

src/events/external_clients.py
reneradoi: Yes, this covers all "known" changes, such as scaling up and down, enabling or disabling TLS, TLS CA rotation and such. The "unknown" changes (failover initiated by Sentinel without a charm event) will have to be covered with a separate solution ([ticket](https://warthogs.atlassian.net/browse/DPE-9752) already planned).

tests/integration/clients/requirer-charm/pyproject.toml
```diff
@@ -0,0 +1,16 @@
...
+[tool.poetry]
+package-mode = false
+requires-poetry = ">=2.0.0"
+
+[tool.poetry.dependencies]
+python = "^3.12"
+ops = "3.5.0"
+pydantic = ">2.12.3"
+charmlibs-interfaces-tls-certificates = ">1.0"
+dpcharmlibs-interfaces = ">=1.0.2"
```
Mehdi-Bendriss, line 13: nice

src/core/models.py
```diff
@@ -72,6 +79,7 @@ class PeerUnitModel(PeerModel):
     client_cert_ready: bool = Field(default=False)
     tls_ca_rotation: str = Field(default="")
     tls_certificate_expiring: bool = Field(default=False)
+    client_user_epoch: float = Field(default=0)
```
Mehdi-Bendriss, line 82: what does this represent?

src/events/external_clients.py
```diff
@@ -0,0 +1,293 @@
...
+            requests = [event.request]
+        else:
+            requests = event.requests
+
+        responses = []
+        for request in requests:
+            username = self.charm.client_manager.get_username(
+                event.relation.id, request.request_id
+            )
+            if self.charm.client_manager.does_username_exist(username):
```
Mehdi-Bendriss, line 104: can this realistically ever happen? or it would in the case of username through the common name?
In which case, this would warrant a BlockedStatus (for a lack of error propagation in the relation due to it not being ported to v1 yet) as the user would not have any feedback from the relation.

src/events/external_clients.py
```diff
@@ -0,0 +1,293 @@
...
+            self.charm.client_manager.add_managed_user(username, password, request.resource)
+
+            response = next(
+                (
+                    res
+                    for res in self.valkey_provides.responses(event.relation, ValkeyResponseModel)
+                    if res.request_id == request.request_id
+                ),
+                None,
+            ) or ValkeyResponseModel(
```
Mehdi-Bendriss, line 119: can you explain the `or`? when would we have `None` as a result of `valkey_provides.responses` ?

src/events/external_clients.py
```diff
@@ -0,0 +1,293 @@
...
+            self.charm.config_manager.set_sentinel_acl_file()
+            # todo: request rolling restart once https://github.com/canonical/valkey-operator/pull/23 is merged
+            self.charm.sentinel_manager.restart_service()
+        except (
+            ValkeyACLLoadError,
+            ValkeyServicesFailedToStartError,
+            ValkeyWorkloadCommandError,
+        ) as e:
+            logger.error(e)
+            self.charm.status.set_running_status(
```
Mehdi-Bendriss: This shouldn't be a running status - a regular blocked status should do

src/literals.py
```diff
@@ -40,7 +40,8 @@
 QUORUM_NUMBER = 2
 INTERNAL_USERS_PASSWORD_CONFIG = "system-users"
 INTERNAL_USERS_SECRET_LABEL_SUFFIX = "internal_users_secret"
-INTERNET_CERTS_SECRET_LABEL_SUFFIX = "internal_certificates_secret"
```
Mehdi-Bendriss, line 43: 😄

tests/integration/clients/requirer-charm/src/client.py
Mehdi-Bendriss: The only use case where the new primary doesn't get broadcasted is during a non orchestrated failover, this will be possible when we have the topology watchdog process in place

tests/integration/clients/requirer-charm/src/charm.py
```diff
@@ -0,0 +1,292 @@
...
+                return {"": None}
+
+            return {
+                self.valkey_interface.fetch_relation_field(
+                    self.valkey_relation.id, "username"
+                ): self.valkey_interface.fetch_relation_field(self.valkey_relation.id, "password")
+            }
+
+        remote_responses = self.remote_responses
+        if not remote_responses:
```
Mehdi-Bendriss: nit, walrus - and in subsequent properties where applies

Mehdi-Bendriss (commented):
Thanks René! Great work. 
I have a few small comments and questions

src/common/client.py
reneradoi: refactored

src/core/models.py
reneradoi: The `client_user_epoch` is used to store (on the unit databag) when a unit has last updated user ACLs and (on the app databag) when the leader has last processed new client user requests. Non-leader units do not receive the resource requests from `data-interfaces`, so they always need to reconcile ACLs in case new users where added.

By storing this timestamp, we can optimize reconciling on the follower units, to only process again when necessary. This optimizes performance a lot, both in charm events as in updating and reloading ACL config on Valkey and Sentinel.

src/events/external_clients.py
reneradoi: Yes it can happen. A new `resource_requested` event [is issued](https://github.com/canonical/data-platform-charmlibs/blob/main/interfaces/src/dpcharmlibs/interfaces/handlers.py#L504) by `data-interfaces` on `relation-changed` events. I've noticed this in the relation establishing process during development, because I wondered why users where "added" several times when a new relation was created. By adding this check, only really "new" resource requests are processed (new as in user has not been created yet).

src/events/external_clients.py
reneradoi: In case of new requests, if no response was previously added to relation data.

src/events/external_clients.py
reneradoi: Agree, updated.

tests/integration/clients/requirer-charm/src/charm.py
reneradoi: updated

tests/integration/clients/requirer-charm/src/charm.py
skourta: I asked because we do in the other charms in the get_credentials actions. I am ok with only the usernames

reneradoi: For the data integrator charm, we will of course output the password, too (see dev branch [here](https://github.com/canonical/data-integrator/blob/add-valkey-interface/src/charm.py#L458)).

skourta (approved):
Thank you @reneradoi. LGTM. Let's try to get the a green client relations test on the CI before we merge (I know it's a bit challenging at the moment 😢).

src/events/external_clients.py
Mehdi-Bendriss: This is worth opening an issue on the dpcharmlibs repo

Mehdi-Bendriss (approved):

src/events/external_clients.py
reneradoi: Agree, will discuss with Neha.

PR #35, opened 2026-04-02
fix: unblock CI by switching to self-hosted runners

reneradoi:
Due to https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2146972, integration tests randomly fail in CI to a state that we are blocked on any PR. While a possible workaround would be to switch to `22.04` runners on Github, this would require a lot of workarounds to be able to use `valkey-glide`, therefor this PR switches the integration tests in CI to self-hosted runners instead.

The PR also includes some minor improvements/fixes for two unstable integration tests:
- `test_charm.py`: grant the secret in time before forcing the deferred event again
-  `test_certificate_rotation.py`: wait for agents to be active/idle after CA has expired before checking access

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -102,20 +102,9 @@ jobs:
     runs-on: ${{ matrix.job.runner }}
     timeout-minutes: 226 # Sum of steps `timeout-minutes` + 5
     steps:
-      - name: Free up disk space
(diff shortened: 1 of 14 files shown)


skourta (approved):
Can we add a TODO to move back when the issue is fixed?
Thank you @reneradoi

Mehdi-Bendriss (approved):

PR #36, opened 2026-04-07
Bugfix: avoid errors for building IP and DNS SANs

reneradoi:
In some cases, it can happen that the charm code tries to build SANs IP and DNS names before the peer relation model is ready. This results in uncaught errors (self-resolving), example: https://github.com/canonical/valkey-operator/actions/runs/23968250108/job/69912976655#step:11:162

The PR adjusts the logic for building SAN names by returning early in these cases. Thereby it also fixes a bug where the unit name was removed from the SAN DNS entries.

### DIFF
--- a/src/managers/tls.py
+++ b/src/managers/tls.py
@@ -97,7 +97,7 @@ class TLSManager(ManagerStatusProtocol):
         """Build the SANs IP for the TLS certificate."""
         sans_ip = set()
 
-        if not self.state.peer_relation:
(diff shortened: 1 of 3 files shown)


skourta (approved):
Thank you @reneradoi.I have ran the TLS tests with the other fix in my PR already. but let's wait for the tests to confirm.

Mehdi-Bendriss (approved):

PR #39, opened 2026-04-13
patch: move Glide to the requirer charm

skourta:
This pull request introduces significant enhancements to the integration test client charm for Valkey, focusing on improved testing capabilities, especially around continuous writes and flexible client configuration. It also includes some internal refactoring and minor fixes. The most important changes are grouped below.

**Integration test charm enhancements:**

* Added a new `continuous_writes.py` daemon script to support robust, background continuous-write testing, including state tracking, atomic state updates, config hot-reload, and graceful shutdown. This enables advanced data consistency and disruption testing for Valkey deployments.
* Extended the `charmcraft.yaml` for the integration test requirer charm with several new actions: `execute`, `start-continuous-writes`, `get-continuous-writes-state`, `assert-continuous-writes-increasing`, `clear-continuous-writes`, `stop-continuous-writes`, and `seed-data`. These actions facilitate advanced integration and consistency testing scenarios.
* Added support for a new `glide-config` config option, allowing the charm to connect directly to Valkey using arbitrary endpoints and credentials, independent of relations.

**Valkey client improvements:**

* Refactored the `ValkeyClient` class to support multiple endpoints, optional password, and improved TLS configuration. Added new methods for seeding data (`seed_data`) and executing arbitrary commands (`execute_command`), with robust result handling for different Valkey response types. [[1]](diffhunk://#diff-78405b5305e4767a2c53df3eab15f7c007389f8a97915c4c008c7963e7b30523L26-R34) [[2]](diffhunk://#diff-78405b5305e4767a2c53df3eab15f7c007389f8a97915c4c008c7963e7b30523L77-R130)

**Charm core/status logic fixes:**

* Changed how statuses are set and retrieved in the core charm code, removing use of the `StartStatuses` enum for running status and updating status management to use the new `unit_server.update` method and improved status list retrieval. [[1]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL200-R201) [[2]](diffhunk://#diff-20b77aa4083c3162fac7cd352e1b8ceef186a967a37f906277cb77346967c113L131-R135) [[3]](diffhunk://#diff-074beea58df119261178a25e474cd0c240fa20f1c6fb6990867d357676e96cefL66)
* Fixed the order of TLS status checks in the TLS manager to ensure correct status reporting.

**Miscellaneous:**

* Minor formatting and consistency fixes in `charmcraft.yaml` and import statements. [[1]](diffhunk://#diff-c3a960460030f011afb5d67251faf1b2341e7d0246ec8c76cb17077abcf6a1b3L4) [[2]](diffhunk://#diff-c3a960460030f011afb5d67251faf1b2341e7d0246ec8c76cb17077abcf6a1b3L48-R47) [[3]](diffhunk://#diff-977a7d8c04cab4f9aaa5659892eadd9b991f51cd8f5814ecb51e3162bf23e31eL36-R36) [[4]](diffhunk://#diff-78405b5305e4767a2c53df3eab15f7c007389f8a97915c4c008c7963e7b30523R6-R8)

These changes collectively provide much more powerful tools for integration and disruption testing of Valkey, as well as cleaner and more robust status management in the charm.


tests/integration/clients/requirer-charm/src/charm.py
```diff
@@ -207,24 +295,27 @@ def private_key(self) -> str | None:
...
+            raise ValueError("No endpoint available.")
+        if not self.credentials:
+            raise ValueError("No credentials available.")
+        if self.tls_enabled and (
+            not self.certificate or not self.private_key or not self.tls_ca_cert
+        ):
+            raise ValueError("TLS is enabled but certificates are not yet available.")
         return ValkeyClient(
-            username="" if self.config.get("use-certificate-auth") else user,
-            password="" if self.config.get("use-certificate-auth") else self.credentials.get(user),
```
reneradoi, line 212: This condition is required to test password-less authentication in the integration test. It should stay or be covered otherwise.

tests/integration/clients/requirer-charm/src/charm.py
```diff
@@ -207,24 +295,27 @@ def private_key(self) -> str | None:
...
-            host=self.primary_endpoint.split(":")[0],
-            port=int(self.primary_endpoint.split(":")[1]),
-            tls_cert=self.certificate.encode() if self.use_mtls else None,
-            tls_key=self.private_key.encode() if self.use_mtls else None,
```
reneradoi, line 216: This condition is required to test connection with and without mTLS in the integration test. It should stay or be covered otherwise.

tests/integration/clients/requirer-charm/src/client.py
```diff
@@ -49,9 +53,9 @@ async def create_client(self) -> GlideClient:
         )
 
         client_config = GlideClientConfiguration(
-            [NodeAddress(host=self.host, port=self.port)],
-            use_tls=True if self.tls_ca_cert else False,
```
reneradoi, line 53: This should remain, otherwise it will not be possible to test TLS without mTLS anymore.

reneradoi (changes requested):
I didn't do a full review yet, but something important to note: So far, the testing logic for various use cases of the client relation was encoded in the requirer charm's client setup. This would now be lost. The integration tests still pass because they essentially always connect the same way (which works), but we don't test the different scenarios anymore (see [here](https://github.com/canonical/valkey-operator/blob/9/edge/tests/integration/clients/test_client_relation.py#L5-L13)).

tests/integration/clients/requirer-charm/src/charm.py
```diff
@@ -275,6 +366,38 @@ def _on_get_action(self, event: ops.ActionEvent) -> None:
             event.fail(f"Failed to read data: {e}")
             logger.error("Failed to read data: %s", e)
 
+    def _on_execute_action(self, event: ops.ActionEvent) -> None:
+        """Handle execute action."""
+        command = str(event.params.get("command", ""))
+        if not command:
```
reneradoi: nit: combine with `:=`

tests/integration/clients/requirer-charm/src/charm.py
```diff
@@ -305,6 +660,54 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
...
+    def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
+        """Hot-reload the continuous-writes daemon when glide-config changes."""
+        self.refresh_tls_certificates_event.emit()
+
+        if not self._use_config or not CWPath.PID.value.exists():
+            return
+
+        try:
+            current_config = DaemonConfig.from_file(CWPath.CONFIG.value)
+        except Exception:
```
reneradoi: Maybe log a message in this case?

reneradoi (commented):
Amazing work, just a few minor nits.

tests/integration/clients/requirer-charm/src/charm.py
```diff
@@ -114,8 +169,24 @@ def remote_responses(self) -> list[ResourceProviderModel] | None:
         ).requests
 
     @property
-    def credentials(self) -> dict[str | None, str | None]:
-        """Retrieve the client credentials provided by Valkey."""
+    def _glide_config(self) -> GlideConfig | None:
+        """Parse the glide-config JSON option, or None if not set."""
+        raw = str(self.config.get("glide-config", "")).strip()
+        if not raw:
```
Mehdi-Bendriss: nit, walrus

tests/integration/clients/requirer-charm/src/charm.py
```diff
@@ -137,7 +208,10 @@ def credentials(self) -> dict[str | None, str | None]:
 
     @property
     def primary_endpoint(self) -> str | None:
-        """Retrieve the write-endpoints provided by Valkey."""
+        """Retrieve the write-endpoints from config or relation."""
+        if (cfg := self._glide_config) is not None:
```
Mehdi-Bendriss: Applies to subsequent usages
```suggestion
        if cfg := self._glide_config:
```

tests/integration/clients/requirer-charm/src/continuous_writes.py
```diff
@@ -0,0 +1,259 @@
...
+    stop = asyncio.Event()
+    reload = asyncio.Event()
+
+    loop = asyncio.get_running_loop()
+    loop.add_signal_handler(signal.SIGTERM, stop.set)
+    loop.add_signal_handler(signal.SIGINT, stop.set)
+    loop.add_signal_handler(signal.SIGUSR1, reload.set)
+
+    counter, count = await _initial_count(config)
+    last_written = counter - 1
```
Mehdi-Bendriss, line 229: why not keep it simple and have a counter that starts at 0? so we don't have to substract 1?

tests/integration/clients/requirer-charm/src/continuous_writes.py
```diff
@@ -0,0 +1,259 @@
...
+    logger.info(
+        "Starting continuous writes from counter=%d (existing list len=%d)", counter, count
+    )
+
+    while not stop.is_set():
+        try:
+            async with glide_client(config) as client:
```
Mehdi-Bendriss: we should not create a new client every iteration, we should mimic production apps that reuse the same client instance refer to here https://github.com/canonical/opensearch-operator/blob/2/edge/tests/integration/ha/continuous_writes.py#L267

If it is due to the change of hosts, we should instead expose an `update` action or a config option, that compares the config of previous glide client and the hosts passed to the charm and only then recreate the client - see here https://github.com/canonical/opensearch-operator/blob/2/edge/tests/integration/ha/continuous_writes.py#L270-L273

tests/integration/clients/requirer-charm/src/continuous_writes.py
```diff
@@ -0,0 +1,259 @@
...
+                raise RuntimeError("LPUSH returned 0/None")
+            last_written = counter
+            count = new_len
+            _write_state_atomic(last_written, count)
+            logger.info("Wrote %d (list len=%d)", counter, count)
+        except Exception as exc:
+            # Write failed — log and skip without updating last_written.
+            # counter still increments so a gap is introduced in the sequence,
+            # making failed writes detectable during consistency checks.
+            logger.warning("Write failed for counter=%d: %s", counter, exc)
```
Mehdi-Bendriss: Shouldn't the exception not be ignored but instead same value retried with the same tried/failed written_value? (re discussion from a few weeks ago and similar to opensearch)
https://github.com/canonical/opensearch-operator/blob/2/edge/tests/integration/ha/continuous_writes.py#L283-L285

tests/integration/clients/requirer-charm/src/glide_helpers.py
```diff
@@ -0,0 +1,138 @@
...
+    NodeAddress,
+    ReadFrom,
+    ServerCredentials,
+    TlsAdvancedConfiguration,
+)
+
+# Maps each Glide class to the set of fields that should be serialized.
+# Tuple values like ``(list, NodeAddress)`` are documentation only — the
+# serialize/deserialize logic recurses structurally, not via this type info.
+SCHEMA: dict[type, dict[str, Any]] = {
```
Mehdi-Bendriss, line 31: Is this and `serialize_*` needed in the client charm? i thought this was needed only on the integration test side. Since the client charm only cares about the deserialization

Mehdi-Bendriss (commented):
I currently reviewed the continuous writes part / test charm, I'll come back to review the tests part

tests/integration/clients/requirer-charm/src/continuous_writes.py
skourta: We do have a counter that starts at 0. But this handles the case where we had existing values.

tests/integration/clients/requirer-charm/src/continuous_writes.py
skourta: I do not see the benefit from doing that. Not retrying gives us the additional benefit of knowing if there was a gap of writes. I do not have a hard opinon about this though.

tests/integration/clients/requirer-charm/src/continuous_writes.py
skourta: Updated. I added recreating the client on error in case we have connection issues

tests/integration/clients/requirer-charm/src/glide_helpers.py
skourta: We need the schema to know which `_GLIDE_CLASSES` are used. I removed the serialise function

tests/integration/clients/requirer-charm/src/client.py
skourta: Could you please double check the conditions are now good?

tests/integration/clients/requirer-charm/src/charm.py
```diff
@@ -89,6 +122,28 @@ def __init__(self, framework: ops.Framework):
...
+            self._on_get_continuous_writes_state_action,
+        )
+        framework.observe(
+            self.on.assert_continuous_writes_increasing_action,
+            self._on_assert_continuous_writes_increasing_action,
+        )
+        framework.observe(self.valkey_interface.on.endpoints_changed, self._on_endpoints_changed)
+        framework.observe(self.on.config_changed, self._on_config_changed)
         framework.observe(self.on.get_credentials_action, self._on_get_credentials_action)
         framework.observe(self.valkey_interface.on.endpoints_changed, self._on_endpoints_changed)
```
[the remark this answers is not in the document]
reneradoi: +1, some observers are duplicates.

tests/integration/clients/requirer-charm/src/client.py
reneradoi: Yes, fine now.

tests/integration/clients/requirer-charm/src/charm.py
reneradoi: fine now.

tests/integration/clients/requirer-charm/src/charm.py
reneradoi: fine now

reneradoi (commented):
The TLS configuration for the client in the requirer charm is fine now, thank you for fixing it.

Mehdi-Bendriss (dismissed):
Thanks Smail!

Mehdi-Bendriss (approved):

reneradoi (approved):
Great work!

PR #44, opened 2026-04-15
fix: CA rotation integration test

reneradoi:
This PR includes a few minor fixes and improvements for the CA rotation integration test, that recently always fails on K8s, presumably because it just takes longer. When K8s tests are executed locally, they always pass.

The changes are:
- catch `ValkeyCannotGetPrimaryIPError` that might be raised by `sentinel_manager.get_primary_ip()` (missed after the error handling was changed on the method), see [this CI run](https://github.com/canonical/valkey-operator/actions/runs/24444439636/job/71433499795#step:12:13271)
- no longer deploy with TLS from the start, to avoid TLS certs to become invalid while startup still in progress, see [this CI run](https://github.com/canonical/valkey-operator/actions/runs/24444439636/job/71455753256#step:11:15)
- simplify the test by removing the disabling and re-enabling of TLS in between; there is actually no need to do that, we can just change the configuration of the TLS provider
- adjust validity time of the certificates in the test to allow for reloading before the certificate actually expires

Unrelated change:
- update `core26` snap to `beta` channel on VM

### DIFF
--- a/tests/integration/tls/test_certificate_rotation.py
+++ b/tests/integration/tls/test_certificate_rotation.py
@@ -32,7 +32,7 @@ NUM_UNITS = 3
 TEST_KEY = "test_key"
 TEST_VALUE = "test_value"
 CERTIFICATE_EXPIRY_TIME = 320
-CA_EXPIRY_TIME = 500
+CA_EXPIRY_TIME = 380
 
 
 def _prepare_units_for_ca_expiration_test(juju: jubilant.Juju) -> None:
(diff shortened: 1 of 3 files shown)


tests/integration/tls/test_certificate_rotation.py
```diff
@@ -98,6 +102,7 @@ async def test_certificate_expiration(juju: jubilant.Juju) -> None:
         old_client_certificate = file.read()
     assert old_client_certificate, "Failed to get current client certificate"
 
+    _prepare_units_for_ca_expiration_test(juju)
```
skourta: Wouldn't this only change the expiration check when we receive the next certs?

reneradoi: Yes that is correct, thank you for noticing. I've moved it to before integrating with the TLS provider, by that we can safe some waiting time.

skourta (approved):
Much appreciated fix! Thank you @reneradoi

Mehdi-Bendriss (approved):

PR #46, opened 2026-04-17
[DPE-9751] add K8s services for primary and replica endpoints

reneradoi:
This PR adds Kubernetes services for primary and replica endpoints, which can be used by clients to connect to Valkey.

**Design:**
- add new class `K8sClient`, that provides methods for creating/updating K8s services and for adding labels to pods
- add new method `reconcile_k8s_services()` to the Sentinel manager class that calls the K8s client methods, based on the current deployment and primary/replica setup
- the services are owned by the stateful set of the application, meaning they will be removed on application removal
- trigger the `reconcile_k8s_services()` on the leader unit on any peer relation change, to update the ports if required
- publish the K8s services instead of unit endpoints to client relation data on K8s

**Other changes:**
- add the services to DNS SANs for the TLS certificate requests
- update rust to `1.94.0` to allow building `lightkube`

Integration testing is covered by the existing tests for client relations. The requirer-charms will now connect via the  services in the tests on K8s.


src/common/k8s_client.py
```diff
@@ -0,0 +1,103 @@
...
+                service.spec.ports = [service_port]
+                self.client.patch(Service, service_name, service, patch_type=PatchType.MERGE)
+                logger.info("Updated Kubernetes service %s to port %s", service_name, port)
+            return
+        except ApiError as e:
+            # 404 will be raised if service does not exist yet
+            if e.status.code != 404:
+                raise KubernetesClientError from e
+
+        pod0 = self.client.get(
```
Mehdi-Bendriss: This may raise an `ApiError` as well

src/managers/tls.py
```diff
@@ -141,6 +147,10 @@ def build_sans_dns(self) -> frozenset[str]:
         sans_dns.add(self.state.unit_server.unit_name.replace("/", ""))
         sans_dns.add(self.state.hostname)
 
+        if self.state.substrate == Substrate.K8S:
+            sans_dns.add(f"{self.state.model.app.name}-{K8sService.PRIMARY.value}")
```
Mehdi-Bendriss, line 151: nice

Mehdi-Bendriss (commented):
Thanks René! Really clean, only 1 comment from my side

src/common/k8s_client.py
reneradoi: Thank you!

src/managers/sentinel.py
```diff
@@ -46,6 +49,12 @@ def __init__(self, state: ClusterState, workload: WorkloadBase):
         self.workload = workload
         self.admin_user = CharmUsers.SENTINEL_CHARM_ADMIN.value
 
+        if self.state.substrate == Substrate.K8S:
+            self.k8s_client = K8sClient(
```
skourta, line 54: Can we add `self.k8s_client: K8sClient | None = None` before this so we get the attribute and is set to None on VM?

src/common/k8s_client.py
```diff
@@ -0,0 +1,106 @@
...
+            return
+        except ApiError as e:
+            # 404 will be raised if service does not exist yet
+            if e.status.code != 404:
+                raise KubernetesClientError from e
+
+        try:
+            pod0 = self.client.get(
+                res=Pod,
+                name=self.app_name + "-0",
```
skourta, line 53: Is this ok? Are we always assured pod-0 will exist?

src/common/k8s_client.py
```diff
@@ -0,0 +1,106 @@
...
+        try:
+            service = self.client.get(res=Service, name=service_name, namespace=self.namespace)
+            if service.spec.ports != [service_port]:
+                service.spec.ports = [service_port]
+                self.client.patch(Service, service_name, service, patch_type=PatchType.MERGE)
+                logger.info("Updated Kubernetes service %s to port %s", service_name, port)
+            return
+        except ApiError as e:
+            # 404 will be raised if service does not exist yet
+            if e.status.code != 404:
```
skourta, line 47: Let's log here an error as well like we do in other places

src/events/external_clients.py
```diff
@@ -178,15 +179,26 @@ def _on_bulk_resources_requested(
...
+        """
+        if not self.charm.state.unit_server.is_started:
+            return
+
+        if self.charm.unit.is_leader():
+            try:
+                self.charm.sentinel_manager.reconcile_k8s_services()
+            except (KubernetesClientError, ValkeyCannotGetPrimaryIPError) as e:
+                logger.error("Error updating Kubernetes services: %s", e)
+                event.defer()
```
skourta, line 200: Should we defer? We usually already have a storm of peer relation events. Is this necessary in all cases?

src/events/external_clients.py
```diff
@@ -178,15 +179,26 @@ def _on_bulk_resources_requested(
...
-        ):
+        """Handle peer relation changes in regard to external client relations.
+
+        This handler catches all changes from scaling operations, TLS switchover, TLS CA rotation,
+        IP changes, etc.
+        """
+        if not self.charm.state.unit_server.is_started:
+            return
+
+        if self.charm.unit.is_leader():
```
skourta: I would prefer we move the k8s substrate check to the event from the `reconcile_k8s_services` manager method

src/managers/sentinel.py
```diff
@@ -132,16 +141,26 @@ def get_primary_ip(self) -> str:
 
     def get_primary_endpoint(self) -> str:
         """Get the endpoint of the primary node, consisting of address and port."""
-        primary_address = self.get_primary_ip()
         port = TLS_PORT if self.state.unit_server.is_tls_enabled else CLIENT_PORT
 
+        if self.state.substrate == Substrate.K8S:
+            # get the DNS name of the K8s service
+            primary_address = f"{self.state.model.app.name}-{K8sService.PRIMARY.value}"
```
skourta, line 149: Should we use this even for internal calls?

skourta (commented):
Small comments. Thank you @reneradoi

src/managers/sentinel.py
reneradoi: Updated

src/common/k8s_client.py
reneradoi: Yes, in K8s we always start with unit `0`, even [after scale down to zero](https://github.com/canonical/valkey-operator/actions/runs/24755152019/job/72427591817#step:7:7341). This is only relevant for the initial creation of the service though, during lifetime the owner of the services will be the stateful set, which still applies even if there are no more pods (confirmed by testing).

src/common/k8s_client.py
reneradoi: Error logging happens in the event handler where this is called (as documentation for why to defer).

src/events/external_clients.py
reneradoi: Yes I think we should. Deferred events are de-duplicated by ops. If we don't defer, we could miss out important updates and the services could be stale (e.g. not updating the port after TLS switchover).

src/events/external_clients.py
reneradoi: updated

src/managers/sentinel.py
reneradoi: I think we should avoid it, to not an additional layer that could fail.

Mehdi-Bendriss (approved):
Thanks René! approved, merge pending CI pass

skourta (approved):
Thank you @reneradoi

PR #47, opened 2026-04-19
patch: Update timout for failvoer to 190s

skourta:
This pull request adjusts the retry logic for the coordinated primary failover process to better accommodate the expected duration of a failover operation. The main change is to increase the total wait time and interval between retries, ensuring the process aligns with the 3-minute timeout used by Sentinel.

Failover process reliability:

* Increased the number of retry attempts from 5 to 19 and the wait interval between retries from 2 seconds to 10 seconds in the `failover_primary_coordinated` method in `src/common/client.py`, allowing up to 3 minutes and 10 seconds for the failover to complete, matching Sentinel's timeout behavior.

### DIFF
--- a/src/common/client.py
+++ b/src/common/client.py
@@ -372,9 +372,11 @@ class SentinelClient(CliClient):
             == "OK"
         )
 
+    # wait for 3m+10 for the failover to finish
+    # sentinel waits 3m before timing out a failover
     @retry(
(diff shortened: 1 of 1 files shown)


Mehdi-Bendriss (approved):

reneradoi (approved):
We should state this in the docs, otherwise users might complain why it takes so long to remove units.

PR #50, opened 2026-04-22
[DPE-9752] Add topology observer

reneradoi:
The PR adds a topology observer to the charm, inspired by the existing solution in [Postgres](https://github.com/canonical/postgresql-operator/blob/main/scripts/cluster_topology_observer.py). The observer checks if the Primary of Valkey has changed and dispatches a Juju event to the charm if so. The event is used for updating pod labels (in K8s) and client relation data (updated endpoints and read-only-endpoints).

**Design:**
- the topology observer runs as a background process and checks continuously if the primary has changed
- the check happens by querying the `sentinel master` with `valkey-py`, because `valkey-glide` does not support Sentinel connections and both clients do not support pubsub connections with Sentinel
- after startup, the leader unit starts the observer
- on config changes, secret changes, peer relations changes (for TLS updates)  or departure events, the observer gets restarted
- the observer writes a log file to `/var/log/topology_observer.log` and stores the CA certificate for client TLS in the default truststore of the host in `/etc/ssl/certs/`
- on unit removal, the observer gets stopped
- when a primary change is detected, the observer process dispatches a custom event `topology_changed`
- this event is used to update pod labels and client relation data in `ExternalClientEvents`

**Side quest:**
Bug fix for updating passwords of Sentinel admin users. These updates where not configured to the Sentinel ACL files and Sentinel was not restarted to pick up the changes. This bug was found while testing the PR manually.


src/events/base_events.py
```diff
@@ -253,6 +253,14 @@ def _on_unit_fully_started(self, event: UnitFullyStarted) -> None:
             self.charm.unit.open_port("tcp", CLIENT_PORT)
         self.charm.unit.open_port("tcp", TLS_PORT)
 
+        if not self.charm.unit.is_leader():
+            return
+
+        try:
+            self.charm.topology_manager.start_observer()
```
Mehdi-Bendriss, line 257: why isn't this run at the beginning of `_on_unit_fully_started`? in general, why are we conditioning the start of this process on the health of the server?

src/events/base_events.py
```diff
@@ -267,6 +275,15 @@ def _on_peer_relation_changed(self, _: ops.RelationChangedEvent) -> None:
         for lock in [StartLock(self.charm.state), RestartLock(self.charm.state)]:
             lock.process()
 
+        if not self.charm.state.unit_server.is_active:
```
Mehdi-Bendriss, line 275: similar question here, the health of the unit's service should not have an impact on the topology observer process start, or am I missing something?

src/events/base_events.py
```diff
@@ -275,10 +292,27 @@ def _on_peer_relation_departed(self, _: ops.RelationDepartedEvent) -> None:
             logger.error(f"Failed to update sentinel quorum: {e}")
             # not critical to defer here, we can wait for the next relation change
 
+        if not self.charm.unit.is_leader() or not self.charm.state.unit_server.is_active:
+            return
+
+        try:
+            self.charm.topology_manager.restart_observer()
```
Mehdi-Bendriss, line 296: why do we restart in this hook?

src/scripts/topology_observer.py
```diff
@@ -0,0 +1,98 @@
...
+            previous_primary = primary_name
+
+        sentinel = Sentinel(
+            sentinels=addresses,
+            socket_timeout=0.1,
+            sentinel_kwargs=sentinel_kwargs,
+        )
+
+        try:
+            primary_name = sentinel.discover_master(PRIMARY_NAME)[0]
```
Mehdi-Bendriss, line 80: I think we can stick to glide using the Sentinel ports, with something along the lines of: 

```
sentinel_config = GlideClientConfiguration(
    addresses=[NodeAddress(host, port) for host, port in addresses],
    request_timeout=100,  # 0.1s in milliseconds
    ....
)

sentinel_client = await GlideClient.create(sentinel_config)

result = await sentinel_client.custom_command(["SENTINEL", "GET-MASTER-ADDR-BY-NAME", PRIMARY_NAME])
primary_name = result[0].decode()
```

src/literals.py
```diff
@@ -26,6 +26,10 @@
 ACL_FILE = "var/lib/valkey/users.acl"
 SENTINEL_ACL_FILE = "var/lib/valkey/sentinel-users.acl"
 
+# todo: these paths require root access, should be moved to dedicated user directories
+TOPOLOGY_OBSERVER_LOG_FILE = "/var/log/topology_observer.log"
+TOPOLOGY_OBSERVER_TLS_CA_FILE = "/etc/ssl/certs/Valkey_CA.pem"
```
Mehdi-Bendriss: can we have lower cases in the names?

Mehdi-Bendriss (commented):
Thanks René! Good work. 
I left a few comments and questions, mainly around why we start/restart in certain hooks and under certain conditions of the current leader unit + the use of valkey-py / Glide

src/events/base_events.py
reneradoi: Not necessarily on the health of the server, rather on the starting procedure to be completed on the leader unit.

Two things to consider here:
- I think there is no need to start an observer if the unit has not fully started yet, given that this is only started on the leader, which usually starts up first -> the observer would only run into connection errors anyway.
- There is also no need to have the observer in place that early, as there are a lot of `peer-relation-changed` events during the startup process. The services and pod labels will only be created anyway as soon as the leader unit is up ([here](https://github.com/canonical/charmed-etcd-operator/pull/224/changes)). And in case of client relations: The relation data will not be published until the leader unit is up, too.

src/events/base_events.py
reneradoi: No, the health doesn't impact the observer, only the starting procedure to be completed (and the unit not being in the process of being removed).

src/events/base_events.py
reneradoi: When a unit leaves, the connection parameters (endpoints) of the observer should be updated to not connect to this unit anymore.

src/literals.py
reneradoi: updated

src/scripts/topology_observer.py
reneradoi: Unfortunately not: Valkey glide does not support Sentinel connections (see [this issue](https://github.com/valkey-io/valkey-glide/issues/2761)), which I also confirmed by testing. It only supports connections to Valkey instances directly, be it in "standalone" mode (primary + replicas) or "cluster" mode.

It would only be possible to connect to Valkey itself and query the replication info or such. But I don't think we should do this.

Once Valkey glide has support for Sentinel, migration should not be to difficult though. There is actually [work going on on this](https://github.com/valkey-io/valkey-glide/pull/5608), but not yet available for us.

src/events/base_events.py
skourta: For context: The `is_active` flag means the unit started (started flag is set) and is not being scaled down.

src/events/base_events.py
```diff
@@ -294,6 +328,12 @@ def _on_leader_elected(self, event: ops.LeaderElectedEvent) -> None:
         if not self.charm.unit.is_leader():
             return
 
+        if self.charm.state.unit_server.is_active:
+            try:
+                self.charm.topology_manager.start_observer()
```
skourta, line 333: Can you please explain why we call the start on leader_elected.

src/charm.py
```diff
@@ -28,6 +29,8 @@
 class ValkeyCharm(ops.CharmBase):
     """Charmed Operator for Valkey."""
 
+    on = TopologyChangedCharmEvents()
```
skourta, line 32: Curiosity: What if we have more than one customm event from different events? How would it look?

src/events/external_clients.py
```diff
@@ -79,6 +89,7 @@ def __init__(self, charm: "ValkeyCharm"):
         self.framework.observe(
             self.certificate_transfer.on.certificates_removed, self._on_ca_removed
         )
+        self.framework.observe(self.charm.on.topology_changed, self._on_topology_changed)
 
     def _on_bulk_resources_requested(
         self, event: BulkResourcesRequestedEvent[RequirerCommonModel] | ResourceRequestedEvent
```
skourta: Missing generic class type for `ResourceRequestedEvent`

src/managers/sentinel.py
```diff
@@ -389,6 +389,11 @@ def reconcile_k8s_services(self) -> None:
         self.k8s_client.ensure_endpoint_service(role=K8sService.PRIMARY.value, port=valkey_port)
         self.k8s_client.ensure_endpoint_service(role=K8sService.REPLICAS.value, port=valkey_port)
 
+    def set_pod_labels(self) -> None:
+        """Set labels for primary and replica pods in Kubernetes."""
+        if self.state.substrate == Substrate.VM:
```
skourta: I would prefer we do this at the event level

src/managers/topology.py
```diff
@@ -0,0 +1,120 @@
...
+            logger.debug("Topology observer already stopped")
+            return
+
+        logger.debug("Stopping topology observer")
+        try:
+            os.kill(int(observer_pid), signal.SIGTERM)
+            logger.info("Topology observer stopped")
+            self.state.unit_server.update({"topology_observer_pid": ""})
+        except OSError:
+            pass
```
skourta, line 114: We should also clear the pid if it errors no?

skourta: Maybe use a `finally`?

src/managers/topology.py
```diff
@@ -0,0 +1,120 @@
...
+                CharmUsers.SENTINEL_CHARM_ADMIN.value,  # username
+                self.state.cluster.internal_users_credentials.get(
+                    CharmUsers.SENTINEL_CHARM_ADMIN.value, ""
+                ),  # password
+                str(self.state.unit_server.is_tls_enabled),
+                self.state.unit_server.unit_name,
+                self.state.charm.charm_dir,
+            ],
+            # File shouldn't close
+            stdout=open(TOPOLOGY_OBSERVER_LOG_FILE, "a"),  # noqa: SIM115
```
skourta, line 95: Should we never close this?

src/scripts/topology_observer.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def dispatch(unit_name: str, charm_dir: str) -> None:
+    """Dispatch a Juju custom event."""
+    custom_event = "topology_changed"
+
+    juju_run_command = "/usr/bin/juju-exec"
+    dispatch_command = f"JUJU_DISPATCH_PATH=hooks/{custom_event} {charm_dir}/dispatch"
+
+    subprocess.run([juju_run_command, "-u", unit_name, dispatch_command])
```
skourta: add check=True to raise if the command fails

src/events/base_events.py
```diff
@@ -384,15 +430,25 @@ def _on_secret_changed(self, event: ops.SecretChangedEvent) -> None:
...
                     event.defer()
                     return
-            return
+
+                # propagate updated credentials to topology observer
+                try:
+                    self.charm.topology_manager.restart_observer()
+                except (ValkeyWorkloadCommandError, ValueError) as e:
+                    logger.error("Failed to restart topology observer: %s", e)
+                finally:
```
skourta: Do we need the finally? we just return without it

src/events/base_events.py
```diff
@@ -264,6 +272,15 @@ def _on_peer_relation_changed(self, _: ops.RelationChangedEvent) -> None:
         for lock in [StartLock(self.charm.state), RestartLock(self.charm.state)]:
             lock.process()
 
+        if not self.charm.state.unit_server.is_active:
+            return
+
+        # need to pick up scaling operations, TLS switchover, CA rotation and so on
+        try:
+            self.charm.topology_manager.restart_observer()
```
skourta, line 280: Something I did on the glide charm to avoid restarting all the time is to save the config used on disk and compare what we get with what we had. This way we only restart when necessary

src/scripts/topology_observer.py
skourta: Can't we use the cli for this?

src/workload_vm.py
```diff
@@ -93,8 +93,8 @@ def install(self, revision: str | None = None, retry_and_raise: bool = True) ->
             revision = str(SNAP_REVISIONS[platform.machine()])
 
         try:
-            # TODO revesit this logic after snapd update is released
-            # refresh snapd to use candidate to bypass risv check issue.
+            # TODO revisit this logic after snapd update is released
```
skourta, line 96: Thank you

skourta (commented):
No major comments. I would prefer using the CLI over using `valkey-py` since that's what we started the charm with.
I would have also preferred we had a general charm maintenance event dispatched that handeled multiple cases including this one.

src/scripts/topology_observer.py
reneradoi: In general I prefer a proper Python client for this. Only `glide` not being able yet let me to use `valkey-py` instead. It should be easier to change this to `glide` at some point than coming from the cli tool.

src/events/base_events.py
reneradoi: If Juju leadership changes (because the previous leader goes away), the new leader unit needs to start the observer.

src/charm.py
reneradoi: See [this](https://github.com/canonical/postgresql-operator/blob/main/src/charm.py#L187) and [this](https://github.com/canonical/postgresql-operator/blob/main/src/cluster_topology_observer.py#L31-L38) for an example how it's done in Postgres.

src/events/external_clients.py
reneradoi: updated

src/managers/sentinel.py
reneradoi: updated

src/managers/topology.py
reneradoi: updated

src/managers/topology.py
reneradoi: The file is closed when the started process stops:

```
root@juju-eb767c-0:/home/ubuntu# ps -efa | grep observer
root       43898       1  0 11:05 ?        00:00:00 /usr/bin/python3 src/scripts/topology_observer.py 

root@juju-eb767c-0:/home/ubuntu# lsof | grep observer
python3   43898                            root    1w      REG                8,1      2272    2680894 /var/log/topology_observer.log
python3   43898                            root    2w      REG                8,1      2272    2680894 /var/log/topology_observer.log

root@juju-eb767c-0:/home/ubuntu# kill 43898
root@juju-eb767c-0:/home/ubuntu# lsof | grep observer
root@juju-eb767c-0:/home/ubuntu# 
```

src/scripts/topology_observer.py
reneradoi: updated

src/events/base_events.py
reneradoi: updated

src/events/base_events.py
reneradoi: I don't think we need that. The restart is cheap, we can keep it simple here.

src/scripts/topology_observer.py
Mehdi-Bendriss: I would be in favor of the cli as well, I don't like the idea of having another client as a dependency for a single operation

reneradoi: I understand your point. Let me elaborate why I'd prefer to use a Python client: With the current design, the topology observer runs in the operator container. It is started as a new process by the operator code, with the Python env of the current container prepared for it to use. Any Python client is available for the topology observer. To use the workload binaries though (which are only installed on the workload container), the process would have to be started on the workload container, which would also require a Python env to be installed and all required packages to be provided. This would require significant refactoring and the design of this solution would have to deviate from the other products. I don't think that it's worth it, just to be able to use the cli tool instead.

As a compromise, I'd suggest that we keep using `valkey-py` for now until `glide` provides support for connections to Sentinel. Then we can update our code, which should not require a lot of effort or refactoring to make it work.

Mehdi-Bendriss (approved):

skourta (approved):
Thank you @reneradoi

Issue #25, opened 2026-03-17
TLS switchover: Ensure Sentinel restarts only one unit at a time


reneradoi:
For enabling or disabling client TLS, it is possible to reload the TLS configuration in Valkey without restarting. For Sentinel, this is not the case, it has to be restarted to load an updated TLS configuration (and currently also TLS files). To avoid service unavailability due to slow restarts of the Sentinel service, we have to ensure that only one Sentinel restarts at any given time.

For more information, see the discussion in the following link: https://github.com/canonical/valkey-operator/pull/18#discussion_r2934298273

reneradoi:
closed with #51

PR #51, opened 2026-04-27
[DPE-9753] Sentinel rolling restart

reneradoi:
The changes introduced with this PR address https://github.com/canonical/valkey-operator/issues/25 and add rolling restarts for Sentinel.

Implementation changes:
- move custom events to separate file
- move `restart_workload` handler to `charm.py` to be able to call it from multiple event handlers
- replace immediate Sentinel restarts with `restart_workload.emit` to request a restart lock and restart Sentinel asynchronously
- add parameter `primary_endpoint` to the `RestartWorkloadEvent`, which is required to be able to set the Sentinel configuration just ahead of the Sentinel restart to avoid race conditions because of Sentinel's own config-rewrite

Safeguards/adjustments:
- return early in non-TLS related `peer_relation_changed` event handlers if a TLS switchover is in progress (for performance improvements)
- ignore `reconfigure_quorum` during TLS switchover to avoid failures if a Sentinel has not been restarted yet
- on integration tests: as the rolling restarts add a performance penalty, the waiting times and timeouts in some integration tests need to be adjusted


src/events/base_events.py
```diff
@@ -275,6 +230,12 @@ def _on_peer_relation_changed(self, _: ops.RelationChangedEvent) -> None:
         if not self.charm.state.unit_server.is_active:
             return
 
+        if self.charm.state.unit_server.model.tls_client_state in (
```
Mehdi-Bendriss, line 234: can you add a comment to why we check this and return?

src/common/custom_events.py
```diff
@@ -0,0 +1,81 @@
...
+                            the charm code earlier (in the event handler emitting this event)
+                            and the config-rewrite that Sentinel does on its own.
+    """
+
+    def __init__(
+        self,
+        handle: ops.Handle,
+        restart_valkey: bool = True,
+        restart_sentinel: bool = True,
+        primary_endpoint: str = "",
```
Mehdi-Bendriss, line 27: I like this

src/events/base_events.py
```diff
@@ -645,56 +604,19 @@ def _set_state_for_going_away(self) -> None:
...
         if (
             not self.charm.state.unit_server.is_active
             or self.charm.state.unit_server.is_being_removed
             or self.model.app.planned_units() == 0
+            # to avoid failures if a Sentinel has not been restarted yet
+            # does not rely on TLS state because databag might be outdated in deferred events
+            or self.charm.state.client_tls_relation
+            and not self.charm.state.unit_server.is_tls_enabled
+            or self.charm.state.unit_server.is_tls_enabled
+            and not self.charm.state.client_tls_relation
```
Mehdi-Bendriss: can you put parenthesis around the logical blocks (the and operator)? otherwise it becomes a bit of a logical gymnastic exercise to reason with the priorities 😬

src/events/external_clients.py
```diff
@@ -205,6 +191,12 @@ def _on_peer_relation_changed(self, event: ops.RelationChangedEvent) -> None:
         if not self.charm.state.unit_server.is_started:
             return
 
+        if self.charm.state.unit_server.model.tls_client_state in (
```
Mehdi-Bendriss, line 195: similar comment on explaining why we o this check and exit here

tests/integration/tls/test_certificate_rotation.py
```diff
@@ -245,12 +245,8 @@ def test_ca_rotation_by_expiration(juju: jubilant.Juju) -> None:
     The rotation is triggered by the expiration of the CA cert on TLS provider side.
     """
     logger.info("Adjust CA and certificate validity on TLS provider")
-    tls_config = {"certificate-validity": "6m", "root-ca-validity": "12m"}
+    tls_config = {"certificate-validity": "10m", "root-ca-validity": "20m"}
     juju.config(app=TLS_NAME, values=tls_config)
-    juju.wait(
-        lambda status: are_agents_idle(status, APP_NAME, idle_period=30, unit_count=NUM_UNITS),
-        timeout=600,
-    )
```
Mehdi-Bendriss, line 253: does this imply that the app is unable to settle even with 10min?

src/common/custom_events.py
```diff
@@ -0,0 +1,81 @@
...
+                            config immediately before restarting Sentinel. This is required to
+                            avoid race conditions between updating the Sentinel config through
+                            the charm code earlier (in the event handler emitting this event)
+                            and the config-rewrite that Sentinel does on its own.
+    """
+
+    def __init__(
+        self,
+        handle: ops.Handle,
+        restart_valkey: bool = True,
```
Mehdi-Bendriss, line 25: we don't seem to ever emit this event with a `restart_valkey=True`, or do we have 2 ways of restarting valkey ?

Mehdi-Bendriss (commented):
Thanks René! good work! I have some questions and questions

src/common/custom_events.py
reneradoi: We do [on config changed](https://github.com/canonical/valkey-operator/blob/sentinel-rolling-restart/src/events/base_events.py#L354) (`restart_valkey=True` is the default).

Fortunately we don't have to restart Valkey a lot, because we can reload a lot of the config through API calls.

tests/integration/tls/test_certificate_rotation.py
reneradoi: Yes, but the issue is not the application that doesn't settle, rather the certs that are renewed every six minutes. It can happen (especially in MicroK8s deployments) that certs are renewed after the config change successfully, and right after that the certs are already renewed again because of expiration. The idle time in between is not enough to fulfill the wait condition, and the overall execution time of both operations exceeds ten minutes.

I decided to remove the wait instead of reducing the idle time or extending the timeout, because the only check we perform at this point is that we can access Valkey with the current certificate. In subsequent steps, the wait is still present and ensures that the certificate rotation is only considered complete once the application has settled.

I could revert the change, but then would have to increase the timeout to a higher value.

src/events/base_events.py
reneradoi: :) updated

src/events/external_clients.py
reneradoi: updated

src/events/base_events.py
reneradoi: updated

tests/integration/tls/test_certificate_rotation.py
reneradoi: I've removed the change and made the `wait` explicit again.

Mehdi-Bendriss (approved):
Thanks René!

skourta (approved):
Very clean work. Thank you @reneradoi

PR #52, opened 2026-04-30
[DPE-9327] Enable storage reuse

reneradoi:
This PR adds the storage reuse feature.

**Functional changes:**
- handle the `storage_attached` event to fix permissions for already existing volumes (to allow snap installation in case of VM deployments)
- on `storage_detaching`, the operator instead explicitly saves the database to disk in a blocking fashion, to avoid units going away without storing their most recent data

**Test coverage:**
- adds an integration test for removing a unit and adding a new one with previous storage
- adds an integration test for scaling down to zero units, while keeping the application and storage, and then scaling up again with previous storage
- adds an integration test for removing the application entirely (VM-only) and deploying a new application with previous storage
- all integration tests use continuous writes and check that the writes are consistent over all units after reusing storage

### DIFF
--- a/src/events/base_events.py
+++ b/src/events/base_events.py
@@ -95,6 +95,9 @@ class BaseEvents(ops.Object):
         super().__init__(charm, key="base_events")
         self.charm = charm
 
+        self.framework.observe(
+            self.charm.on[DATA_STORAGE].storage_attached, self._on_storage_attached
+        )
         self.framework.observe(self.charm.on.install, self._on_install)
         self.framework.observe(self.charm.on.start, self._on_start)
         self.framework.observe(
@@ -111,6 +114,16 @@ class BaseEvents(ops.Object):
         self.framework.observe(self.restart_workload, self._on_restart_workload)
         self.framework.observe(
             self.charm.on[DATA_STORAGE].storage_detaching, self._on_storage_detaching
+        )
+
+    def _on_storage_attached(self, event: ops.StorageAttachedEvent) -> None:
+        """Handle storage attachment."""
+        if self.charm.state.substrate == Substrate.K8S:
+            return
+
+        # fix the permissions of the directory if re-attaching existing storage
+        self.charm.workload.exec(
+            ["chmod", "-R", "750", self.charm.workload.working_dir.as_posix()]
         )
 
     def _on_install(self, event: ops.InstallEvent) -> None:
--- a/src/events/base_events.py
+++ b/src/events/base_events.py
@@ -611,6 +611,9 @@ class BaseEvents(ops.Object):
         if self.charm.unit.is_leader():
             self.charm.topology_manager.stop_observer()
 
+        logger.info("Save dataset to disk")
+        self.charm.cluster_manager.save_database_blocking()
+
         # stop valkey and sentinel processes
         self.charm.workload.stop()
         active_sentinels = [
--- a/src/events/base_events.py
+++ b/src/events/base_events.py
@@ -569,10 +569,13 @@ class BaseEvents(ops.Object):
             return
 
         active_sentinels = self.charm.sentinel_manager.get_active_sentinel_ips(primary_ip)
-        if (
-            primary_ip == self.charm.state.unit_server.get_endpoint(self.charm.state.substrate)
-            and len(active_sentinels) > 1
-        ):
+        unit_is_primary = (
+            True
+            if primary_ip == self.charm.state.unit_server.get_endpoint(self.charm.state.substrate)
+            else False
+        )
+
+        if unit_is_primary and len(active_sentinels) > 1:
             logger.debug("Triggering sentinel failover on primary IP %s", primary_ip)
             self.charm.sentinel_manager.failover()
             primary_ip = self.charm.sentinel_manager.get_primary_ip()
@@ -583,6 +586,10 @@ class BaseEvents(ops.Object):
 
         if self.charm.unit.is_leader():
             self.charm.topology_manager.stop_observer()
+
+        if not unit_is_primary:
+            logger.info("Waiting for replica to be fully-synced before saving the dataset")
+            self.charm.cluster_manager.wait_for_replica_fully_synced(primary_ip)
 
         logger.info("Save dataset to disk")
         self.charm.cluster_manager.save_database_blocking()
--- a/src/managers/config.py
+++ b/src/managers/config.py
@@ -77,6 +77,10 @@ class ConfigManager(ManagerStatusProtocol):
         config_properties["aclfile"] = self.workload.acl_file.as_posix()
         config_properties["dir"] = self.workload.working_dir.as_posix()
 
+        # avoid potentially inconsistent save overwriting the save triggered by the charm on shutdown
+        config_properties["shutdown-on-sigint"] = "nosave"
+        config_properties["shutdown-on-sigterm"] = "nosave"
+
         config_properties["bind"] = self.state.endpoint
 
         # replica related config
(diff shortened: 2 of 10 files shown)


src/managers/config.py
```diff
@@ -77,6 +77,10 @@ def get_config_properties(self, primary_endpoint: str) -> dict[str, str]:
         config_properties["aclfile"] = self.workload.acl_file.as_posix()
         config_properties["dir"] = self.workload.working_dir.as_posix()
 
+        # avoid potentially inconsistent save overwriting the save triggered by the charm on shutdown
+        config_properties["shutdown-on-sigint"] = "nosave"
+        config_properties["shutdown-on-sigterm"] = "nosave"
```
Mehdi-Bendriss: What about an unmanaged sigint or sigterm? why shouldn't we keep the default to be `save` and let it be on a best effort basis for k8s but still persistent for VM?

src/events/base_events.py
```diff
@@ -571,6 +584,9 @@ def _on_storage_detaching(self, event: ops.StorageDetachingEvent) -> None:
         if self.charm.unit.is_leader():
             self.charm.topology_manager.stop_observer()
```
Mehdi-Bendriss, line 589: we are missing a check here:
```
if replica leaving: (replica == different than the previous primary if it's the one who's leaving)
    # get the current data *offset* of the primary
    # block/wait until the local data catches up to it (>= primary offset at the time we get here) 
```

tests/integration/ha/test_storage_reuse.py
```diff
@@ -0,0 +1,262 @@
...
+    juju.add_unit(
+        APP_NAME,
+        attach_storage=[data_storage_id] if substrate == Substrate.VM else None,
+    )
+    juju.wait(
+        lambda status: are_apps_active_and_agents_idle(
+            status, APP_NAME, unit_count=NUM_UNITS, idle_period=60
+        )
+    )
+
```
Mehdi-Bendriss, line 120: we are missing a check to confirm that only the delta of the data has been synced

tests/integration/ha/test_storage_reuse.py
```diff
@@ -0,0 +1,262 @@
...
+    logger.info("Stopping continuous writes")
+    cw_stats = stop_continuous_writes(juju)
+    addresses = get_cluster_addresses(juju, APP_NAME)
+    assert_continuous_writes_consistent(
+        endpoints=addresses,
+        username=CharmUsers.VALKEY_ADMIN.value,
+        password=get_password(juju, user=CharmUsers.VALKEY_ADMIN),
+        last_written_value=cw_stats.last_written_value,
+    )
```
Mehdi-Bendriss, line 285: we are missing a check where we scale up the cluster with the rest of the units / storages and see that replication works and that only the delta in transferred

Mehdi-Bendriss (commented):
Thanks René! Great work! I have some comments

src/managers/config.py
skourta: Are we being defensive or did you see this while working on the tests?

tests/integration/ha/test_storage_reuse.py
```diff
@@ -0,0 +1,262 @@
...
+        APP_NAME,
+        attach_storage=[data_storage_id] if substrate == Substrate.VM else None,
+    )
+    juju.wait(
+        lambda status: are_apps_active_and_agents_idle(
+            status, APP_NAME, unit_count=NUM_UNITS, idle_period=60
+        )
+    )
+
+    # update hostnames after scale down
```
skourta: old comment

src/events/base_events.py
skourta, line 546: When scaling down to 0 we talked about pausing writes and sync all units before shutting down to ensure all units have the same data

tests/integration/ha/test_storage_reuse.py
```diff
@@ -0,0 +1,262 @@
...
+
+    juju.wait(
+        lambda status: are_apps_active_and_agents_idle(
+            status, APP_NAME, unit_count=NUM_UNITS, idle_period=60
+        )
+    )
+
+    logger.info("Ensure previous data is available again")
+    addresses = get_cluster_addresses(juju, APP_NAME)
+    assert_continuous_writes_consistent(
```
skourta, line 174: I love this. Thank you.

tests/integration/ha/test_storage_reuse.py
skourta: ~~That might not always be the case based on the environement speed and how long the tests take in between steps. If the offset is large enough a full resync is performed even if we had all the old data. We are waiting for 20s before stopping the writes at the start of the test.~~
Sorry you can dismiss the comment we are seeding data before the tests which should take care of this problem

skourta (commented):
Thank you @reneradoi. Very clean implementation. I left a couple of questions mostly.

src/managers/config.py
reneradoi: Hm, I just think that it doesn't make sense to save the database twice. 

If we want to ensure that the data is saved on shutdown, we have to make it explicit. That's why I added the `save` before shutdown. Now, why waiting for another save on the actual shutdown if we have already saved the same data? If it is a large dataset, that will be felt by users when removing a unit on VM (plus: might increase the chance of users running `remove --force --no-wait`, which we don't want). And on K8s, it will probably be cancelled anyway by Pebble.

tests/integration/ha/test_storage_reuse.py
reneradoi: By starting the continuous writes when only one unit is deployed, and checking that they are consistent on all units after scaling up, we do ensure that replication works.

Regarding the delta transfer: How is that check supposed to look like? As soon as we scale up, (partial) synchronization will be started/managed by Valkey itself. I don't even think that we can find a  _reliable_ way from the tests to detect the exact point of synchronization when a unit is added, as we wait until replication is complete on the startup workflow before we consider a unit fully started ([here](https://github.com/canonical/valkey-operator/blob/9/edge/src/events/base_events.py#L192)).

tests/integration/ha/test_storage_reuse.py
reneradoi: Same question: How should we do that? We always wait for a replica to be fully in sync before considering it started. See also comment [here](https://github.com/canonical/valkey-operator/pull/52#discussion_r3180355858).

src/managers/config.py
Mehdi-Bendriss: There are 2 cases:
- managed shutdown, doesn't need the save
- unmanaged shutdown, needs the save

Here we are effectively putting unmanaged shutdowns at risk.

tests/integration/ha/test_storage_reuse.py
reneradoi: removed

src/events/base_events.py
reneradoi: added

src/managers/config.py
reneradoi: updated

tests/integration/ha/test_storage_reuse.py
reneradoi: I have added a check based on the `sync_full` statistics of the primary.

tests/integration/ha/test_storage_reuse.py
reneradoi: I have added a check based on the `sync_full` statistics of the primary.

src/events/base_events.py
reneradoi: Let's postpone this, to get the feature merged in time.

Mehdi-Bendriss (approved):
Thanks René!

skourta (approved):

tests/integration/ha/test_storage_reuse.py
reneradoi: CI run shows that is is not reliable. We can get the number of full syncs, but Valkey may decide to do a full instead of partial sync if it takes too long. I've disabled the check on scaling up after application removal (it is still there on regular scale up).

PR #53, opened 2026-05-05
Fix docs link in metadata.yaml

a-velasco:
The previous link (with `/en/`) leads to a "Page not found" page because the URL schema is configured without languages on Read The Docs.

- [x] Have you updated relevant documentation?


reneradoi (approved):
Thank you Andreia!

skourta (approved):

PR #56, opened 2026-05-13
DA261: apply backup-safety config options

delgod:
Implements DA261 by applying seven backup-safety directives via Python in
`ConfigManager.get_config_properties`, rather than editing the upstream `valkey.conf` template. Two values are derived at render time:

- `min-replicas-to-write` — `1` when `app.planned_units() >= 2`, else `0`,
  so single-unit deployments are not write-frozen.
- `repl-backlog-size` — `256mb` only on hosts with > 1 GiB total RAM;
  smaller hosts fall back to the server default (10mb).

The remaining five (`repl-diskless-load`, `dual-channel-replication-enabled`,
`save`, `maxmemory-policy`, `min-replicas-max-lag`) are static.

The `planned_units` parameter is plumbed through `get_config_properties`,
`set_config_properties`, and `configure_services`, with `self.charm.app.planned_units()`
passed at the four caller sites (`base_events.py` ×2, `tls.py` ×2).

- All overrides live in `ConfigManager` Python, not in the template file.
- No `upgrade_charm` observer added. Existing deployments pick up the new
  values opportunistically on the next event that already triggers
  `configure_services` (VM IP change, TLS event, password rotation, peer
  relation churn). New deployments get the values on first start.
- `repl-backlog-size` is gated on host RAM > 1 GiB to avoid the +246 MiB
  allocation on small hosts.
- `min-replicas-to-write` is dynamic per topology, not the static `1` DA261
  describes.

## Commits

- `refactor(config): plumb planned_units through ConfigManager` — no behavior change
- `feat(config): apply DA261 backup-safety directive overrides` — six directives + dynamic min-replicas-to-write
- `feat(config): gate repl-backlog-size on host RAM > 1 GiB` — RAM-gated seventh directive

### DIFF
--- a/src/events/base_events.py
+++ b/src/events/base_events.py
@@ -219,7 +219,7 @@ class BaseEvents(ops.Object):
         # the rendered config ships min-replicas-to-write=1; reassert the
         # topology-correct runtime value now the server is up, as CONFIG SET
         # does not persist across a restart
-        self._reconcile_min_replicas_to_write()
+        self.charm.cluster_manager.reconcile_min_replicas_to_write()
 
         if self.charm.state.unit_server.tls_client_state != TLSState.TLS:
             self.charm.unit.open_port("tcp", CLIENT_PORT)
@@ -245,7 +245,7 @@ class BaseEvents(ops.Object):
 
         # reassert min-replicas-to-write to match the (possibly changed) topology
         if self.charm.state.unit_server.is_started:
-            self._reconcile_min_replicas_to_write()
+            self.charm.cluster_manager.reconcile_min_replicas_to_write()
 
         if not self.charm.unit.is_leader():
             return
@@ -279,7 +279,7 @@ class BaseEvents(ops.Object):
 
         # a unit departed (scale down): relax min-replicas-to-write if we dropped below 3
         if self.charm.state.unit_server.is_started:
-            self._reconcile_min_replicas_to_write()
+            self.charm.cluster_manager.reconcile_min_replicas_to_write()
 
         if not self.charm.unit.is_leader() or not self.charm.state.unit_server.is_active:
             return
@@ -644,21 +644,6 @@ class BaseEvents(ops.Object):
             )
 
         self.charm.state.unit_server.update({"scale_down_state": ScaleDownState.GOING_AWAY.value})
-
-    def _reconcile_min_replicas_to_write(self) -> None:
-        """Reassert min-replicas-to-write at runtime to match the current topology.
-
-        The rendered config ships 0; this raises it to 1 only when >= 3 units
-        are currently active (and leaves smaller or partially-rolled-out
-        clusters at 0) so a primary is not write-frozen when its only replica
-        is unavailable. Runs on every unit because CONFIG SET is local and does
-        not survive a restart.
-        """
-        try:
-            self.charm.cluster_manager.reconcile_min_replicas_to_write()
-        except (ValkeyConfigSetError, ValkeyWorkloadCommandError) as e:
-            # not critical to defer; reasserted on the next event or restart
-            logger.error("Failed to reconcile min-replicas-to-write: %s", e)
 
     def _reconfigure_quorum_if_necessary(self) -> None:
         """Reconfigure the sentinel quorum if it does not match the current cluster size."""
--- a/src/events/base_events.py
+++ b/src/events/base_events.py
@@ -648,10 +648,10 @@ class BaseEvents(ops.Object):
     def _reconcile_min_replicas_to_write(self) -> None:
         """Reassert min-replicas-to-write at runtime to match the current topology.
 
-        The rendered config ships 0; this raises it to 1 on >= 3-unit
-        deployments (and leaves 1- and 2-unit clusters at 0) so a small
-        cluster's primary is not write-frozen when its only replica is
-        unavailable. Runs on every unit because CONFIG SET is local and does
+        The rendered config ships 0; this raises it to 1 only when >= 3 units
+        are currently active (and leaves smaller or partially-rolled-out
+        clusters at 0) so a primary is not write-frozen when its only replica
+        is unavailable. Runs on every unit because CONFIG SET is local and does
         not survive a restart.
         """
         try:
--- a/src/events/base_events.py
+++ b/src/events/base_events.py
@@ -216,6 +216,11 @@ class BaseEvents(ops.Object):
         self.charm.state.unit_server.update({"start_state": StartState.STARTED.value})
         StartLock(self.charm.state).release_lock()
 
+        # the rendered config ships min-replicas-to-write=1; reassert the
+        # topology-correct runtime value now the server is up, as CONFIG SET
+        # does not persist across a restart
+        self._reconcile_min_replicas_to_write()
+
         if self.charm.state.unit_server.tls_client_state != TLSState.TLS:
             self.charm.unit.open_port("tcp", CLIENT_PORT)
             self.charm.unit.open_port("tcp", SENTINEL_PORT)
@@ -237,6 +242,10 @@ class BaseEvents(ops.Object):
         except ValkeyWorkloadCommandError as e:
             logger.error(f"Failed to update sentinel quorum: {e}")
             # not critical to defer here, we can wait for the next relation change
+
+        # reassert min-replicas-to-write to match the (possibly changed) topology
+        if self.charm.state.unit_server.is_started:
+            self._reconcile_min_replicas_to_write()
 
         if not self.charm.unit.is_leader():
             return
@@ -267,6 +276,10 @@ class BaseEvents(ops.Object):
         except ValkeyWorkloadCommandError as e:
             logger.error(f"Failed to update sentinel quorum: {e}")
             # not critical to defer here, we can wait for the next relation change
+
+        # a unit departed (scale down): relax min-replicas-to-write if we dropped below 3
+        if self.charm.state.unit_server.is_started:
+            self._reconcile_min_replicas_to_write()
 
         if not self.charm.unit.is_leader() or not self.charm.state.unit_server.is_active:
             return
@@ -631,6 +644,21 @@ class BaseEvents(ops.Object):
             )
 
         self.charm.state.unit_server.update({"scale_down_state": ScaleDownState.GOING_AWAY.value})
+
+    def _reconcile_min_replicas_to_write(self) -> None:
+        """Reassert min-replicas-to-write at runtime to match the current topology.
+
+        The rendered config ships 0; this raises it to 1 on >= 3-unit
+        deployments (and leaves 1- and 2-unit clusters at 0) so a small
+        cluster's primary is not write-frozen when its only replica is
+        unavailable. Runs on every unit because CONFIG SET is local and does
+        not survive a restart.
+        """
+        try:
+            self.charm.cluster_manager.reconcile_min_replicas_to_write()
+        except (ValkeyConfigSetError, ValkeyWorkloadCommandError) as e:
+            # not critical to defer; reasserted on the next event or restart
+            logger.error("Failed to reconcile min-replicas-to-write: %s", e)
 
     def _reconfigure_quorum_if_necessary(self) -> None:
         """Reconfigure the sentinel quorum if it does not match the current cluster size."""
--- a/src/managers/cluster.py
+++ b/src/managers/cluster.py
@@ -59,6 +59,25 @@ class ClusterManager(ManagerStatusProtocol):
         if not client.acl_load(hostname=self.state.endpoint):
             raise ValkeyACLLoadError("Could not load ACL file into Valkey cluster.")
 
+    def reconcile_min_replicas_to_write(self) -> None:
+        """Set min-replicas-to-write at runtime to match the current topology.
+
+        Enabled (1) only on clusters that can tolerate losing a replica
+        (>= 3 planned units); disabled (0) on 1- and 2-unit deployments so the
+        primary is not write-frozen when its sole replica is unavailable (e.g.
+        during a rolling restart or scale-down). The rendered config ships 0,
+        so the enabled (1) value must be reasserted after every (re)start
+        because CONFIG SET does not persist.
+        """
+        value = "1" if self.state.charm.app.planned_units() >= 3 else "0"
+        client = self._get_valkey_client()
+        if not client.config_set(
+            hostname=self.state.endpoint,
+            parameter="min-replicas-to-write",
+            value=value,
+        ):
+            raise ValkeyConfigSetError("Could not set min-replicas-to-write on Valkey server.")
+
     def update_primary_auth(self) -> None:
         """Update the primaryauth runtime configuration on the Valkey server."""
         client = self._get_valkey_client()
--- a/src/managers/cluster.py
+++ b/src/managers/cluster.py
@@ -62,14 +62,17 @@ class ClusterManager(ManagerStatusProtocol):
     def reconcile_min_replicas_to_write(self) -> None:
         """Set min-replicas-to-write at runtime to match the current topology.
 
-        Enabled (1) only on clusters that can tolerate losing a replica
-        (>= 3 planned units); disabled (0) on 1- and 2-unit deployments so the
-        primary is not write-frozen when its sole replica is unavailable (e.g.
-        during a rolling restart or scale-down). The rendered config ships 0,
-        so the enabled (1) value must be reasserted after every (re)start
-        because CONFIG SET does not persist.
+        Enabled (1) only on clusters that can currently tolerate losing a
+        replica (>= 3 active units); disabled (0) otherwise so the primary is
+        not write-frozen when a replica is unavailable (e.g. during a rolling
+        restart, a scale-down, or a scale-up that is still rolling out).
+        Counting active units rather than planned units avoids freezing writes
+        when the planned count has already grown but the new units are not up
+        yet. The rendered config ships 0, so the enabled (1) value must be
+        reasserted after every (re)start because CONFIG SET does not persist.
         """
-        value = "1" if self.state.charm.app.planned_units() >= 3 else "0"
+        active_units = len([server for server in self.state.servers if server.is_active])
+        value = "1" if active_units >= 3 else "0"
         client = self._get_valkey_client()
         if not client.config_set(
             hostname=self.state.endpoint,
--- a/src/managers/cluster.py
+++ b/src/managers/cluster.py
@@ -68,18 +68,25 @@ class ClusterManager(ManagerStatusProtocol):
         restart, a scale-down, or a scale-up that is still rolling out).
         Counting active units rather than planned units avoids freezing writes
         when the planned count has already grown but the new units are not up
-        yet. The rendered config ships 0, so the enabled (1) value must be
+        yet. The rendered config ships 1, so the relaxed (0) value must be
         reasserted after every (re)start because CONFIG SET does not persist.
+
+        A failed CONFIG SET is logged and swallowed rather than raised: the
+        value is non-critical and gets reasserted on the next event or restart.
         """
         active_units = len([server for server in self.state.servers if server.is_active])
         value = "1" if active_units >= 3 else "0"
-        client = self._get_valkey_client()
-        if not client.config_set(
-            hostname=self.state.endpoint,
-            parameter="min-replicas-to-write",
-            value=value,
-        ):
-            raise ValkeyConfigSetError("Could not set min-replicas-to-write on Valkey server.")
+        try:
+            client = self._get_valkey_client()
+            if not client.config_set(
+                hostname=self.state.endpoint,
+                parameter="min-replicas-to-write",
+                value=value,
+            ):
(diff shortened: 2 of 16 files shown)


src/managers/config.py
```diff
@@ -297,7 +316,14 @@ def _generate_sentinel_configs(self, primary_endpoint: str) -> dict[str, str]:
...
         sentinel_configs["down-after-milliseconds"] = f"{PRIMARY_NAME} 30000"
-        sentinel_configs["failover-timeout"] = f"{PRIMARY_NAME} 180000"
+        # DA261's dual-channel-replication slows the post-failover resync. With the
+        # default 180s failover-timeout, a SIGKILLed primary that restarts (e.g. after
+        # a long backoff) rejoins as a second master and Sentinel takes too long to
+        # force it back to replica, leaving the sentinels split on the primary. 60s
+        # lets Sentinel reconfigure the rejoined node well within the recovery window
+        # and the cluster converges on a single primary. Validated against the full
+        # HA failover suite with dual-channel enabled.
+        sentinel_configs["failover-timeout"] = f"{PRIMARY_NAME} 60000"
```
skourta, line 323: This should be added to the spec

src/workload_k8s.py
```diff
@@ -157,3 +157,37 @@ def stop(self) -> None:
...
+        except (FileNotFoundError, OSError, pathops.PebbleConnectionError):
+            return self._read_meminfo_total()
+        if raw == "max":
+            return self._read_meminfo_total()
+        try:
+            return int(raw)
+        except ValueError:
+            return 0
+
+    def _read_meminfo_total(self) -> int:
```
skourta: I've tested this on OS and it does not work with juju constraints. The memory we get will be the hosts memory rather than the constraint we set with juju

src/workload_vm.py
```diff
@@ -205,3 +206,17 @@ def stop(self) -> None:
             raise ValkeyServicesCouldNotBeStoppedError(
                 "Valkey services are still alive after stop."
             )
+
+    @override
+    def total_memory_bytes(self) -> int:
```
skourta, line 282: We can also make use of the `/proc/meminfo`

tests/integration/ha/test_failover.py
```diff
@@ -54,6 +54,112 @@
...
+# longer reconf_slaves phase. Every post-disruption check in this module must
+# therefore poll rather than assert once. 24 x 10s = 240s covers the worst case
+# observed locally and in CI.
+_RECOVERY_ATTEMPTS = 24
+_RECOVERY_WAIT = 10
+
+
+def _wait_for_connected_replicas(juju: jubilant.Juju, expected: int, tls_enabled: bool) -> None:
+    """Poll connected_slaves on the primary until it reaches `expected`."""
+    for attempt in Retrying(
```
skourta: I initially included retries for these. However, following a discussion with @Mehdi-Bendriss, it was decided we should not use them to avoid masking any underlying bugs.

skourta (commented):
Thank you @delgod. Minor comments. Already looks to be in a good shape.

src/managers/config.py
```diff
@@ -79,6 +86,16 @@ def get_config_properties(self, primary_endpoint: str) -> dict[str, str]:
 
         config_properties["bind"] = self.state.endpoint
 
+        # DA261 backup-safety overrides
+        config_properties["repl-diskless-load"] = "on-empty-db"
+        config_properties["dual-channel-replication-enabled"] = "yes"
+        config_properties["save"] = "900 1 300 100 60 10000"
+        config_properties["maxmemory-policy"] = "noeviction"
+        config_properties["min-replicas-max-lag"] = "10"
+        config_properties["min-replicas-to-write"] = "1" if planned_units >= 2 else "0"
```
reneradoi: Should we only set it if `planned_units >= 3` to avoid outages in case the replica is unavailable in two-node deployments?

src/managers/config.py
```diff
@@ -46,9 +49,13 @@ def __init__(self, state: ClusterState, workload: WorkloadBase):
         self.state = state
         self.workload = workload
 
-    def get_config_properties(self, primary_endpoint: str) -> dict[str, str]:
+    def get_config_properties(self, primary_endpoint: str, planned_units: int) -> dict[str, str]:
```
reneradoi: I would prefer if we could not pass the `planned_units` as an argument to this method. This cascades signature changes on several other methods and tests, while we could just get the value from `self.state.charm.app.planned_units()` instead.

src/managers/config.py
```diff
@@ -79,6 +86,16 @@ def get_config_properties(self, primary_endpoint: str) -> dict[str, str]:
...
         config_properties["bind"] = self.state.endpoint
 
+        # DA261 backup-safety overrides
+        config_properties["repl-diskless-load"] = "on-empty-db"
+        config_properties["dual-channel-replication-enabled"] = "yes"
+        config_properties["save"] = "900 1 300 100 60 10000"
+        config_properties["maxmemory-policy"] = "noeviction"
+        config_properties["min-replicas-max-lag"] = "10"
+        config_properties["min-replicas-to-write"] = "1" if planned_units >= 2 else "0"
+        if self.workload.total_memory_bytes() > _REPL_BACKLOG_MIN_RAM_BYTES:
```
reneradoi, line 103: I think we can expect 256mb of memory to be available and avoid the additional checks. Valkey is an in-memory database and will probably not work well anyway if less than that.

tests/integration/clients/test_client_relation.py
```diff
@@ -38,6 +40,25 @@
 REQUIRER_TLS_PROVIDER = "ssc-req"
 
 
+def _run_action_with_retry(juju: jubilant.Juju, *args, max_attempts: int = 24, **kwargs):
```
reneradoi: I don't think we should do that. From a client perspective, it might not be acceptable to wait this long for a command to succeed. If we really have an issue here, we should surface that and not hide it in the integration tests just to have green CI.

tests/integration/ha/test_two_unit_availability.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    model_full_name = juju.model
+    admin_password = get_password(juju, CharmUsers.VALKEY_ADMIN)
+    primary_ip = get_primary_ip(juju, APP_NAME)
+    primary_unit = get_unit_name_from_primary_ip(juju, primary_ip, substrate)
+    replica_unit = next(
+        name for name in juju.status().apps[APP_NAME].units if name != primary_unit
+    )
+
+    # 1. Baseline — write succeeds.
```
reneradoi: Can you please modify these comments to be log messages instead? This makes it easier to follow/debug integration test output.

We have established this throughout our integration tests, please see here for an example: https://github.com/canonical/valkey-operator/blob/9/edge/tests/integration/test_charm.py#L88

tests/integration/ha/test_two_unit_availability.py
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/env python3
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""DA261 — 2-unit write-availability trade-off.
```
reneradoi: Follow-up question to [this comment](https://github.com/canonical/valkey-operator/pull/56/changes#r3302379762): Is this really what we want? Please consider that this also means a two-node cluster will be unavailable during any rolling restart. I think we should at least make this configurable for users, otherwise I would rather avoid it. Two node Valkey setups are not that uncommon, from my experience, and I think this would not be a good user experience for these cases.

tests/integration/cw_helpers.py
```diff
@@ -248,3 +247,34 @@ def assert_continuous_writes_consistent(
         len(endpoints),
         len(reference or []),
     )
+
+
+def assert_continuous_writes_consistent(
```
reneradoi: This will effect all existing integration tests - shouldn't we make this change specific for failover tests?

reneradoi (changes requested):
Thanks Mykola! I have a few things that should be adjusted, and a few questions we should discuss, mainly about whether we really want to make two-nodes-cluster this unpleasant for users.

src/managers/config.py
delgod: I would prefer to be safe; it is not a big deal to check the amount of RAM rather than guessing why the customer cannot deploy.

src/workload_k8s.py
delgod: It is not about following Juju constraints; the goal of the change is to avoid a crash on startup. We don't want the Valkey process hunted by the OOM Killer.
So, we follow OOM Killer logic here: if `/sys/fs/cgroup/memory.max` is not set, the process's cgroup doesn't have a limit on memory and can eat up `/proc/meminfo` memory.

src/workload_vm.py
delgod: Should be LXD compatible now.

tests/integration/ha/test_failover.py
delgod: thank you, removed in https://github.com/canonical/valkey-operator/pull/56/changes/c4c03a35e8e142997f151d0a67892b9db2f384f0

src/managers/config.py
delgod: Thank you, fixed in https://github.com/canonical/valkey-operator/pull/56/commits/58c620f6a7c02ee3b63d5135df9613e3a587323a

src/managers/config.py
delgod: Thank you, fixed in https://github.com/canonical/valkey-operator/pull/56/changes/58c620f6a7c02ee3b63d5135df9613e3a587323a

tests/integration/clients/test_client_relation.py
delgod: thank you, removed in https://github.com/canonical/valkey-operator/pull/56/changes/b3998e7b8c92b3083be0fe88fbb44204877fe706

tests/integration/ha/test_two_unit_availability.py
delgod: Thank you, fixed in [3b6d6d0](https://github.com/canonical/valkey-operator/pull/56/commits/3b6d6d008fa9c3554d83dfdc7431b876dd74d8ac)

tests/integration/ha/test_two_unit_availability.py
delgod: Thank you, fixed in https://github.com/canonical/valkey-operator/pull/56/changes/3b6d6d008fa9c3554d83dfdc7431b876dd74d8ac

tests/integration/cw_helpers.py
delgod: Thank you, fixed in https://github.com/canonical/valkey-operator/pull/56/changes/dca63f1512ed71f1363206b92defcc1b360198ad

delgod:
Sorry, I made a lot of mess, so I force-pushed after all

src/workload_k8s.py
skourta: I agree with you. But the value on /proc/meminfo will always be the host's memory on k8s. 
Also this function is duplicated between the workloads and can be moved to base.

src/managers/config.py
```diff
@@ -81,6 +87,20 @@ def get_config_properties(self, primary_endpoint: str) -> dict[str, str]:
 
         config_properties["bind"] = self.state.endpoint
 
+        # DA261 backup-safety overrides
+        config_properties["repl-diskless-load"] = "on-empty-db"
+        config_properties["save"] = "900 1 300 100 60 10000"
+        config_properties["maxmemory-policy"] = "noeviction"
+        config_properties["min-replicas-max-lag"] = "10"
```
skourta, line 94: We are missing explicitely setting `dual-channel-replication-enabled` as the spec describes.

skourta (commented):
Minor changes. PR looks clean now.

src/managers/config.py
delgod: I removed `dual-channel-replication-enabled` value change from both spec and PR, because the config option was just introduced in 8.0 and looks like it made current integration tests unstable without retries; it needs a separate investigation of instabilities on its own.  
spec explicetely say "no change"

skourta: Wasn't it the same as the default value? There should be no change of behavior from Valkey.

delgod: `dual-channel-replication-enabled` is disabled by default and in our charm. Charm behave diferently when I enable it, retries needed in many places.

src/managers/config.py
Mehdi-Bendriss: As discussed in the spec, this should always be set to `1` in valkey.conf. And any change to `0` should happen at runtime through `CONFIG SET min-replicas-to-write 0`.

This is the right logic that should be implemented depending on both scale down / up and (re)start use cases: 
```
_on_peer_relation_departed:  # for scale down
     if planned_units < 3:
          workload.exec("CONFIG SET min-replicas-to-write 0")

_on_peer_relation_changed:
     if planned_units >= 3:  # scale up (post start of the new unit)
          workload.exec("CONFIG SET min-replicas-to-write 1")
     else:  # scale down as a safety precaution
          workload.exec("CONFIG SET min-replicas-to-write 0")

in the valkey `(re)start` flow - as it won't persist post restart
```

src/workload_vm.py
```diff
@@ -277,3 +277,52 @@ def stop(self) -> None:
...
+        MAAS, or cloud VM. Returns 0 if nothing is readable. Cgroup v1 is not
+        supported (Ubuntu noble is cgroup v2 only).
+        """
+        for cgroup_file in ("sys/fs/cgroup/memory.max", "sys/fs/cgroup/memory.high"):
+            limit = self._read_cgroup_limit(cgroup_file)
+            if limit is not None:
+                return limit
+        return self._read_meminfo_total()
+
+    def _read_cgroup_limit(self, relative_path: str) -> int | None:
```
Mehdi-Bendriss: Since the implementation seems to be pretty common, could you put `_read_cgroup_limit` and `_read_meminfo_total` in the base class? 

I would prefer that we have the below code per substrate, and if we have better ways to fetch the memory budget in the future, we will enrich each substrate separately: 
```
class K8sWorkload(WorkloadBase):
    def total_memory_bytes(self) -> int:
        return (
            self._read_cgroup_limit("sys/fs/cgroup/memory.max")
            or self._read_meminfo_total()
        )

class VMWorkload(WorkloadBase):
    def total_memory_bytes(self) -> int:
        return (
            self._read_cgroup_limit("sys/fs/cgroup/memory.max")
            or self._read_cgroup_limit("sys/fs/cgroup/memory.high")
            or self._read_meminfo_total()
        )
```

Mehdi-Bendriss (commented):
Thanks Mykola, I have 2 comments.

tests/spread/k8s/test_two_unit_availability.py/task.yaml
```diff
@@ -0,0 +1,9 @@
+summary: test_two_unit_availability.py
+environment:
+  TEST_MODULE: ha/test_two_unit_availability.py
+systems:
+  - self-hosted-linux-amd64-noble-medium
```
reneradoi: We don't use self-hosted runners anymore, this can be removed.

tests/spread/vm/test_two_unit_availability.py/task.yaml
```diff
@@ -0,0 +1,9 @@
+summary: test_two_unit_availability.py
+environment:
+  TEST_MODULE: ha/test_two_unit_availability.py
+systems:
+  - self-hosted-linux-amd64-noble-medium
```
reneradoi: Same here, no more self-hosted runners.

src/workload_vm.py
reneradoi: I agree with Mehdi's proposal, this is much cleaner and easier to read.

reneradoi (commented):
Minor comments, most issues have been addressed and are good now.

src/managers/config.py
reneradoi: In general we should avoid mixing `CONFIG SET` with the file-based approach, but for this specific case I agree.

src/workload_vm.py
delgod: great idea, fixed in https://github.com/canonical/valkey-operator/pull/56/changes/0407ac434e0c947f0880dc6b7cb9340ba620cb35

src/managers/config.py
delgod: sure, fixed in https://github.com/canonical/valkey-operator/pull/56/changes/a7b1dd67d7ef0bf3393939e450d838ce86368358

tests/spread/k8s/test_two_unit_availability.py/task.yaml
delgod: fixed in https://github.com/canonical/valkey-operator/pull/56/changes/029722f8c741ef0ba1433d179095be7f87ad3d07

tests/spread/vm/test_two_unit_availability.py/task.yaml
delgod: fixed in https://github.com/canonical/valkey-operator/pull/56/changes/029722f8c741ef0ba1433d179095be7f87ad3d07

src/managers/cluster.py
```diff
@@ -59,6 +59,25 @@ def reload_acl_file(self) -> None:
...
+        """Set min-replicas-to-write at runtime to match the current topology.
+
+        Enabled (1) only on clusters that can tolerate losing a replica
+        (>= 3 planned units); disabled (0) on 1- and 2-unit deployments so the
+        primary is not write-frozen when its sole replica is unavailable (e.g.
+        during a rolling restart or scale-down). The rendered config ships 0,
+        so the enabled (1) value must be reasserted after every (re)start
+        because CONFIG SET does not persist.
+        """
+        value = "1" if self.state.charm.app.planned_units() >= 3 else "0"
```
reneradoi: The logic has evolved a bit, but with the way this is set now, I believe we should consider edge cases a bit more. For example, if a two-unit deployment is scaled up to three units while one of the initial two is down, the cluster could suddenly become unavailable during scale-up because with `planned_units` now 3, the primary would need one replica to confirm the write, which will not be possible when one unit is down and the third one is not up yet.

Therefore I believe we should use `len([unit for unit in self.state.servers if unit.is_active])` instead of `planned_units`.

reneradoi (commented):
Only one comment remaining, other than that I'm fine with the changes.

src/managers/cluster.py
delgod: Fixed in https://github.com/canonical/valkey-operator/pull/56/changes/ce2631f75e1aa234c4ec7433316315d5ce7b69d7

reneradoi (dismissed):
Thank you Mykola!

src/managers/config.py
```diff
@@ -92,12 +92,13 @@ def get_config_properties(self, primary_endpoint: str) -> dict[str, str]:
...
-        # rolling restart).
-        planned_units = self.state.charm.app.planned_units()
-        config_properties["min-replicas-to-write"] = "1" if planned_units >= 3 else "0"
+        # min-replicas-to-write gates writes on having an in-sync replica. It
+        # ships as 0 (writes always allowed); ClusterManager.reconcile_min_replicas_to_write
+        # raises it to 1 at runtime only when >= 3 units are active, leaving
+        # smaller or partially-rolled-out clusters at 0 so the primary is not
+        # write-frozen when its sole replica is unavailable (e.g. during a
+        # rolling restart or an in-progress scale-up).
+        config_properties["min-replicas-to-write"] = "0"
```
Mehdi-Bendriss: This should be set to `1`, and the docstrings on `src/events/base_events.py::_reconcile_min_replicas_to_write` and `src/managers/cluster.py::reconcile_min_replicas_to_write` should be corrected to reflect that the default is 1 and not 0.

src/events/base_events.py
```diff
@@ -632,6 +645,21 @@ def _set_state_for_going_away(self) -> None:
...
+        """Reassert min-replicas-to-write at runtime to match the current topology.
+
+        The rendered config ships 0; this raises it to 1 only when >= 3 units
+        are currently active (and leaves smaller or partially-rolled-out
+        clusters at 0) so a primary is not write-frozen when its only replica
+        is unavailable. Runs on every unit because CONFIG SET is local and does
+        not survive a restart.
+        """
+        try:
+            self.charm.cluster_manager.reconcile_min_replicas_to_write()
```
skourta: nit: You can catch the error in the manager function and remove this helper altogether

skourta (dismissed):
Thank you Mykola.

src/managers/config.py
delgod: Fixed in https://github.com/canonical/valkey-operator/pull/56/changes/e06a860084b801a02e7d6a242daf7bdaade39cf4

src/events/base_events.py
delgod: Fixed in https://github.com/canonical/valkey-operator/pull/56/changes/e06a860084b801a02e7d6a242daf7bdaade39cf4

Mehdi-Bendriss (approved):
Thank you!

reneradoi (approved):
Thank you Mykola! Let's please assure the integration tests pass on rerun before merging.

skourta (approved):
Thank you

PR #57, opened 2026-05-13
pyright: wire up CI/CD

delgod:
#### Summary

- Adds a working static tox environment that runs pyright against src/, completing the stub previously referenced by env_list = format, lint, static, unit in tox.ini
(which had no matching [testenv:static] section).
- Declares pyright in a new optional Poetry group static so the type checker is reproducible via the lockfile rather than ambient.
- Adds a static job to .github/workflows/ci.yaml that mirrors the unit-test setup (libprotobuf for valkey-glide) and runs tox run -e static.

#### Why

The repo already had a [tool.pyright] block in pyproject.toml (include = ["src/**/*.py", "lib/**/*.py"]) but nothing actually invoked pyright — no dev dependency, no tox
env, no CI job. Ruff handles lint/format but does not type-check (F rules catch undefined names, not type mismatches). This wires the existing config into the standard
developer and CI flow so type regressions become enforceable.

#### Changes

- pyproject.toml: new [tool.poetry.group.static] (optional) with pyright = "*".
- poetry.lock: regenerated; adds pyright 1.1.409 and its nodeenv transitive dep into the static group; typing-extensions now also belongs to static.
- tox.ini: new [testenv:static] — installs main,charm-libs,unit,static groups and runs poetry run pyright {src_path}. Installing the runtime + unit groups gives pyright
the imports it needs to resolve symbols (ops, charmlibs, valkey, lightkube, pytest, etc.); without them most files would surface as unresolved-import errors.
- .github/workflows/ci.yaml: new static job, ubuntu-latest, 5 min timeout, installs libprotobuf-dev/protobuf-compiler to match unit-test (needed because valkey-glide is
built from source), then tox run -e static.


src/common/k8s_client.py
```diff
@@ -37,7 +37,7 @@ def ensure_endpoint_service(self, role: str, port: int) -> None:
 
         try:
             service = self.client.get(res=Service, name=service_name, namespace=self.namespace)
-            if service.spec.ports != [service_port]:
+            if service.spec and service.spec.ports != [service_port]:
```
Mehdi-Bendriss: I think this additional check will make this method fail silently and let the caller code assume that the endpoint service has been successfully created. 

If this can _realistically_ happen (which i'm not sure it will), we should rather remove this check and catch an `AttributeError` and raise a KubernetesClientError

src/core/cluster_state.py
```diff
@@ -120,7 +121,10 @@ def external_client_relations(self) -> set[ops.Relation]:
     @property
     def bind_address(self) -> str:
         """The network binding address from the peer relation."""
-        if not (binding := self.model.get_binding(self.peer_relation)):
+        if not (peer_relation := self.peer_relation):
```
Mehdi-Bendriss: is this for linting or did you face this during testing?

src/events/base_events.py
```diff
@@ -89,10 +90,14 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
             return
 
-        try:
-            self.charm.workload.install()
-        except RuntimeError:
-            raise RuntimeError("Failed to install the Valkey snap")
+        # install() is snap-specific and only exists on the VM workload; on K8s we
+        # already returned above.
+        workload = self.charm.workload
+        if isinstance(workload, ValkeyVmWorkload):
```
Mehdi-Bendriss: This is not needed as we return earlier if the substrate is k8s

src/events/external_clients.py
```diff
@@ -310,13 +318,16 @@ def _on_ca_relation_created(self, event: ops.RelationCreatedEvent) -> None:
         relation = self.charm.model.get_relation(
             relation_name=CERTIFICATE_TRANSFER_RELATION, relation_id=event.relation.id
         )
+        if relation is None:
+            return
```
Mehdi-Bendriss: This should never happen, it's safe to drop.

src/events/external_clients.py
```diff
@@ -310,13 +318,16 @@ def _on_ca_relation_created(self, event: ops.RelationCreatedEvent) -> None:
...
         relation.data[self.model.app]["version"] = "1"
 
     def _on_ca_available(self, event: CertificatesAvailableEvent) -> None:
         """Handle the CA certificates available event."""
         if not (
             ca_certs := self.certificate_transfer.get_all_certificates(
-                relation_id=event.relation_id
+                # `relation_id` is always an int; the library's snapshot typing widens it.
+                relation_id=cast(int, event.relation_id)
```
Mehdi-Bendriss: This cast is not needed, `event.relation_id` is an `int` and typed as such

src/managers/config.py
```diff
@@ -311,8 +313,13 @@ def set_sentinel_config_properties(self, primary_endpoint: str) -> None:
...
         sentinel_config = self.get_sentinel_config_properties(primary_endpoint=primary_endpoint)
 
+        # the "sentinel" entry is always the nested dict of sentinel directives
+        sentinel_section = sentinel_config["sentinel"]
+        if not isinstance(sentinel_section, dict):
+            sentinel_section = {}
+
         sentinel_config_string = "\n".join(
-            f"sentinel {key} {value}" for key, value in sentinel_config["sentinel"].items()
+            f"sentinel {key} {value}" for key, value in sentinel_section.items()
```
Mehdi-Bendriss: Why do we need this, since it is guaranteed we get a dict? I believe this is overly defensive. 
if this block is non existent i would rather we crash with a KeyError since we don't expect it to be missing and it would indicate a parsing error

src/managers/tls.py
```diff
@@ -71,17 +73,21 @@ def set_ca_rotation_state(self, state: TLSCARotationState) -> None:
...
-            private_key (PrivateKey): The private key.
+            private_key (PrivateKey | None): The private key. When ``None`` (e.g. the key
+                is unchanged), the existing key file on disk is kept as-is.
         """
         self.workload.tls_dir.mkdir(exist_ok=True)
         self.workload.tls_paths.ca_certs_dir.mkdir(exist_ok=True)
 
-        self.workload.write_file(private_key.raw, self.workload.tls_paths.client_key)
+        if private_key is not None:
+            self.workload.write_file(private_key.raw, self.workload.tls_paths.client_key)
```
Mehdi-Bendriss: I read the caller code, and there shouldn't be a case where the private key is None, I think we can revert and be more more assertive on its existence

Mehdi-Bendriss (commented):
Thanks Mykola, I left some comments

src/managers/tls.py
delgod: I think the `None` case can be real. `get_assigned_certificates()` is typed `tuple[list[ProviderCertificate],
  PrivateKey | None]` and the unit tests (`test_client_certificate_available` and friends) explicitly pass `([certificate], None)`. Reverting to a required `PrivateKey` + asserting non-None reintroduces a latent `None.raw` crash and fails those tests, so I would keep the `| None` handling. Happy to revisit if you think the tests are modeling something that can't actually happen.

src/common/k8s_client.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/57/commits/5602d27dd9009eb3460adabc47f7f98d267bdb7f

src/core/cluster_state.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/57/commits/5602d27dd9009eb3460adabc47f7f98d267bdb7f

src/events/base_events.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/57/commits/5602d27dd9009eb3460adabc47f7f98d267bdb7f

src/events/external_clients.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/57/commits/5602d27dd9009eb3460adabc47f7f98d267bdb7f

src/events/external_clients.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/57/commits/5602d27dd9009eb3460adabc47f7f98d267bdb7f

src/managers/config.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/57/commits/5602d27dd9009eb3460adabc47f7f98d267bdb7f

src/managers/tls.py
Mehdi-Bendriss: Sorry i wasn't clear in my previous comment. What i mean by the caller code is: the charm code of `_on_certificate_available` (not the certificates lib). 

From our flow, we make the assumption that the private_key will always be present when we are at that stage, either:
- user provided
- generated by the lib (`_ensure_private_key`)

Because we cannot receive the `certificate_available` hook if the CSR wasn't generated (which requires a private key), so the lib is just being overly defensive as they expose that `get_assigned_certificates` method for other uses.

And if for some reason the private key is missing by the time we get this hook, we should let it crash when we attempt writing it on disk, and not ignore the missing private key.

delgod: Makes sense — you're right that a CSR (and therefore a private key) must already exist for certificate_available to fire, so the | None is just the lib being defensive for its other callers. Reverted write_certificate to require PrivateKey and write unconditionally, so a missing key crashes on write rather than being silently skipped. The call site still receives the lib's PrivateKey | None, so I left a # pyright: ignore[reportArgumentType] there instead of adding a runtime check. Done in 2164c32.

src/common/k8s_client.py
skourta: I think if our objective is to go full in with typing we should do the check in the opposite way.
`if not service.spec` -> raise Exception
Adding a pyright ignore in this case seems like a code smell to me

src/core/cluster_state.py
skourta: Same case. The only reason `None` is not considered is because the places we call the property are code paths that happen after the peer relation has been established. If we want to adopt typing 100% this is a valid check.

src/events/base_events.py
skourta: I actully contemplated adding an empty install function to the K8s workload to fix particular issue

skourta (commented):
I really like the direction and intent of this PR. However, it brings up a larger architectural question: should we enforce 100% typing, or continue with partial adoption? I'm a bit hesitant about merging multiple pyright ignores, as it can make the codebase feel inconsistent and slightly harder to maintain. I think we need to discuss the best path forward for our typing standards.

src/managers/cluster.py
```diff
@@ -156,7 +158,7 @@ def get_version(self) -> str:
         client = self._get_valkey_client()
         server_info = client.info_server(hostname=self.state.endpoint)
 
-        return server_info.get("valkey_version")
+        return server_info["valkey_version"]
```
reneradoi, line 161: The previous code was safe against `KeyError`, after the change this could raise and break the charm hook in case the version is not returned from the client command. I would rather keep it as it was previously.

src/managers/sentinel.py
```diff
@@ -382,13 +385,19 @@ def set_quorum(self, quorum: int) -> None:
 
     def reconcile_k8s_services(self) -> None:
         """Create or update the services in Kubernetes."""
+        if self.k8s_client is None:
+            raise KubernetesClientError("K8s client is only available on the Kubernetes substrate")
+
```
reneradoi: No need for this safeguard, we check on substrate on event handler level.

src/managers/sentinel.py
```diff
@@ -382,13 +385,19 @@ def set_quorum(self, quorum: int) -> None:
...
         valkey_port = TLS_PORT if self.state.unit_server.is_tls_enabled else CLIENT_PORT
 
         self.k8s_client.ensure_endpoint_service(role=K8sService.PRIMARY.value, port=valkey_port)
         self.k8s_client.ensure_endpoint_service(role=K8sService.REPLICAS.value, port=valkey_port)
 
     def set_pod_labels(self) -> None:
         """Set labels for primary and replica pods in Kubernetes."""
+        if self.k8s_client is None:
+            raise KubernetesClientError("K8s client is only available on the Kubernetes substrate")
+
```
reneradoi: Same comment: No need for this safeguard, we check on substrate on event handler level.

src/managers/tls.py
```diff
@@ -110,7 +112,7 @@ def build_sans_ip(self) -> frozenset[str]:
         if self.extra_sans_config_is_valid() and (
             extra_sans_config := self.state.config.get("certificate-extra-sans")
         ):
-            extra_sans = [san.strip() for san in extra_sans_config.split(",")]
+            extra_sans = [san.strip() for san in str(extra_sans_config).split(",")]
```
reneradoi: The config for `extra_sans` is defined as `string` in `config.yaml`, no need to cast.

src/managers/tls.py
```diff
@@ -137,7 +139,7 @@ def build_sans_dns(self) -> frozenset[str]:
...
```
reneradoi: Same comment: The config for `extra_sans` is defined as `string` in `config.yaml`, no need to cast.

reneradoi (commented):
Thank you Mykola for the initivative! I've got only a few minor comments, but looks good otherwise.

src/core/cluster_state.py
Mehdi-Bendriss: @skourta the code paths should be predictable, we need to be able to know at any given time what is the expected state of our system. Being overly defensive masks that and makes reading and reasoning about code hard, and may allow things to fail silently (if defensive too gracefully, like (not) doing an operation conditionally etc)

Mehdi-Bendriss (approved):
Thanks Mykola

src/managers/cluster.py
delgod: After some digging with @Mehdi-Bendriss, we think that if `valkey_version` is not returned, we hide the problem, and it is better to raise `KeyError` here. We also expect that `KeyError` shouldn't happen because `client.info_server` must raise earlier if `valkey_version` is not returned from Valkey.

src/managers/tls.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/57/changes/ed8bd1b539b50559e1c1baf9bedcf4163dc64421

src/managers/tls.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/57/changes/ed8bd1b539b50559e1c1baf9bedcf4163dc64421

src/managers/sentinel.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/57/changes/ed8bd1b539b50559e1c1baf9bedcf4163dc64421

src/managers/sentinel.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/57/changes/ed8bd1b539b50559e1c1baf9bedcf4163dc64421

Mehdi-Bendriss (approved):

skourta (approved):

reneradoi (approved):

PR #58, opened 2026-05-13
DA262: add binary streaming exec primitive (exec_stream)

delgod:
## Summary

Adds a `ProcessHandle` protocol and `exec_stream(command)` method to `WorkloadBase`
for spawning long-running commands whose stdout streams to the caller as raw bytes,
plus VM and K8s implementations. Also extracts `CliClient.build_command_prefix(...)`
from the existing `exec_cli_command` to make the valkey-cli argv construction
reusable.

This is generic platform infrastructure — useful for any feature that needs to
pipe binary output from the workload to a downstream consumer (e.g. uploading
RDB snapshots, streaming logs, exporting backups). The first consumer will be
the S3 backup feature (separate PR).

## Why this is split out

`exec()` buffers stdout, applies a short timeout, and operates in text mode.
None of those are appropriate for streaming binary payloads. Rather than bolt
the new primitive into the backup PR, it lives here as a standalone improvement.

### DIFF
--- a/src/common/client.py
+++ b/src/common/client.py
@@ -39,7 +39,7 @@ class CliClient:
         The password is intentionally NOT placed on the command line --
         ``--pass`` is visible via /proc/<pid>/cmdline to any same-UID (or,
         on K8s, same-pod) process. Callers must supply it through the
-        ``REDISCLI_AUTH`` environment variable instead (see
+        ``VALKEYCLI_AUTH`` environment variable instead (see
         :meth:`exec_cli_command`).
 
         Args:
@@ -95,7 +95,7 @@ class CliClient:
         cli_command = (
             self.build_command_prefix(json_output=json_output, hostname=hostname) + command
         )
-        output, error = self.workload.exec(cli_command, env={"REDISCLI_AUTH": self.password})
+        output, error = self.workload.exec(cli_command, env={"VALKEYCLI_AUTH": self.password})
         output = output.strip()
         if error:
             logger.error(
--- a/src/common/client.py
+++ b/src/common/client.py
@@ -95,9 +95,7 @@ class CliClient:
         cli_command = (
             self.build_command_prefix(json_output=json_output, hostname=hostname) + command
         )
-        output, error = self.workload.exec(
-            cli_command, env={"REDISCLI_AUTH": self.password}
-        )
+        output, error = self.workload.exec(cli_command, env={"REDISCLI_AUTH": self.password})
         output = output.strip()
         if error:
             logger.error(
--- a/src/common/client.py
+++ b/src/common/client.py
@@ -33,6 +33,42 @@ class CliClient:
         self.tls = tls
         self.workload = workload
 
+    def build_command_prefix(self, json_output: bool, hostname: str | None = None) -> list[str]:
+        """Construct the argv prefix for a valkey-cli invocation.
+
+        Args:
+            json_output: If True, appends ``--json``.
+            hostname: Optional host; when provided, inserts ``-h <hostname>``.
+        """
+        cmd: list[str] = [self.workload.cli, "--no-auth-warning"]
+        if hostname is not None:
+            cmd.extend(["-h", hostname])
+        cmd.extend(
+            [
+                "-p",
+                str(self.port),
+                "--user",
+                self.username,
+                "--pass",
+                self.password,
+            ]
+        )
+        if json_output:
+            cmd.append("--json")
+        if self.tls:
+            cmd.extend(
+                [
+                    "--tls",
+                    "--cert",
+                    self.workload.tls_paths.client_cert.as_posix(),
+                    "--key",
+                    self.workload.tls_paths.client_key.as_posix(),
+                    "--cacertdir",
+                    self.workload.tls_paths.ca_certs_dir.as_posix(),
+                ]
+            )
+        return cmd
+
     def exec_cli_command(
         self,
         command: list[str],
@@ -52,30 +88,9 @@ class CliClient:
         Raises:
             ValkeyWorkloadCommandError: If the CLI command fails to execute.
         """
-        port = self.port
-        cli_command: list[str] = [
-            self.workload.cli,
-            "--no-auth-warning",
-            "-h",
-            hostname,
-            "-p",
-            str(port),
-            "--user",
-            self.username,
-            "--pass",
-            self.password,
-        ] + (["--json"] if json_output else [])
-
-        if self.tls:
-            cli_command.append("--tls")
-            cli_command.append("--cert")
-            cli_command.append(self.workload.tls_paths.client_cert.as_posix())
-            cli_command.append("--key")
-            cli_command.append(self.workload.tls_paths.client_key.as_posix())
-            cli_command.append("--cacertdir")
-            cli_command.append(self.workload.tls_paths.ca_certs_dir.as_posix())
-
-        cli_command = cli_command + command
+        cli_command = (
+            self.build_command_prefix(json_output=json_output, hostname=hostname) + command
+        )
         output, error = self.workload.exec(cli_command)
         output = output.strip()
         if error:
--- a/src/common/client.py
+++ b/src/common/client.py
@@ -47,7 +47,7 @@ class CliClient:
             hostname: Optional host; when provided, inserts ``-h <hostname>``.
         """
         cmd: list[str] = [self.workload.cli, "--no-auth-warning"]
-        if hostname is not None:
+        if hostname:
             cmd.extend(["-h", hostname])
         cmd.extend(
             [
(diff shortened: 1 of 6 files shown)


tests/unit/test_backup.py
```diff
@@ -0,0 +1,181 @@
+#!/usr/bin/env python3
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Unit tests for the workload streaming exec primitive."""
+
+
+def test_workload_base_declares_exec_stream():
+    import inspect
```
skourta: Can we import everything needed on the top of the file instead of runtime imports?

tests/unit/test_workload_exec_stream.py
```diff
@@ -0,0 +1,181 @@
+#!/usr/bin/env python3
```
skourta, line 1: The name of the file should be changed

skourta (commented):
These are changes to be used in the backup PR, correct?

src/common/client.py
```diff
@@ -52,31 +92,10 @@ def exec_cli_command(
...
-            cli_command.append(self.workload.tls_paths.client_key.as_posix())
-            cli_command.append("--cacertdir")
-            cli_command.append(self.workload.tls_paths.ca_certs_dir.as_posix())
-
-        cli_command = cli_command + command
-        output, error = self.workload.exec(cli_command)
+        cli_command = (
+            self.build_command_prefix(json_output=json_output, hostname=hostname) + command
+        )
+        output, error = self.workload.exec(cli_command, env={"REDISCLI_AUTH": self.password})
```
reneradoi: 😅 Your agent needs an update. Let's please use `VALKEYCLI_AUTH` instead.

src/common/client.py
```diff
@@ -33,6 +33,46 @@ def __init__(
         self.tls = tls
         self.workload = workload
 
+    def build_command_prefix(self, json_output: bool, hostname: str | None = None) -> list[str]:
+        """Construct the argv prefix for a valkey-cli invocation.
+
+        The password is intentionally NOT placed on the command line --
+        ``--pass`` is visible via /proc/<pid>/cmdline to any same-UID (or,
+        on K8s, same-pod) process. Callers must supply it through the
+        ``REDISCLI_AUTH`` environment variable instead (see
```
reneradoi: `sed s/REDIS/VALKEY/g`

reneradoi (commented):
Hey @delgod this PR implies that you want to download backup files to the operator container? This would introduce several complications:
- the connection could fail while the file is streamed
- this would bypass the `multiple storage volumes` architecture we pursue
- the available storage on the operator container might not be sufficient.

I would rather recommend a more robust architecture, dumping the file directly on the workload container and uploading from there. This would not only address the issues mentioned above, it would also allow for "offline backups" (offline as in Valkey is not started), copying the last saved rdb file on shutdown, whereas the cli command would fail if the server is not started.

Furthermore you could eliminate the need to handle the "streaming" process (and thereby this PR) ourselves and leave this up to Valkey (using the `bgsave` command, which does exactly that).

As for the uploading part: AWS and Azure client libs can handle large files and split them into chunks when uploading, which might also be a more robust approach than streaming directly.

delgod:
> I would rather recommend a more robust architecture, dumping the file directly on the workload container and uploading from there. 

We can discuss this in the spec. In general, then it is enough time to reliably implement a backup system; in most cases, streaming is preferred, as it reduces the backup window (time with reduced performance) and resource consumption. Also, in cases when we have encryption at rest and an S3 backend with Server-Side Encryption, we can avoid storing a plain-text dump. e.g., you can check that most backup tools for PostgreSQL, MySQL, and MongoDB are streaming.

src/common/client.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/58/commits/eeef57d527e5fa6438134f91e1d8e417bbde321a

src/common/client.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/58/commits/eeef57d527e5fa6438134f91e1d8e417bbde321a

tests/unit/test_backup.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/58/commits/eeef57d527e5fa6438134f91e1d8e417bbde321a

tests/unit/test_workload_exec_stream.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/58/commits/eeef57d527e5fa6438134f91e1d8e417bbde321a

src/common/client.py
```diff
@@ -33,6 +33,46 @@ def __init__(
...
+        on K8s, same-pod) process. Callers must supply it through the
+        ``VALKEYCLI_AUTH`` environment variable instead (see
+        :meth:`exec_cli_command`).
+
+        Args:
+            json_output: If True, appends ``--json``.
+            hostname: Optional host; when provided, inserts ``-h <hostname>``.
+        """
+        cmd: list[str] = [self.workload.cli, "--no-auth-warning"]
+        if hostname is not None:
```
Mehdi-Bendriss: nit
```suggestion
        if hostname:
```

src/workload_k8s.py
```diff
@@ -29,6 +32,62 @@
...
+    """
+
+    def __init__(self, process: ops.pebble.ExecProcess):
+        self._process = process
+        self.stdout = process.stdout
+        self._stderr_buf: collections.deque[bytes] = collections.deque(maxlen=64)
+        self._stderr_thread = threading.Thread(target=self._drain_stderr, daemon=True)
+        self._stderr_thread.start()
+
+    def _drain_stderr(self) -> None:
```
Mehdi-Bendriss, line 57: nit, could you add docstrings to this method and its VM counterpart?

src/workload_vm.py
```diff
@@ -134,14 +184,17 @@ def restart(self, service: str) -> None:
...
+        self, command: list[str], env: dict[str, str] | None = None
+    ) -> tuple[str, str | None]:
         try:
             output = subprocess.run(
                 command,
                 check=True,
                 text=True,
                 capture_output=True,
                 timeout=10,
+                env={**os.environ, **env} if env else None,
```
Mehdi-Bendriss:
```suggestion
                env={**os.environ, **env} if env else os.environ,
```

Mehdi-Bendriss (commented):
Thank you Mykola, the implementation is sound, only 1 comment on the environment being passed to exec and a couple of nits.
 
Just for a mental visualisation of the caller code, the streaming should be as follows:
```
process_handle = self.workload.exec_stream(command)
try:
    with open(target_path, "wb") as target:
        while chunk := process_handle.stdout.read(1024 * 1024):
            target.write(chunk)

    rc, stderr = process_handle.wait()
    if rc != 0:
        raise ...
except Exception:
    handle.kill()
    raise 
```

src/workload_vm.py
delgod: fixed https://github.com/canonical/valkey-operator/pull/58/changes/01e9736bf9228f7d13350a60822b8a6b37ed50e4

src/workload_k8s.py
delgod: fixed https://github.com/canonical/valkey-operator/pull/58/changes/01e9736bf9228f7d13350a60822b8a6b37ed50e4

src/common/client.py
delgod: fixed https://github.com/canonical/valkey-operator/pull/58/changes/01e9736bf9228f7d13350a60822b8a6b37ed50e4

Mehdi-Bendriss (dismissed):

src/workload_vm.py
```diff
@@ -151,6 +210,18 @@ def exec(self, command: list[str]) -> tuple[str, str | None]:
...
 
+    @override
+    def exec_stream(self, command: list[str], env: dict[str, str] | None = None) -> ProcessHandle:
+        return _VmProcessHandle(
+            subprocess.Popen(
+                command,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                bufsize=0,
+                env={**os.environ, **env} if env else None,
```
reneradoi: I think we should make it `if env else os.environ` here too, or any specific reason why not?

reneradoi (dismissed):
Just a minor nit on my side, otherwise good.

src/workload_vm.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/58/commits/71413ac6152bc464ea8d60ed002fbdd12b98a2e2

skourta (approved):
Re-approval. Thank you @delgod

Mehdi-Bendriss (approved):

PR #59, opened 2026-05-13
DA262: S3 backup (create-backup + list-backups)

delgod:
## Summary

Adds two Juju actions — `create-backup` and `list-backups` — that stream a fresh Valkey RDB snapshot from the targeted unit's `valkey-cli --rdb -` stdout directly to S3 via boto3 multipart upload. **Any unit** (leader or follower) can run the actions; locking is per-unit.

Credentials are supplied by relating the charm to the upstream [`s3-integrator`](https://charmhub.io/s3-integrator) via a new `s3-credentials` relation (interface `s3`, limit 1, optional).

## Depends on

https://github.com/canonical/valkey-operator/pull/58 (`workload-exec-stream`). Must be merged first — this PR uses `workload.exec_stream` and `CliClient.build_command_prefix`.

## What's in the PR
```text
src/
literals.py                # S3_RELATION_NAME, BACKUP_ID_FORMAT, BACKUP_CA_FILENAME
statuses.py                # BackupStatuses enum (3 entries)
common/exceptions.py       # ValkeyBackupError
core/base_workload.py      # tls_paths.backup_ca property
core/models.py             # s3_credentials on PeerAppModel,
                           # backup_id on PeerUnitModel,
                           # ValkeyServer.is_backup_in_progress,
                           # ValkeyCluster.s3_credentials (parsed JSON envelope)
core/cluster_state.py      # s3_relation property
managers/backup.py         # NEW – BackupManager + status protocol
events/backup.py           # NEW – BackupEvents + _exists_preventing_reason
events/base_events.py      # storage_detaching guard during backup
charm.py                   # wire BackupManager + BackupEvents;
                           # register with StatusHandler;
                           # restart_workload guard during backup
metadata.yaml                # requires.s3-credentials
actions.yaml                 # create-backup, list-backups
pyproject.toml               # boto3, mypy-boto3-s3
lib/charms/data_platform_libs/v0/s3.py   # via charmcraft fetch-lib
tests/unit/test_backup.py    # ~40 ops-scenario unit tests
```

## Design highlights

- **Streaming, no archive storage.** `bucket.upload_fileobj(proc.stdout, key, Config=TransferConfig(multipart_chunksize=8 MiB))` reads chunks from the pipe; no on-disk staging.
- **Per-unit lock.** The `backup_id` field in the running unit's peer-unit  databag is both the lock value and the operation identifier. Two backups  can run concurrently on *different* units (each gets a distinct S3 key);  two on the *same* unit are rejected.
- **Cleanup on failure.** Partial S3 objects are deleted (`bucket.Object(key).delete()`) when `valkey-cli` exits non-zero or `upload_fileobj` raises.
- **boto3 quirks honoured.** `Config(request_checksum_calculation="when_required", response_checksum_validation="when_required")` (boto3#4400);
  `CreateBucketConfiguration` omitted for `us-east-1` (aws-sdk-js#3647);
  idempotent error tokens (`BucketAlreadyOwnedByYou`, `BucketAlreadyExists`, `BucketNameUnavailable`).
- **TLS CA chain stored on every unit** (not only the leader) so any unit can  use TLS to talk to a self-signed S3 endpoint (e.g. RadosGW, MicroCeph).
- **Leader-switchover recovery.** Re-observes `leader_elected` to re-trigger  `_on_s3_credentials_changed` — covers the case where `credentials_gone`  fired only on the old leader.

## Action UX

```bash
# Any unit — leader or follower
juju run valkey/N create-backup
# → { "backup-id": "2026-05-13T10:00:00Z" }

juju run valkey/N list-backups
# → { "backups": "backup-id … | backup-status\n-…-\n2026-05-13T10:00:00Z | finished" }
```

Status surface:
- `BACKUP_IN_PROGRESS` (maintenance, unit scope) while streaming
- `BACKUP_S3_PARAMETERS_MISSING` (blocked, app scope) when the relation is present but the integrator hasn't supplied bucket/credentials yet
- `BACKUP_FAILED` (blocked, unit scope) after a failure


src/core/models.py
```diff
@@ -224,6 +231,24 @@ def __init__(
         self.app = component
         self.data_interface = data_interface
 
+    @property
+    def s3_credentials(self) -> dict[str, object] | None:
```
Mehdi-Bendriss: I think we can stick to a return type of `dict[str, str]` (explanation below)

src/core/models.py
```diff
@@ -224,6 +231,24 @@ def __init__(
...
+        if not self.model:
+            return None
+        raw = self.model.s3_credentials
+        if not raw:
+            return None
+        try:
+            return json.loads(raw)
+        except (TypeError, json.JSONDecodeError) as e:
+            logger.warning("Failed to parse s3_credentials JSON: %s", e)
+            return None
```
Mehdi-Bendriss: we can simplify this, while keeping the same guarantees on the caller side (i.e: `if s3_credentials`) with the below:

```suggestion
        if not self.model:
            return {}
        return json.loads(self.model.s3_credentials if self.model.s3_credentials else "{}")
```

That should also reduce the code when we introduce both azure and gcs (service-account based)

src/events/backup.py
```diff
@@ -0,0 +1,188 @@
...
+        return None
+
+    # ── event handlers ──────────────────────────────────────────────────
+
+    def _on_s3_credentials_changed(
+        self, event: CredentialsChangedEvent | ops.LeaderElectedEvent
+    ) -> None:
+        """Handle initial and updated S3 integrator credentials."""
+        s3 = self.s3_requirer.get_s3_connection_info()
+        if not s3:
```
Mehdi-Bendriss: 1. nit: walrus, and maybe `s3_info` instead of `s3`?
2. let's add a log line

src/events/backup.py
```diff
@@ -0,0 +1,188 @@
...
+        s3["endpoint"] = s3.get("endpoint", "").rstrip("/")
+        s3["path"] = s3.get("path", "").strip("/")
+        s3["bucket"] = s3.get("bucket", "").strip("/")
+
+        # Validate AFTER normalisation: path="/" or bucket="/" collapse to ""
+        # here, and an empty path makes list_backups enumerate the whole
+        # bucket (cross-tenant leak in a shared bucket).
+        required = ["bucket", "endpoint", "path", "access-key", "secret-key"]
+        missing = [p for p in required if not s3.get(p)]
+        if missing:
```
Mehdi-Bendriss: nit, walrus and more expressive variable names
```suggestion
        required_params = ["bucket", "endpoint", "path", "access-key", "secret-key"]
        if missing_params := [p for p in required if not s3.get(p)]:
```

src/events/backup.py
```diff
@@ -0,0 +1,188 @@
...
+        self.charm.state.cluster.update({"s3_credentials": json.dumps(s3)})
+
+    def _on_s3_credentials_gone(self, event: CredentialsGoneEvent) -> None:
+        """Handle removal of the S3 credentials relation."""
+        if self.charm.state.unit_server.is_backup_in_progress:
+            logger.warning("Backup in progress; deferring credentials_gone")
+            event.defer()
+            return
+        self.charm.backup_manager.remove_tls_ca_chain()
+        if self.charm.unit.is_leader():
```
Mehdi-Bendriss, line 126: nit, a bit of breathing space between distinct blocks

src/events/backup.py
```diff
@@ -0,0 +1,188 @@
...
+        )
+        self.framework.observe(self.s3_requirer.on.credentials_gone, self._on_s3_credentials_gone)
+        # Recover credentials when leadership moves.
+        self.framework.observe(self.charm.on.leader_elected, self._on_s3_credentials_changed)
+        self.framework.observe(self.charm.on.create_backup_action, self._on_create_backup_action)
+        self.framework.observe(self.charm.on.list_backups_action, self._on_list_backups_action)
+
+    # ── guard ────────────────────────────────────────────────────────────
+
+    def _blocking_reason(self) -> str | None:
```
Mehdi-Bendriss: This method should be used by all 3 actions: `create-backup, restore, list-backups`, so:

1. can we put this method after the hooks?
2. can we add an argument along the lines of: `check_running_operations: bool = True` (we can use it to check whether we check that backup/restore are in progress)

src/events/backup.py
```diff
@@ -0,0 +1,188 @@
...
+        self.charm.backup_manager.remove_tls_ca_chain()
+        if self.charm.unit.is_leader():
+            self.charm.state.cluster.update({"s3_credentials": ""})
+
+    def _on_create_backup_action(self, event: ops.ActionEvent) -> None:
+        """Run a streaming RDB backup of this unit's Valkey instance to S3."""
+        reason = self._blocking_reason()
+        if reason is None and self.charm.state.unit_server.is_backup_in_progress:
+            reason = "A backup is already in progress on this unit."
+        if reason:
```
Mehdi-Bendriss: 1. Let's put this condition in `self._blocking_reason()`
2. walrus operator

src/events/backup.py
```diff
@@ -0,0 +1,188 @@
...
+            event.set_results({"error": reason})
+            event.fail(reason)
+            return
+        # Audit the invocation itself, not just the manager-level transfer
+        # (P1-24): ties a specific Juju action run to the resulting backup
+        # for forensics on a leaked or unexpected RDB.
```
Mehdi-Bendriss: what do you mean by "leaked"?

src/events/backup.py
```diff
@@ -0,0 +1,188 @@
...
+        except ValkeyBackupError as e:
+            logger.exception("Backup failed")
+            event.set_results({"error": _safe_error(e)})
+            event.fail("Backup failed. Check juju debug-log for details.")
+            return
+        event.set_results({"backup-id": backup_id})
+
+    def _on_list_backups_action(self, event: ops.ActionEvent) -> None:
+        """List backups currently in S3, newest first."""
+        if reason := self._blocking_reason():
```
Mehdi-Bendriss: related to a previous comment, we could use it with the argument `check_running_operations = False`

src/events/base_events.py
```diff
@@ -530,6 +530,12 @@ def _update_internal_users_password(self, secret_id: str) -> None:
 
     def _on_storage_detaching(self, event: ops.StorageDetachingEvent) -> None:
         """Handle removal of the data storage mount, e.g. when removing a unit."""
+        if self.charm.state.unit_server.is_backup_in_progress:
+            logger.warning(
+                "Backup in progress on this unit; refusing to scale down until it finishes."
+            )
+            return
```
Mehdi-Bendriss: we shouldn't return, otherwise the teardown flow will proceed. We should either raise an Exception or an infinite loop `while backup_in_progress: sleep(1)`

And I think this may need to land in the `scale_down_lock` logic below

src/managers/backup.py
```diff
@@ -0,0 +1,336 @@
...
+    # pyright being strict about mutable-variable invariance, and the same
+    # narrowing is used by the other managers.
+    state: "ClusterState"
+
+    def __init__(self, state: "ClusterState", workload: "WorkloadBase"):
+        self.state = state  # pyright: ignore[reportIncompatibleVariableOverride]
+        self.workload = workload
+
+    @property
+    def _ca_chain_path(self) -> pathlib.Path:
```
Mehdi-Bendriss: this should be defined in the Workload classes

src/managers/backup.py
```diff
@@ -0,0 +1,336 @@
...
+    def create_backup(self) -> str:
+        """Stream a fresh RDB from the local Valkey instance to S3.
+
+        Sets a per-unit lock on the running unit's databag, streams
+        ``valkey-cli --rdb -`` stdout into ``bucket.upload_fileobj``,
+        and cleans up the S3 object on failure.
+        """
+        s3_parameters = self.state.cluster.s3_credentials
+        if s3_parameters is None:
+            raise ValkeyBackupError("S3 credentials are not configured")
```
Mehdi-Bendriss: No need for this check, we already have the `blocking_reason` check that performs it upfront

src/managers/backup.py
```diff
@@ -0,0 +1,336 @@
...
+        backup_id = datetime.now(timezone.utc).strftime(BACKUP_ID_FORMAT)
+        key = f"{s3_parameters['path']}/{backup_id}"
+        started = datetime.now(timezone.utc)
```
Mehdi-Bendriss: nit, for consistency to not have diverging values. Can we:

```suggestion

        started = datetime.now(timezone.utc)
        backup_id = started.strftime(BACKUP_ID_FORMAT)
        key = f"{s3_parameters['path']}/{backup_id}"
        
```

src/managers/backup.py
```diff
@@ -0,0 +1,336 @@
...
+        bucket = self._get_bucket_resource(s3_parameters)
+        # Pass the admin password via VALKEYCLI_AUTH, never on argv (P1-2).
+        proc = self.workload.exec_stream(
+            self._build_rdb_command(),
+            env={"VALKEYCLI_AUTH": self.state.unit_server.valkey_admin_password},
+        )
+        reader = _CountingReader(proc.stdout)
+
+        try:
+            try:
```
Mehdi-Bendriss: let's unnest the `try`s and put multiple except sequentially
```
try:
except ValkeyBackupError:
except <full list of expected exceptions> as e:
     proc.kill()
     raise Valkey...
else:
finally:
```

src/managers/backup.py
```diff
@@ -0,0 +1,336 @@
...
+                # No outer retry: proc.stdout is a non-rewindable pipe, so a
+                # retry would upload only the bytes left after the first
+                # attempt. boto3 retries individual multipart parts itself.
```
Mehdi-Bendriss: I find this phrasing complex, maybe:
```
# Do not retry the whole upload: reader is backed by proc.stdout and cannot be rewound.
```

src/managers/backup.py
```diff
@@ -0,0 +1,336 @@
...
+                bucket.upload_fileobj(
+                    # _CountingReader duck-types the read()-only interface
+                    # upload_fileobj actually uses; the stub wants a full IO.
```
Mehdi-Bendriss: we can remove

src/managers/backup.py
```diff
@@ -0,0 +1,336 @@
...
+                    Config=TransferConfig(multipart_chunksize=8 * 1024 * 1024),
+                )
+            except ClientError as e:
+                proc.kill()
+                self._delete_object_best_effort(bucket, key)
+                raise ValkeyBackupError(e) from e
+
+            rc, stderr = proc.wait()
+            if rc != 0:
+                self._delete_object_best_effort(bucket, key)
```
Mehdi-Bendriss: 1. Isn't the s3 client taking care automatically of removals if a multipart upload fails?
2. this can be done in an `except ValkeyBackupError` block

src/managers/backup.py
```diff
@@ -0,0 +1,336 @@
...
+                raise ValkeyBackupError(f"valkey-cli --rdb exited {rc}: {stderr}")
+
+            # valkey-cli can exit 0 having written nothing (or an error blob)
```
Mehdi-Bendriss, line 274: is this verified?

src/managers/backup.py
```diff
@@ -0,0 +1,336 @@
...
+            logger.info(
+                "backup.completed backup_id=%s bytes=%d elapsed_seconds=%.1f",
+                backup_id,
+                reader.bytes_read,
+                elapsed,
+            )
+        finally:
+            # Never let a databag write failure here mask the real error
+            # (or, on success, crash the action). A leaked lock is recovered
+            # via the troubleshooting runbook / the lock TTL.
```
Mehdi-Bendriss: we can remove

src/managers/backup.py
```diff
@@ -0,0 +1,336 @@
...
+            try:
+                self.state.unit_server.update({"backup_id": ""})
+            except Exception:
```
Mehdi-Bendriss: this will never raise an exception, we can remove

src/managers/backup.py
```diff
@@ -0,0 +1,336 @@
...
+            username=CharmUsers.VALKEY_ADMIN.value,
+            password=self.state.unit_server.valkey_admin_password,
+            tls=self.state.unit_server.is_tls_enabled,
+            workload=self.workload,
+        )
+        prefix = client.build_command_prefix(json_output=False, hostname=self.state.endpoint)
+        return prefix + ["--rdb", "-"]
+
+    @staticmethod
+    def _delete_object_best_effort(bucket: "Bucket", key: str) -> None:
```
Mehdi-Bendriss, line 322: missing docstrings

src/managers/backup.py
```diff
@@ -0,0 +1,336 @@
...
+        try:
+            bucket.Object(key).delete()
+        except ClientError as e:
```
Mehdi-Bendriss: I think the list is bigger than that, maybe makes sense to use `Exception` here
```suggestion
        except Exception as e:
```

src/events/backup.py
```diff
@@ -0,0 +1,188 @@
...
+            event.fail(reason)
+            return
+        # Audit the invocation itself, not just the manager-level transfer
+        # (P1-24): ties a specific Juju action run to the resulting backup
+        # for forensics on a leaked or unexpected RDB.
+        logger.info(
+            "audit: create-backup action invoked action_id=%s unit=%s",
+            event.id,
+            self.charm.unit.name,
+        )
```
Mehdi-Bendriss, line 142: I think we are missing a running (async?) status set here

actions.yaml
```diff
@@ -8,3 +8,14 @@ status-detail:
...
+create-backup:
+  description: |
+    Stream a fresh RDB backup of this unit's Valkey instance to S3.
+    Runs on the targeted unit (any unit, leader or follower).
+    S3 credentials are taken from the related s3-integrator charm.
+
+list-backups:
+  description: |
+    List all backup objects in the configured S3 bucket, newest first.
+    Safe to run while another backup is uploading.
```
Mehdi-Bendriss, line 21: We are missing a parameter for the format (json / table) - [ref here](https://github.com/canonical/opensearch-operator/blob/b55ac3966f145379e0aaa521432655b92df047cb/actions.yaml#L48C3-L54C90)
(I just realized we're also missing that in etcd...)

src/events/backup.py
```diff
@@ -0,0 +1,188 @@
...
+            event.fail("Backup failed. Check juju debug-log for details.")
+            return
+        event.set_results({"backup-id": backup_id})
+
+    def _on_list_backups_action(self, event: ops.ActionEvent) -> None:
+        """List backups currently in S3, newest first."""
+        if reason := self._blocking_reason():
+            event.set_results({"error": reason})
+            event.fail(reason)
+            return
```
Mehdi-Bendriss, line 177: We are missing a format parameter here
```
if (output_format := event.params.get("output", "").lower()) not in {"json", "table"}:
    event.fail("Failed: invalid output format, must be either 'json' or 'table'.")
    return
```

src/events/backup.py
```diff
@@ -0,0 +1,188 @@
...
+            self.charm.unit.name,
+        )
+        try:
+            ids = self.charm.backup_manager.list_backups()
+        except ValkeyBackupError as e:
+            logger.exception("List backups failed")
+            event.set_results({"error": _safe_error(e)})
+            event.fail("List backups failed. Check juju debug-log for details.")
+            return
+        event.set_results({"backups": self.charm.backup_manager.format_backup_list(ids)})
```
Mehdi-Bendriss: This needs to be conditional on the format, if `json` VS `table`

Mehdi-Bendriss (commented):
Thank you Mykola! Good work! I left some comments

src/events/backup.py
delgod: Fixed in https://github.com/canonical/valkey-operator/pull/59/changes/453101a87c7d217f06dfabd40312563f893bd6f5

src/events/backup.py
delgod: Fixed in https://github.com/canonical/valkey-operator/pull/59/changes/453101a87c7d217f06dfabd40312563f893bd6f5

actions.yaml
delgod: Fixed in https://github.com/canonical/valkey-operator/pull/59/changes/453101a87c7d217f06dfabd40312563f893bd6f5

src/events/backup.py
delgod: Fixed in https://github.com/canonical/valkey-operator/pull/59/changes/453101a87c7d217f06dfabd40312563f893bd6f5

src/events/backup.py
delgod: Fixed in https://github.com/canonical/valkey-operator/pull/59/changes/453101a87c7d217f06dfabd40312563f893bd6f5

src/core/models.py
delgod: Fixed in https://github.com/canonical/valkey-operator/pull/59/changes/453101a87c7d217f06dfabd40312563f893bd6f5

src/events/backup.py
delgod: Fixed in https://github.com/canonical/valkey-operator/pull/59/changes/453101a87c7d217f06dfabd40312563f893bd6f5

src/events/backup.py
delgod: Fixed in https://github.com/canonical/valkey-operator/pull/59/changes/453101a87c7d217f06dfabd40312563f893bd6f5

src/events/backup.py
delgod: Fixed in https://github.com/canonical/valkey-operator/pull/59/changes/453101a87c7d217f06dfabd40312563f893bd6f5

src/events/backup.py
delgod: Added — create-backup sets BACKUP_IN_PROGRESS (is_action=True) before streaming and clears it in a finally, so it shows in juju status during  the action. Fixed in 453101a.

src/events/backup.py
delgod: list-backups now calls _blocking_reason(check_running_operations=False). Fixed in 453101a.

src/events/base_events.py
delgod: Agreed, return doesn't hold teardown — now raises  `ValkeyBackupInProgressError` so Juju retries `storage-detaching` until the  backup finishes. Gated at the top of the handler rather than in  `scale_down_lock`. Fixed in 453101a.

src/managers/backup.py
delgod: Removed (here and in list_backups); the property returns {} now, so the  guard was dead. Fixed in 453101a.

src/managers/backup.py
delgod: Done — started computed once, backup_id derived from it (also removes a possible cross-second skew). Fixed in 453101a.

src/managers/backup.py
delgod: Done — single try with sequential except ValkeyBackupError / except ClientError / else / finally. Fixed in 453101a.

src/managers/backup.py
delgod: Used your phrasing. Fixed in 453101a.

src/managers/backup.py
delgod: Removed. Fixed in 453101a.

src/managers/backup.py
delgod: Removed. Fixed in 453101a.

src/managers/backup.py
delgod: Removed the try/except — peer relation is present in this path, so the write won't raise. Fixed in 453101a.

src/managers/backup.py
delgod: Done — except Exception so best-effort cleanup can't mask the original error. Fixed in 453101a.

src/managers/backup.py
delgod: Added a docstring to _delete_object_best_effort. Fixed in 453101a.

src/core/models.py
delgod: Not quite — the envelope carries tls-ca-chain, which the s3 lib gives us as a list[str], so the values aren't uniformly strings and dict[str, str] would be inaccurate (boto3's verify=/CA handling reads that list). Went with dict[str, Any] and adopted your simplification below (returns {} instead of None). Fixed in 453101a.

src/events/backup.py
delgod: "Leaked" meant an RDB snapshot ending up somewhere it shouldn't — wrong bucket/tenant, or copied out of the cluster. The audit log line exists so a specific Juju action-run can be tied back to a backup-id if that ever needs investigating. I dropped the jargon and reworded it to "if an RDB later turns up somewhere unexpected." Fixed in 453101a.

src/managers/backup.py
delgod: Correct on both. boto3's managed upload_fileobj aborts the multipart upload itself on a mid-stream failure (and a sub-threshold PutObject leaves no object at all), so the delete on the ClientError path was a no-op — removed it; that branch now only kills the producer. The explicit _delete_object_best_effort is kept solely in the except ValkeyBackupError block, for the case the object did fully upload but is invalid (non-zero exit code or bad RDB magic). Fixed in 453101a.

src/managers/backup.py
delgod: It's a defensive guard rather than a reproduced failure: valkey-cli --rdb - normally exits non-zero on error, but this stops a 0-byte or error-blob write from being recorded as a valid backup if it ever exits 0 with junk on stdout. I'd keep it as cheap insurance, but happy to drop it if you'd rather rely on the exit code alone.

Mehdi-Bendriss (dismissed):
Thank you Mykola!

src/common/exceptions.py
```diff
@@ -70,3 +70,15 @@ class ValkeyCertificatesNotReadyError(Exception):
 
 class KubernetesClientError(Exception):
     """Custom Exception if a connection to the Kubernetes Cluster API fails."""
+
+
+class ValkeyBackupError(Exception):
+    """Raised when a backup operation fails."""
+
+
+class ValkeyBackupInProgressError(Exception):
```
skourta: Should be based on ValkeyBackupError

src/core/base_workload.py
```diff
@@ -99,6 +102,19 @@ class WorkloadBase(ABC):
...
+        """Path to the S3 endpoint CA bundle used by the backup client.
+
+        Deliberately a charm-process-local path, NOT a ``tls_paths`` entry:
+        boto3 runs in the charm process, not the workload container, so on
+        K8s the two do not share a filesystem and the bundle could not live
+        in the workload's (container) TLS dir. Keeping it out of that dir
+        also stops the S3 endpoint CA being trusted as a Valkey client CA.
+        ``JUJU_CHARM_DIR`` is the charm dir Juju exports for every hook.
+        """
+        return pathlib.Path(os.environ["JUJU_CHARM_DIR"]) / BACKUP_CA_FILENAME
```
skourta: We can do this the ops way [`charm.charm_dir`](https://canonical.com/juju/docs/ops/latest/reference/ops/#ops.CharmBase.charm_dir)

concierge-k8s.yaml
```diff
@@ -15,3 +15,6 @@ host:
       channel: latest/edge
       connections:
         - jhack:dot-local-share-juju snapd
+    # S3-compatible backend for the backup integration tests.
+    microceph:
```
skourta, line 19: This will install microceph for all k8s tests right?

skourta: Likewise on VM

src/managers/backup.py
```diff
@@ -0,0 +1,316 @@
...
+    def store_tls_ca_chain(self, s3_parameters: dict[str, Any]) -> None:
+        """Write the S3 endpoint CA chain to the charm-local path for boto3."""
+        chain = s3_parameters.get("tls-ca-chain")
+        if not chain:
+            return
+        # A misconfigured integrator may send a bare string; "\n".join would
+        # then iterate characters and write a corrupt CA bundle. Require a
+        # list of PEM strings.
+        if not isinstance(chain, list) or not all(isinstance(c, str) for c in chain):
+            logger.warning("tls-ca-chain is malformed (not a list of strings); ignoring")
```
skourta: We should also check if the list items are PEM strings like [here](https://github.com/canonical/valkey-operator/blob/9/edge/src/managers/tls.py#L181-L185).

src/events/backup.py
```diff
@@ -0,0 +1,215 @@
...
+        """Handle initial and updated S3 integrator credentials."""
+        if not (s3_info := self.s3_requirer.get_s3_connection_info()):
+            return
+        logger.info("S3 credentials changed; refreshing backup configuration")
+
+        # CA chain must be on disk for every unit so any unit can use TLS to S3.
+        self.charm.backup_manager.store_tls_ca_chain(s3_info)
+
+        if not self.charm.unit.is_leader():
+            return
```
skourta, line 86: This creates a race. The non leader units will need to wait for the bucket creation before doing any further operations. We can do this check in `_blocking_reason`

src/events/backup.py
```diff
@@ -0,0 +1,215 @@
...
+        if not self.charm.state.peer_relation:
+            event.defer()
+            return
+
+        for k, v in list(s3_info.items()):
+            if isinstance(v, str):
+                s3_info[k] = v.strip()
+        s3_info["endpoint"] = s3_info.get("endpoint", "").rstrip("/")
```
skourta: Can you please explain why we do this (lines 88-90)?

src/events/backup.py
```diff
@@ -0,0 +1,215 @@
...
+        # bucket (cross-tenant leak in a shared bucket).
+        required_params = ["bucket", "endpoint", "path", "access-key", "secret-key"]
+        if missing_params := [p for p in required_params if not s3_info.get(p)]:
+            logger.warning(
+                "S3 integrator parameters missing or empty after normalisation: %s",
+                missing_params,
+            )
+            return
+
+        # leader_elected re-fires this handler; skip the create_bucket round
```
skourta, line 101: How so?

src/managers/backup.py
```diff
@@ -0,0 +1,316 @@
...
+
+    from core.base_workload import WorkloadBase
+    from core.cluster_state import ClusterState
+
+logger = logging.getLogger(__name__)
+
+# RDB streams begin with a magic header: "REDIS" on upstream Redis, "VALKEY"
+# on Valkey. An upload that does not start with one of these is not a valid
+# snapshot and must not be recorded as a backup.
+_RDB_MAGIC = (b"REDIS", b"VALKEY")
```
skourta, line 41: We will always have `VALKEY` though no?

src/managers/backup.py
```diff
@@ -0,0 +1,316 @@
...
+    # narrowing is used by the other managers.
+    state: "ClusterState"
+
+    def __init__(self, state: "ClusterState", workload: "WorkloadBase"):
+        self.state = state  # pyright: ignore[reportIncompatibleVariableOverride]
+        self.workload = workload
+
+    # ── boto3 client construction ────────────────────────────────────────
+
+    def _get_bucket_resource(self, s3_parameters: dict[str, object]):
```
skourta: nit: Maybe we should parse the s3_parameter dict into a pydantic or data class and avoid the casting

src/managers/backup.py
```diff
@@ -0,0 +1,316 @@
...
+        return s3.Bucket(cast("str", s3_parameters["bucket"]))
+
+    # ── bucket lifecycle ────────────────────────────────────────────────
+
+    def create_bucket(self, s3_parameters: dict[str, Any]) -> None:
+        """Create the configured bucket; idempotent across S3 implementations."""
+        bucket = self._get_bucket_resource(s3_parameters)
+        region = s3_parameters.get("region")
+        try:
+            if region and region != "us-east-1":
```
skourta, line 143: why is `us-east-1` special?

src/managers/backup.py
```diff
@@ -0,0 +1,316 @@
...
+            logger.info(
+                "backup.completed backup_id=%s bytes=%d elapsed_seconds=%.1f",
+                backup_id,
+                reader.bytes_read,
+                elapsed,
+            )
+        finally:
+            self.state.unit_server.update({"backup_id": ""})
+
+        return backup_id
```
skourta, line 306: Is it possible to do any kind of verification on what we just wrote?

skourta (commented):
Thank you @delgod. I left some comments/questions.

concierge-k8s.yaml
delgod: Yes, for this test https://github.com/canonical/valkey-operator/pull/60 it is not ready for review yet

src/managers/backup.py
reneradoi: This can be helpful when migrating from Redis instances through a backup/restore, so good to keep.

src/managers/backup.py
reneradoi: It does not work with a location constraint, see here for reference: https://github.com/aws/aws-sdk-js/issues/3647

We also have this switch in others: https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/src/managers/backup.py#L101-L102

lib/charms/data_platform_libs/v0/s3.py
```diff
@@ -0,0 +1,792 @@
+# Copyright 2023 Canonical Ltd.
```
reneradoi: I think we should try to use the `object-storage-charmlib` instead. This would/should provide us with a common interface and s3/azure/gcs and we could use it as a Python lib instead of soon-to-be-deprecated charmlibs.

src/managers/backup.py
skourta: This only gets executed from out charm on creating backups. Our charm only uses valkey. @delgod raised a valid point of other people potentially deploying our charm with another image like that of redis.

concierge-k8s.yaml
skourta: Talked about this with @delgod. If it does not impact the CI's stability we can leave it as is. It is more elegant than having to install microceph the opensearch charm way with fictures.

src/common/exceptions.py
delgod: Fixed in 332c9c7

src/events/backup.py
skourta: Discussed it with @delgod and it is ok.

src/core/base_workload.py
delgod: Thank you, fixed in b4605ee.

skourta (commented):
LGTM. The only remaining concern is if the backup procedure (streaming) takes a long time the create backup action will time out.

src/managers/backup.py
delgod: Fixed in 7b3425d — store_tls_ca_chain now rejects the chain unless every item carries a PEM armour header (-----BEGIN ...-----), using the same regex shape as managers/tls.py. Added a unit test for the non-PEM-items case.

src/events/backup.py
delgod: This follows the established data-platform pattern — see charmed-etcd-operator `src/events/backup.py::_on_s3_credentials_changed`, which does exactly this: store the TLS chain on all units, then on the leader only validate/normalise the envelope, call `create_bucket(...)`, and publish creds to the app databag; its action guard (`_exists_preventing_reason`) checks relation/creds/health and never creates the bucket.

 The wait is intrinsic to the leader-publishes-creds model regardless of where the bucket is created: a follower can't back up until the leader has validated credentials and written `s3_credentials` to the app databag — which `_blocking_reason` already surfaces as "S3 credentials unavailable."

src/managers/backup.py
```diff
@@ -0,0 +1,341 @@
...
+        """
+        return self.state.charm.charm_dir / BACKUP_CA_FILENAME
+
+    # ── boto3 client construction ────────────────────────────────────────
+
+    def _get_bucket_resource(self, s3_parameters: dict[str, object]):
+        """Build a boto3 Bucket resource configured per the s3-integrator envelope."""
+        verify: bool | str = True
+        if s3_parameters.get("tls-ca-chain"):
+            verify = str(self._backup_ca_path)
```
reneradoi: `PathProtocol` objects don't need to be casted, we can use `.as_posix()` instead.

src/managers/backup.py
delgod: Confirmed on a live unit — Valkey 9.0.1 writes magic VALKEY080:
```
$ head -c 16 /var/lib/valkey/dump.rdb | xxd
00000000: 5641 4c4b 4559 3038 30fa 0a76 616c 6b65  VALKEY080..valke
```
So yes, in normal operation it's always VALKEY. I kept REDIS in the tuple as cheap cross-version defense: Valkey 7.2/8.0 (forked from Redis 7.2) wrote REDIS magic

src/managers/backup.py
```diff
@@ -0,0 +1,341 @@
...
+        boto3 runs in the charm process, not the workload container, so on
+        K8s the two do not share a filesystem and the bundle could not live
+        in the workload's (container) TLS dir. Keeping it out of that dir
+        also stops the S3 endpoint CA being trusted as a Valkey client CA.
+        """
+        return self.state.charm.charm_dir / BACKUP_CA_FILENAME
+
+    # ── boto3 client construction ────────────────────────────────────────
+
+    def _get_bucket_resource(self, s3_parameters: dict[str, object]):
```
reneradoi: nit: Missing return value

src/managers/backup.py
```diff
@@ -0,0 +1,341 @@
...
+        """Build a boto3 Bucket resource configured per the s3-integrator envelope."""
+        verify: bool | str = True
+        if s3_parameters.get("tls-ca-chain"):
+            verify = str(self._backup_ca_path)
+
+        # Scope the credentials to a Session so they are not free-floating
+        # kwargs that show up in repr(args) of any boto3 traceback.
+        session = boto3.Session(
```
reneradoi, line 113: Why are we using a session and not the newer, cleaner resource instead?
(see for comparison: https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/src/managers/backup.py#L56)

reneradoi (commented):
Thank you Mykola, and sorry for the late review. I agree with the overall concept and have only some minor comments and questions.

reneradoi:
Hint for the failing integration tests: https://github.com/canonical/valkey-operator/pull/76

src/events/backup.py
delgod: Documented in b2450b8 — the block trims whitespace then strips separators so they don't corrupt S3 key paths (trailing / on endpoint doubles in the URL; leading/trailing `/` on path/bucket yields `//<id>`, and stripping collapses `path="/"` to `""` so the validation below rejects it).

src/managers/backup.py
delgod: Yes — `_CountingReader` wraps the stream boto3 uploads, capturing the first 16 bytes and the byte count. After upload `create_backup` rejects the object unless `bytes_read > 0` and `head.startswith(_RDB_MAGIC)` (backup.py:264), deleting it otherwise. That validates the exact bytes persisted to S3. A separate `HEAD`/`GetObject` read-back would only re-confirm existence/size (S3 already integrity-checks each multipart part). Happy to add a `bucket.Object(key).load()` size assert if needed

src/managers/backup.py
delgod: Intentional security reason: boto3.Session(...) scopes the credentials so they don't ride along as free-floating kwargs in repr(args) of any boto3 traceback. etcd passes them straight to boto3.resource(...). I'd keep the Session for the credential-hygiene reason, but it's a wash functionally — say the word if you'd rather match etcd exactly.
e.g., practical case: avoid s3 credentials leak when backup got an exception in GitHub CI/CD

reneradoi: Thank you for the explanation. We can keep it, I was mostly asking for understanding.

lib/charms/data_platform_libs/v0/s3.py
delgod: Fixed in https://github.com/canonical/valkey-operator/pull/59/changes/81e38634764dc0e26805b749979183fdfb910640

src/managers/backup.py
delgod: fixed in https://github.com/canonical/valkey-operator/pull/59/changes/ce02e4dc786c170f510bcfd6550f54f4247d9808

src/managers/backup.py
delgod: Fixed in 13ac2ad

src/managers/backup.py
delgod: Fixed in 13ac2ad

Mehdi-Bendriss (approved):

reneradoi (commented):
Hey Mykola! I've tested the PR, it looks really good UX wise and the implementation is also robust against failure scenarios. I am happy to approve, there is only one issue that we need to either solve in this PR or in the next one, and it is about safeguarding the backup creation on units in parallel: I was able to produce a situation where two units create a backup at the same time, which results in the exact same name for the backup file to be uploaded to s3 storage. Both units stream their backup (according to the logs), and there is no error. But we end up with only one backup in the s3 storage (see screenshot below, sorry it is a bit chaotic :sweat: ). It is not clear which unit won the race, whether the file is corrupted or not, and there is no error from the s3 client either. This might be an edge case, but it could happen for example with scheduled backups. I think we need to safeguard it. The most simple solution would be to add the unit name to the backup file, but that's just an idea.

Image

reneradoi (approved):
The issue mentioned above will be treated in a separate PR and will be tracked with https://github.com/canonical/valkey-operator/issues/77.

skourta (approved):

PR #60, opened 2026-05-13
DA261 S3 backup end-to-end against MicroCeph

delgod:
## Summary

Adds an end-to-end integration test suite for the S3 backup feature, backed by a MicroCeph-based S3 fixture. Exercises the **any-unit** guarantee by running `create-backup` on three different units and `list-backups` on a fourth, then verifying the resulting objects in the bucket.

## Depends on

https://github.com/canonical/valkey-operator/pull/59 must be merged first — this PR exercises actions and the relation surface introduced there.

## What's in the PR

- `tests/integration/backup/__init__.py` — package marker
- `tests/integration/backup/conftest.py`:
  - Module-scoped `microceph` fixture: `snap install microceph`, bootstrap,    `disk add loop,4G,3 --wipe`, enable rgw with `--ssl-port=445 --ssl-certificate=auto`,    create a `test` user, generate a unique bucket name, expose the rgw    self-signed cert
  - `s3_bucket` fixture: boto3 `Bucket` resource (creates the bucket eagerly,    `verify=False` for the self-signed endpoint)
  - `s3_secret_content`: access/secret-key dict for the Juju secret used by    `s3-integrator`
- `tests/integration/backup/test_s3_backup.py`:
  - Deploys 3 valkey units + s3-integrator on Ubuntu 24.04 via `jubilant`
  - Provisions an `s3-creds` Juju secret, configures the integrator (bucket, endpoint, region, path, path-style URIs), waits for `active/idle`
  - Integrates valkey ↔ s3-integrator
  - **Runs `create-backup` on valkey/0** — asserts a `backup-id` matching `%Y-%m-%dT%H:%M:%SZ`
  - **Runs `create-backup` on valkey/1** — asserts a distinct id
  - **Runs `list-backups` on valkey/2** — asserts both ids appear in the table in reverse-chronological order
  - Verifies via boto3 that two objects exist under the configured prefix and the first object's body starts with the RDB magic bytes (`REDIS` or `VALKEY`)

## Notes for reviewers

- The MicroCeph fixture is adapted from `canonical/charmed-etcd-operator`'s `tests/integration/backup/conftest.py`
  (3.6/edge). The single test deviation from etcd-operator: a  `--rdb` magic-byte verification of the uploaded object replaces etcd's  "restore on a new model" round-trip, since restore is not yet implemented.

### DIFF
--- a/tests/integration/backup/conftest.py
+++ b/tests/integration/backup/conftest.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python3
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Fixtures for S3 backup integration tests, backed by MicroCeph."""
+
+from __future__ import annotations
+
+import json
+import secrets
+import subprocess
+from pathlib import Path
+
+import boto3
+import pytest
+from botocore.config import Config
+
+
+def _run(*cmd: str, **kwargs) -> str:
+    return subprocess.check_output(cmd, text=True, **kwargs).strip()
+
+
+@pytest.fixture(scope="module")
+def microceph() -> dict:
+    """Install MicroCeph snap, bootstrap, enable rgw with TLS, create a user + bucket."""
+    _run("sudo", "snap", "install", "microceph", "--channel=squid/stable")
+    _run("sudo", "microceph", "cluster", "bootstrap")
+    _run("sudo", "microceph", "disk", "add", "loop,4G,3", "--wipe")
+    _run("sudo", "microceph", "enable", "rgw", "--ssl-port=445", "--ssl-certificate=auto")
+
+    user_json = _run(
+        "sudo",
+        "radosgw-admin",
+        "user",
+        "create",
+        "--uid=test",
+        "--display-name=test",
+    )
+    user = json.loads(user_json)
+    access_key = user["keys"][0]["access_key"]
+    secret_key = user["keys"][0]["secret_key"]
+
+    bucket_name = f"valkey-backup-{secrets.token_hex(4)}"
+    cert = Path("/var/snap/microceph/current/conf/ssl-cert.pem").read_text()
+
+    return {
+        "endpoint": "https://127.0.0.1:445",
+        "access-key": access_key,
+        "secret-key": secret_key,
+        "bucket": bucket_name,
+        "region": "default",
+        "path": "valkey",
+        "tls-ca-chain": [cert],
+    }
+
+
+@pytest.fixture(scope="module")
+def s3_bucket(microceph):
+    """Return a boto3 Bucket resource for the test bucket, creating it eagerly."""
+    s3 = boto3.resource(
+        "s3",
+        region_name=microceph["region"],
+        endpoint_url=microceph["endpoint"],
+        aws_access_key_id=microceph["access-key"],
+        aws_secret_access_key=microceph["secret-key"],
+        config=Config(
+            request_checksum_calculation="when_required",
+            response_checksum_validation="when_required",
+        ),
+        verify=False,
+    )
+    bucket = s3.Bucket(microceph["bucket"])
+    bucket.create()
+    bucket.wait_until_exists()
+    return bucket
+
+
+@pytest.fixture(scope="module")
+def s3_secret_content(microceph) -> dict:
+    return {
+        "access-key": microceph["access-key"],
+        "secret-key": microceph["secret-key"],
+    }
--- a/tests/integration/backup/conftest.py
+++ b/tests/integration/backup/conftest.py
@@ -2,52 +2,129 @@
 # Copyright 2026 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-"""Fixtures for S3 backup integration tests, backed by MicroCeph."""
+"""Fixtures for S3 backup integration tests, backed by MicroCeph.
+
+MicroCeph's RGW is fronted with a self-signed TLS certificate generated here,
+so the suite exercises the charm's full S3-over-TLS path (CA-chain
+distribution + boto3 verification), not just plaintext S3. The certificate's
+SAN covers the host's routable IP because the charm units (k8s pods / lxd
+machines) reach the gateway over that address -- never loopback, which from a
+unit resolves to the unit itself.
+
+Every MicroCeph step is idempotent so the suite can be re-run locally without
+tearing the cluster down between runs.
+"""
 
 from __future__ import annotations
 
+import base64
 import json
 import secrets
+import socket
 import subprocess
+import time
 from pathlib import Path
 
 import boto3
 import pytest
 from botocore.config import Config
 
+RGW_SSL_PORT = 445
+# Self-signed cert + key, persisted so repeated local runs reuse the exact
+# material the already-running gateway serves -- regenerating without
+# re-enabling RGW would break TLS verification against the old certificate.
+_CERT_DIR = Path.home() / ".cache" / "valkey-itest"
+_CERT = _CERT_DIR / "rgw-cert.pem"
+_KEY = _CERT_DIR / "rgw-key.pem"
+
 
 def _run(*cmd: str, **kwargs) -> str:
     return subprocess.check_output(cmd, text=True, **kwargs).strip()
 
 
+def _ok(*cmd: str) -> bool:
+    """Return True if the command exits 0 (idempotent check-then-act helper)."""
+    return subprocess.run(cmd, capture_output=True).returncode == 0
+
+
+def _host_ip() -> str:
+    """First routable IPv4 of this host, reachable from charm units."""
+    return _run("hostname", "-I").split()[0]
+
+
+def _port_open(host: str, port: int) -> bool:
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+        sock.settimeout(1)
+        return sock.connect_ex((host, port)) == 0
+
+
+def _ensure_microceph() -> None:
+    """Install + bootstrap MicroCeph with OSDs; each step idempotent."""
+    if not _ok("snap", "list", "microceph"):
+        _run("sudo", "snap", "install", "microceph", "--channel=squid/stable")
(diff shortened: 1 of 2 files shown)


tests/integration/backup/conftest.py
```diff
@@ -0,0 +1,152 @@
...
+    return _run("hostname", "-I").split()[0]
+
+
+def _port_open(host: str, port: int) -> bool:
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+        sock.settimeout(1)
+        return sock.connect_ex((host, port)) == 0
+
+
+def _ensure_microceph() -> None:
```
reneradoi, line 59: Idempotence here is nice!

tests/integration/backup/conftest.py
```diff
@@ -0,0 +1,152 @@
...
+_KEY = _CERT_DIR / "rgw-key.pem"
+
+
+def _run(*cmd: str, **kwargs) -> str:
+    return subprocess.check_output(cmd, text=True, **kwargs).strip()
+
+
+def _ok(*cmd: str) -> bool:
+    """Return True if the command exits 0 (idempotent check-then-act helper)."""
+    return subprocess.run(cmd, capture_output=True).returncode == 0
```
reneradoi: We can use `subprocess.run(..., check_output=True) instead of comparing the return code. This will raise an `CalledProcessError` if case of failure and also provide the error in `stderr`.

tests/integration/backup/conftest.py
```diff
@@ -0,0 +1,152 @@
...
+# Self-signed cert + key, persisted so repeated local runs reuse the exact
+# material the already-running gateway serves -- regenerating without
+# re-enabling RGW would break TLS verification against the old certificate.
+_CERT_DIR = Path.home() / ".cache" / "valkey-itest"
+_CERT = _CERT_DIR / "rgw-cert.pem"
+_KEY = _CERT_DIR / "rgw-key.pem"
+
+
+def _run(*cmd: str, **kwargs) -> str:
+    return subprocess.check_output(cmd, text=True, **kwargs).strip()
```
reneradoi: Why not use `subprocess.run` for a function called `run`?

tests/integration/backup/conftest.py
```diff
@@ -0,0 +1,152 @@
...
+            break
+        time.sleep(1)
+    return _CERT.read_text()
+
+
+def _ensure_user(uid: str = "test") -> tuple[str, str]:
+    """Get-or-create an RGW user; return (access_key, secret_key)."""
+    if _ok("sudo", "radosgw-admin", "user", "info", f"--uid={uid}"):
+        out = _run("sudo", "radosgw-admin", "user", "info", f"--uid={uid}")
+    else:
```
reneradoi: With the proposed refactoring of `_run`, we could use 

```
try:
    out = _run(user info)
except CalledProcessError:
    out = _run(user create)
```
here to have a cleaner solution.

tests/integration/backup/test_s3_backup.py
```diff
@@ -0,0 +1,111 @@
...
+import time
+
+import jubilant
+
+APP_NAME = "valkey"
+S3_INTEGRATOR_APP = "s3-integrator"
+BACKUP_ID_RE = re.compile(r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z$")
+
+
+def _wait_active(juju: jubilant.Juju, *apps: str, timeout: int = 600) -> None:
```
reneradoi: Please make use of the `are_apps_active_and_agents_idle` helper in `tests/integration/helpers.py`. We use this all over our integration tests to have a standarized `wait` which always provides the same guarantees.

tests/integration/backup/test_s3_backup.py
```diff
@@ -0,0 +1,111 @@
...
+                for app in apps
+            )
+            and jubilant.all_agents_idle(status, *apps)
+        ),
+        timeout=timeout,
+    )
+
+
+def test_backup_and_list(juju: jubilant.Juju, microceph: dict, s3_bucket) -> None:
+    juju.deploy(APP_NAME, channel="9/edge", num_units=3, trust=True, base="ubuntu@24.04")
```
reneradoi: We should always deploy the local version instead of the released one from Charmhub, otherwise we cannot test our code changes. Please see the `charm` fixture in `tests/integration/conftest.py`.

tests/integration/backup/test_s3_backup.py
```diff
@@ -0,0 +1,111 @@
...
+            "secret-key": microceph["secret-key"],
+        },
+    )
+
+    _wait_active(juju, S3_INTEGRATOR_APP)
+    juju.integrate(APP_NAME, S3_INTEGRATOR_APP)
+    _wait_active(juju, APP_NAME, S3_INTEGRATOR_APP)
+
+    # Backup from unit/0
+    task0 = juju.run(f"{APP_NAME}/0", "create-backup")
```
reneradoi: nit: get the unit name dynamically, e.g. `unit = list(juju.status().get_units(APP_NAME))[0]`.

reneradoi (commented):
Only a few comments from my side, the main one is about always deploying the local version of the charm, not the one from Charmhub, because what's the point of testing then ;)

Thank you Mykola!

tests/integration/backup/conftest.py
```diff
@@ -0,0 +1,152 @@
...
+
+import boto3
+import pytest
+from botocore.config import Config
+
+RGW_SSL_PORT = 445
+# Self-signed cert + key, persisted so repeated local runs reuse the exact
+# material the already-running gateway serves -- regenerating without
+# re-enabling RGW would break TLS verification against the old certificate.
+_CERT_DIR = Path.home() / ".cache" / "valkey-itest"
```
skourta: Why name it `valkey-itest` we can have it generic across products I usually run OS, etcd, and Valkey on the same microceph.

tests/integration/backup/conftest.py
skourta: I believe `run` was added in python subprocess to rule them all.

tests/integration/backup/test_s3_backup.py
```diff
@@ -0,0 +1,111 @@
...
+            )
+            and jubilant.all_agents_idle(status, *apps)
+        ),
+        timeout=timeout,
+    )
+
+
+def test_backup_and_list(juju: jubilant.Juju, microceph: dict, s3_bucket) -> None:
+    juju.deploy(APP_NAME, channel="9/edge", num_units=3, trust=True, base="ubuntu@24.04")
+    juju.deploy(S3_INTEGRATOR_APP, channel="latest/edge")
```
skourta: Let's please use `2/edge`. I believe `latest/edge` is a very old one and not updated/maintained.

tests/integration/backup/test_s3_backup.py
```diff
@@ -0,0 +1,111 @@
...
+
+    # Credentials are supplied through the sync-s3-credentials action, not a
+    # config option. Wait for the agent to settle (charm deployed, action
+    # registered) before dispatching -- a freshly-allocated unit reports
+    # "no actions defined" until then.
+    juju.wait(
+        lambda status: jubilant.all_agents_idle(status, S3_INTEGRATOR_APP),
+        timeout=600,
+    )
+    juju.run(
```
skourta: These are passed as a secret for version 2 of the s3-integrator

skourta (commented):
Tested against s3-integrator 2/edge. Everyting works. I left some comments.

tests/integration/backup/conftest.py
delgod: Done. _run now wraps subprocess.run(..., check=True, capture_output=True) and returns stdout; _ok is gone. Failures now raise CalledProcessError with stderr attached instead of being swallowed.

tests/integration/backup/conftest.py
delgod: Done. _run now wraps subprocess.run(..., check=True, capture_output=True) and returns stdout; _ok is gone. Failures now raise CalledProcessError with stderr attached instead of being swallowed.

tests/integration/backup/conftest.py
delgod: Done, exactly as suggested — _ensure_user, _ensure_microceph, and the RGW re-enable branch now use try/except CalledProcessError.

tests/integration/backup/test_s3_backup.py
delgod: Done. Dropped _wait_active; both waits now use are_apps_active_and_agents_idle(..., idle_period=30) with delay=5, successes=3.

tests/integration/backup/test_s3_backup.py
delgod: Fixed. Now deploys the charm fixture with resources=IMAGE_RESOURCE on k8s (same pattern as test_charm.py); no longer pulls 9/edge from Charmhub.

tests/integration/backup/test_s3_backup.py
delgod: Done. units = list(juju.status().get_units(APP_NAME)), then index in. Kept three distinct units so the any-unit guarantee is still exercised.

tests/integration/backup/conftest.py
delgod: Renamed to ~/.cache/microceph-rgw — it's the MicroCeph gateway cert, so it's shared across OS/etcd/Valkey on one MicroCeph.

tests/integration/backup/test_s3_backup.py
delgod: Switched to 2/edge.

tests/integration/backup/test_s3_backup.py
delgod: Done. Dropped sync-s3-credentials; now add-secret + grant-secret and set credentials=<uri> on s3-integrator 2/edge. tls-ca-chain config is unchanged (still present in v2).

reneradoi (approved):
Looks good!

skourta (approved):
LGTM thank you @delgod

PR #64, opened 2026-05-22
test(ha): harden flaky tests

delgod:
test_network_cut_primary flakes in CI for two independent reasons:

- VM: the new-primary election wait was ~30s (stop_after_attempt(4) x wait_fixed(10)), but Sentinel's down-after-milliseconds is 30s (src/managers/config.py), so the final attempt landed exactly when failover begins -- before any replica is promoted. Widen the window to ~140s, still under the 180s failover-timeout. The loop breaks as soon as a new primary appears, so the happy path is unaffected.

- K8s: applying the Chaos Mesh NetworkChaos manifest races the mnetworkchaos.kb.io admission webhook coming up and fails with "connection refused". Retry the idempotent apply for a bounded window so the webhook has time to start serving.

- [ ] Have you updated relevant documentation?

### DIFF
--- a/tests/integration/ha/test_network_cut.py
+++ b/tests/integration/ha/test_network_cut.py
@@ -144,8 +144,11 @@ def test_network_cut_primary(  # noqa: C901
     logger.info("Verifying new primary election...")
 
     new_primary_ip = None
-    # failover should happen after 30s
-    for attempt in Retrying(stop=stop_after_attempt(4), wait=wait_fixed(10), reraise=True):
+    # Sentinel only marks the old primary down after `down-after-milliseconds` (30s,
+    # see src/managers/config.py) and then needs time to reach quorum and promote a
+    # replica, so a new primary cannot appear before ~30s. Wait well past that (but
+    # under the 180s `failover-timeout`) to absorb CI scheduling jitter.
+    for attempt in Retrying(stop=stop_after_attempt(15), wait=wait_fixed(10), reraise=True):
         with attempt:
             try:
                 # we exclude the old primary ip because on k8s the unit is reachable by ip
--- a/spread.yaml
+++ b/spread.yaml
@@ -8,7 +8,7 @@ backends:
     allocate: |
       hash=$(python3 -c "import hashlib; print(hashlib.sha256('$SPREAD_PASSWORD'.encode()).hexdigest()[:6])")
       VM_NAME="${VM_NAME:-${SPREAD_SYSTEM//./-}-${hash}}"
-      DISK="${DISK:-20}"
+      DISK="${DISK:-25}"
       CPU="${CPU:-4}"
       MEM="${MEM:-8}"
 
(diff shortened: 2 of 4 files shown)


tests/integration/ha/test_network_cut.py
```diff
@@ -144,8 +144,11 @@ def test_network_cut_primary(  # noqa: C901
     logger.info("Verifying new primary election...")
 
     new_primary_ip = None
-    # failover should happen after 30s
-    for attempt in Retrying(stop=stop_after_attempt(4), wait=wait_fixed(10), reraise=True):
+    # Sentinel only marks the old primary down after `down-after-milliseconds` (30s,
+    # see src/managers/config.py) and then needs time to reach quorum and promote a
+    # replica, so a new primary cannot appear before ~30s. Wait well past that (but
+    # under the 180s `failover-timeout`) to absorb CI scheduling jitter.
+    for attempt in Retrying(stop=stop_after_attempt(15), wait=wait_fixed(10), reraise=True):
```
Mehdi-Bendriss, line 151: Shouldn't we assume that for a small cluster (not 10s of Gb of data) that the failover should happen fast (right after `down-after-milliseconds`) and if it takes longer it's abnormal and potentially a bug to be investigated ?

Mehdi-Bendriss (commented):
Thanks Mykola! Looks good, I only have 1 question

tests/integration/ha/test_network_cut.py
delgod: In my personal opinion, failover should happen under 1-3 seconds with `failover-timeout = 2 x down-after-milliseconds`. But definitely separate spec.

skourta (approved):

skourta (commented):
Can we run the CI a few times to confirm the stability?

.github/workflows/integration_test.yaml
```diff
@@ -126,6 +126,40 @@ jobs:
         with:
           pattern: ${{ inputs.artifact-prefix }}-*
           merge-multiple: true
+      - name: Free disk space on /
```
reneradoi: Did you add this for a particular reason or is it a leftover before we moved the integration tests to Github hosted runners? From what I can see, we usually have 80 to 90 GB of disk space available _after_ the env setup, which should be enough for the tests we run.

reneradoi (commented):
Just a small question from my side, other than that I'm fine with the changes.

tests/integration/ha/test_network_cut.py
Mehdi-Bendriss: tracked [here](https://github.com/canonical/valkey-operator/issues/66)

Mehdi-Bendriss (approved):

.github/workflows/integration_test.yaml
delgod: Each of my PRs failed CD/CD at least once because of free space :(

reneradoi (approved):
Thank you Mykola!

.github/workflows/integration_test.yaml
delgod: removed

PR #65, opened 2026-05-22
fix(ha): don't wedge the app when scale-down lock can't be acquired

delgod:
Removing the whole application fires storage-detaching on every unit at once. Each unit races for a single scale-down lock; the losers raised RequestingLockTimedOutError, which failed the hook and parked the unit in error state. Its agent is then torn down before juju can retry the hook, so the unit stays wedged and the application can never finish removing — `juju.wait(APP not in status)` times out (seen in k8s/test_scaling.py:: test_scale_down_remove_application).

- ScaleDownLock.request_lock now genuinely retries under contention. The retry count was `min(timeout // 5, 1)` (always 1) and a contended SET NX returns falsy WITHOUT raising, so tenacity never re-polled. Replace it with an explicit poll loop using `max(...)` so the timeout is honored. On retry it re-resolves the primary with the resilient get_primary_ip_for_scale_down() (which rides out transient sentinel failures during teardown) and builds the ValkeyClient once.

- _on_storage_detaching no longer raises when the lock is unavailable during full-app removal (planned_units() == 0): it goes away gracefully instead of wedging, logging at warning with the unit name. A partial single-unit scale-down still raises so juju retries the hook.

- While acquiring the lock, ValkeyCannotGetPrimaryIPError and ValkeyWorkloadCommandError (e.g. from get_active_sentinel_ips) are tolerated the same way: go away on full-app removal, otherwise re-raise so juju retries — rather than silently skipping failover and the dataset save on a partial scale-down.

Covered by unit tests in tests/unit/test_locks.py and tests/unit/test_scaledown.py.

- [ ] Have you updated relevant documentation?

### DIFF
--- a/src/common/locks.py
+++ b/src/common/locks.py
@@ -8,8 +8,6 @@ import time
 from abc import abstractmethod
 from typing import TYPE_CHECKING, Protocol, override
 
-from tenacity import Retrying, stop_after_attempt, wait_fixed
-
 from common.client import ValkeyClient
 from core.cluster_state import ClusterState
 from literals import CharmUsers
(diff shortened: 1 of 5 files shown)


src/events/base_events.py
```diff
@@ -547,8 +548,44 @@ def _on_storage_detaching(self, event: ops.StorageDetachingEvent) -> None:
...
+            # free. Don't wedge this unit in error: its agent gets torn down
+            # before juju can retry the hook, which would leave the whole
+            # application stuck mid-removal. Just go away.
+            if self.charm.app.planned_units() == 0:
+                logger.info(
+                    "%s: whole application is being removed; going away without the scale-down lock.",
+                    self.charm.state.unit_server.unit_name,
+                )
+                self._set_state_for_going_away()
+                return
```
skourta, line 588: If we return here we will not save the databse to disk before stopping the workload. The unit will just be killed by juju

src/common/locks.py
```diff
@@ -246,38 +244,47 @@ def request_lock(self, timeout: int | None = None, primary_ip: str | None = None
...
-                ):
-                    logger.debug(
-                        "%s acquired %s lock.", self.charm.state.unit_server.unit_name, self.name
-                    )
-                    return True
+        # Poll the lock: a contended SET NX returns falsy WITHOUT raising, so we
+        # must re-attempt on a fixed interval until it is acquired or the timeout
+        # elapses. With no timeout we make a single attempt and let the caller
+        # decide how to wait (e.g. defer the hook so juju re-runs it later).
+        number_of_retries = max(timeout // 5, 1) if timeout else 1
```
skourta, line 251: This is a good catch. I thinkwe should still use tenacity to manage the retries though.

src/events/base_events.py
```diff
@@ -547,8 +548,44 @@ def _on_storage_detaching(self, event: ops.StorageDetachingEvent) -> None:
...
+                # peers that would coordinate are going away too). There is nothing left
+                # to coordinate with, so go away rather than wedge the unit in error.
+                logger.warning(
+                    "%s: primary/sentinels unreachable while acquiring the scale-down lock "
+                    "during full-app removal; going away. (%s)",
+                    self.charm.state.unit_server.unit_name,
+                    e,
+                )
+                self._set_state_for_going_away()
+                return
```
skourta, line 569: On full app removal, we currently still orchestrate the shutdown to be executed one unit at a time but whther we change that to all units doing it at the same time or not, saving the DB to disk on line 636 should always be executed before the workload is shutdown.

src/literals.py
```diff
@@ -51,6 +51,11 @@
 
 DATA_STORAGE = "data"
 
+# How long a unit waits in-process to acquire the scale-down lock before either
+# deferring the storage-detaching hook (so juju re-runs it) or, on full-app
+# removal, going away. Kept well under juju's hook timeout.
+SCALE_DOWN_LOCK_TIMEOUT = 30
```
skourta, line 57: I think we should increase it to something like 300. This should enable most units to avoid erroring out in storage detaching

PR #67, opened 2026-06-05
patch: Fix jhack tail on the CI

skourta:
This pull request makes a small change to the integration test workflow to improve the reliability of the `jhack tail` step. The command is now wrapped with `script` and will not fail the workflow if the command itself fails.

- Workflow robustness:
  * Updated the `jhack tail` step in `.github/workflows/integration_test.yaml` to run the command inside a `script` shell, redirect output, and ensure the step does not fail the workflow if the command fails.


delgod (approved):

reneradoi (approved):
Thank you Smail!

Mehdi-Bendriss (approved):

PR #74, opened 2026-06-22
[DPE-10363] bugfix: set correct directory ownership when attaching a storage volume

reneradoi:
This PR fixes a bug related to deploying on Canonical K8s.

**Issue:**
When a storage volume is attached, some K8s clouds (such as Canonical K8s) create a `lost+found` folder (owned by root) on the mount point. This also sets the ownership of the storage path to `root` and makes the directory not accessible for writing files by the charm workload (which uses the `_daemon_` user).

This issue can be seen in this integration test: https://github.com/canonical/data-integrator/actions/runs/27890725877/job/82591369371?pr=280

Another evidence of this case can be found [here](https://github.com/canonical/postgresql-k8s-operator/issues/1009).

**Solution:**
When the storage-attached hook is run, the charm should ensure its storage volumes have the correct permissions and ownership.

**Testing:**
The issue was reproduced locally and the fix verified locally as well. Currently integration tests for Canonical K8s are not yet available.

### DIFF
--- a/src/events/base_events.py
+++ b/src/events/base_events.py
@@ -89,6 +89,8 @@ class BaseEvents(ops.Object):
                 )
             except ValkeyWorkloadCommandError as e:
                 logger.error("Error when ensuring storage ownership: %s", e)
+                event.defer()
+                return
 
         # fix the permissions of the directory if re-attaching existing storage
         self.charm.workload.exec(
--- a/src/events/base_events.py
+++ b/src/events/base_events.py
@@ -93,9 +93,14 @@ class BaseEvents(ops.Object):
                 return
 
         # fix the permissions of the directory if re-attaching existing storage
-        self.charm.workload.exec(
-            ["chmod", "-R", "750", self.charm.workload.working_dir.as_posix()]
-        )
+        try:
+            self.charm.workload.exec(
+                ["chmod", "-R", "750", self.charm.workload.working_dir.as_posix()]
+            )
+        except ValkeyWorkloadCommandError as e:
+            logger.error("Error when setting storage permissions: %s", e)
+            event.defer()
+            return
 
     def _on_install(self, event: ops.InstallEvent) -> None:
         """Handle install event."""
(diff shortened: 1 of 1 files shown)


src/events/base_events.py
```diff
@@ -76,7 +76,16 @@ def __init__(self, charm: "ValkeyCharm"):
     def _on_storage_attached(self, event: ops.StorageAttachedEvent) -> None:
         """Handle storage attachment."""
         if self.charm.state.substrate == Substrate.K8S:
-            return
+            # some K8s clouds create a lost+found folder (owned by root) when attaching a volume
+            # we need to ensure the path is accessible for the charm
+            self.charm.workload.exec(
+                [
```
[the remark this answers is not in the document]
reneradoi: Pebble connection errors are handled in a separate PR.

src/events/base_events.py
```diff
@@ -76,7 +76,19 @@ def __init__(self, charm: "ValkeyCharm"):
...
+                self.charm.workload.exec(
+                    [
+                        "chown",
+                        "-R",
+                        f"{self.charm.workload.user}:{self.charm.workload.user}",
+                        self.charm.workload.working_dir.as_posix(),
+                    ]
+                )
+            except ValkeyWorkloadCommandError as e:
+                logger.error("Error when ensuring storage ownership: %s", e)
```
skourta, line 91: Shouldn't we defer to retry at least?

reneradoi: You are right.

skourta (dismissed):
Thank you. This fixes the issue I was just debuging on replicas not being able to sync correctly from primary.
This won't be needed after rootless I believe.

reneradoi:
> This won't be needed after rootless I believe.

Hm, not sure. If the volume mount is with root permissions from the K8s side (and not from the charm code), we would still have this issue I believe.

Mehdi-Bendriss (approved):

skourta (approved):
Thank you

skourta:
We do not have the issue with rootless. And adding this fix actually breaks the charm. We won't be able to `chown` the `lost+found` created dir and the exec will fail. Setting `charm-user: non-root` will make juju automatically add an fsgroup podspec with group 170 and kubelet will chown all file there to 170 group. The charm can then create its files under the `_daemon_` user and group.
image
The `_daemon_` user has the 170 group. 
image

reneradoi:
Perfect, thank we can consider this a "temporary" fix and remove it with the "rootless" PR.

PR #78, opened 2026-06-30
feat(DPE-10424): add logs and archive storage volumes

delgod:
## Summary

Adds two Juju filesystem storages — `logs` and `archive` — to the Valkey charm,
mirroring the `charmed-etcd-operator` layout and adapted for Valkey's dual VM + K8s
substrate. valkey-server and valkey-sentinel now log to the `logs` volume instead of
stdout/journald; `archive` is forward-provisioned as the staging area for a future
backup restore. RDB/AOF persistence is unchanged — it stays on the `data` volume.

## Testing

- `tox run -e lint`, `tox run -e unit`, `tox run -e static` (pyright) — all pass.
- New unit tests cover both substrates; integration test deploys and verifies
  `logs`/`archive` attach and `valkey.log`/`sentinel.log` appear in the volume.


tests/integration/storage/test_storage.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python3
```
reneradoi: There is already an integration test for storage reuse in `tests/integration/ha/test_storage_reuse.py`, it would be good if it could be integrated there instead of having a separate test for it.

tests/unit/test_storage.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
```
reneradoi, line 1: I really like that we have more unit test coverage making use of `ops.testing` this time. While this may not be "true" unit testing, it ensures that also the ops integration of our code is working. For the storages, we could add some more concrete test coverage and actually try to write/read a file through ops on the storages. See the following link for an example/reference: https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/tests/unit/test_charm.py#L332-L354

src/events/base_events.py
```diff
@@ -76,29 +79,36 @@ def __init__(self, charm: "ValkeyCharm"):
         )
 
     def _on_storage_attached(self, event: ops.StorageAttachedEvent) -> None:
-        """Handle storage attachment."""
+        """Set ownership/permissions on the attached storage dir.
+
+        Mirrors the etcd reference: ownership is set here only, never in the
```
reneradoi: nit: Comments are good, but sometimes a bit too verbose. Could you please try to comment only what is necessary?

src/managers/topology.py
```diff
@@ -34,6 +36,25 @@ def __init__(self, state: ClusterState, workload: WorkloadBase):
         self.state = state
         self.workload = workload
 
+    def _open_observer_log(self):
```
reneradoi: I believe we can simplify this method: On K8s, the observer logfile will always be written on the charm container per design (this is where the observer process is created). In this container, there is no `log` storage, so we can return early for K8s.

reneradoi (commented):
Thank you for taking this over, Mykola!

I have only a few comments for consideration. In addition, please coordinate with Smail's PR for "rootless K8s", in which the need for `chown` on storage attachment is removed. We should ensure this also works for the two newly added storages.

src/events/base_events.py
```diff
@@ -55,9 +57,10 @@ def __init__(self, charm: "ValkeyCharm"):
         super().__init__(charm, key="base_events")
         self.charm = charm
 
-        self.framework.observe(
-            self.charm.on[DATA_STORAGE].storage_attached, self._on_storage_attached
-        )
+        for storage_name in (DATA_STORAGE, LOG_STORAGE, ARCHIVE_STORAGE):
+            self.framework.observe(
+                self.charm.on[storage_name].storage_attached, self._on_storage_attached
+            )
```
reneradoi, line 63: We currently only listen to `DATA_STORAGE`.storage_detaching, but I think with this change, we need to listen to all of them, and whichever comes first, should run the `_on_storage_detaching` logic. Otherwise it could happen that the workload crashes if cannot write the logfile anymore. Which is more important in the case of Valkey than it may in other cases, because we want to scale down safely and flush data to disk before stopping the server.

src/managers/topology.py
skourta: I think we should have the same path on both substrates.

src/workload_vm.py
```diff
@@ -114,6 +114,8 @@ def __init__(self) -> None:
         self.acl_file = self.root_dir / SNAP_CURRENT_PATH / SNAP_ACL_FILE
         self.sentinel_acl_file = self.root_dir / SNAP_CURRENT_PATH / SNAP_SENTINEL_ACL_FILE
         self.working_dir = self.root_dir / SNAP_COMMON_PATH / "var/lib/charmed-valkey"
+        self.log_dir = self.root_dir / SNAP_COMMON_PATH / "var/log/charmed-valkey"
+        self.archive_dir = self.root_dir / SNAP_COMMON_PATH / "archive"
```
skourta: This is different from k8s. We usually have all paths similar after SNAP_COMMON_PATH.

tests/integration/storage/test_storage.py
skourta: If we decide to keep a separate file we need to add spread file for it to run on the CI

skourta (commented):
I left some comments thank you @delgod

tests/integration/storage/test_storage.py
delgod: Fixed in c253ee6. Folded the two checks (`logs`/`archive` attached, and `valkey.log`/`sentinel.log` present in the volume) into `tests/integration/ha/test_storage_reuse.py` and deleted the standalone suite.

tests/unit/test_storage.py
delgod: Fixed in 6638971. Added `test_storage_is_a_readable_writable_mount`, parametrised over `logs` and `archive`, following the etcd reference

src/events/base_events.py
delgod: I will try to control how many opuses Opus writes 😬  check now

src/managers/topology.py
delgod: Fixed in bac54ed. Went with @skourta's "same path on both substrates": the observer log stays on the legacy `/var/log/topology_observer.log`.

src/events/base_events.py
delgod: Fixed in d16f64e. I think idempotency is becoming more critical in such cases

src/events/base_events.py
```diff
@@ -71,34 +74,36 @@ def __init__(self, charm: "ValkeyCharm"):
...
-        """Handle storage attachment."""
+        """Set ownership/permissions on the attached storage dir."""
+        storage_dirs = {
+            DATA_STORAGE: self.charm.workload.working_dir,
+            LOG_STORAGE: self.charm.workload.log_dir,
+            ARCHIVE_STORAGE: self.charm.workload.archive_dir,
+        }
+        target = storage_dirs.get(event.storage.name)
+        if target is None:
+            return
```
skourta, line 91: I think this deserves a warning log

skourta (dismissed):
One small comment. LGTM. Thank y ou @delgod

src/events/base_events.py
```diff
@@ -71,34 +74,36 @@ def __init__(self, charm: "ValkeyCharm"):
...
     def _on_storage_attached(self, event: ops.StorageAttachedEvent) -> None:
-        """Handle storage attachment."""
+        """Set ownership/permissions on the attached storage dir."""
+        storage_dirs = {
+            DATA_STORAGE: self.charm.workload.working_dir,
+            LOG_STORAGE: self.charm.workload.log_dir,
+            ARCHIVE_STORAGE: self.charm.workload.archive_dir,
+        }
+        target = storage_dirs.get(event.storage.name)
+        if target is None:
```
Mehdi-Bendriss: nit: walrus

src/events/base_events.py
```diff
@@ -572,6 +584,10 @@ def _on_storage_detaching(self, event: ops.StorageDetachingEvent) -> None:
                 "Backup in progress on this unit; refusing to scale down until it finishes."
             )
 
+        self._scale_down_unit()
```
Mehdi-Bendriss, line 587: Is your view that if we detach a storage volume (e.g archive or logs) without removing the unit, that we should consider the current unit scaled down / removed?

Mehdi-Bendriss (commented):
Thank you Mykola! I only have 1 main question

Mehdi-Bendriss (dismissed):

src/events/base_events.py
delgod: I validated this against Juju on VM. The scenario doesn't actually arise: data/logs/archive are singular storage (no `multiple:`), and juju detach-storage refuses them outright — cannot detach, storage is singular, even with `--force`. So a volume can't be detached without removing the unit in the first place; the charm never sees a storage-detaching for that case. Making them detachable (`multiple: {range: 0-1}`) would work but defaults the count to `0`, so a plain juju deploy would stop provisioning logs/archive — which defeats the point of this PR. So I've left the detach path as-is (scale-down fires on teardown, whichever volume detaches first) and dropped the "recover on re-attach" idea as unreachable.

src/events/base_events.py
delgod: Fixed in https://github.com/canonical/valkey-operator/pull/78/commits/ae10a9f62be43d9a135851cf011c91d10a26289c

src/events/base_events.py
delgod: Fixed in https://github.com/canonical/valkey-operator/pull/78/commits/ae10a9f62be43d9a135851cf011c91d10a26289c

reneradoi (approved):
Thank you Mykola! Please ensure integration tests are passing before merging, currently a lot of them fail because of issues with the snap store.

skourta (approved):
Thank you @delgod

Mehdi-Bendriss (approved):

PR #79, opened 2026-07-01
[DPE-10650] S3 backup restoration (restore-backup action)

delgod:
## What

  Adds a `restore-backup` action that restores the cluster's dataset from an S3
  backup produced by `create-backup`. Serves both **in-place rollback** and
  **disaster recovery** onto a freshly-deployed cluster.

  juju run valkey/leader restore-backup backup-id=2026-06-29T14:32:45Z

  Leader-only and asynchronous: the action validates and initiates, then the
  operator watches `juju status` (a `RESTORE_IN_PROGRESS` maintenance status)
  until the cluster returns to `active`.

  ## Approach

  valkey is primary/replica via Sentinel with automatic full-resync, so — unlike
  etcd's restore-every-node model — restore lands the RDB on **only the current
  primary** and lets replicas resync from it. Coordination is an etcd-style
  databag state machine (extends the existing `BackupManager`/`BackupEvents`):

  `DOWNLOAD → RESTORE → RESYNC → COMPLETED`, with an app-level target instruction
  vs. a per-unit completed step; the leader advances only when every participant
  reaches the current step. The genuinely hard part is coordinating Sentinel so it
  doesn't fail over while the primary briefly restarts.

  Restore replaces **only the dataset** — ACLs, `CharmUsers` passwords, and TLS are
  charm-managed config and are untouched.
  
  ## Design highlights

- **Failover suppression:** raises `down-after-milliseconds` on every sentinel
  for the restore window (`DOWNLOAD → RESYNC`); **symmetric teardown** resumes it
  on *every* abort path, so a failed restore can't leave Sentinel failover
  disabled cluster-wide.
- **Tuple-match dispatch** `(instruction, prior_step)`: a unit acts only from its
  exact prior step — a unit that missed DOWNLOAD can never run the destructive
  RESTORE.
- **Fail-closed barrier:** iterates a snapshotted participant set; a departed
  participant stalls (never silently dropped), so a mid-restore scale change
  can't wedge or bypass the gate.
- **Safe RDB handling:** magic-byte validation in-stream during download, written
  to a temp name and atomically renamed onto the final name only on full success
  (a partial download never carries the final name); `dump.rdb` preserved as
  `dump.rdb.pre-restore` for rollback.
- **Ordering/idempotency:** stop only valkey-server (not Sentinel) to bracket the
  swap; `stop_service`-first rollback to defeat supervisor auto-restart;
  idempotent move-aside so a redelivered hook can't clobber the good pre-restore
  copy; generous **bounded** dataset-aware waits (never hang, never false-pass).
- **Restore-awareness:** peer-relation-changed handlers in `base_events`,
  `external_clients`, and `tls`, plus `restart_workload`, skip work during a
  restore; `storage-detaching` **refuses** a scale-down mid-restore (it would
  otherwise issue a manual Sentinel failover and stop a participant).
  
  ## Testing

- **Unit:** 202/202; lint + static clean. Covers each state transition, the
  fail-closed barrier, the tuple-match guard, rollback + suppression-resume on
  any step failure, bounded-wait timeouts, and the guard matrix.
- **Integration** (`tests/integration/backup/`, MicroCeph + s3-integrator):
  rollback, disaster recovery, and a corrupt-restore that asserts the cluster
  keeps its old data **and Sentinel failover still works** afterward.

## Backward compatibility

New peer-databag fields (`restore_id`, `restore_instruction`,
`restore_participants`, `restore_step`, `restore_role`) default falsy and
tolerate an old-revision databag (9/edge rolls unit-by-unit); the guards are
pure no-ops until a restore is initiated — no impact on normal operation or the
existing backup feature.


src/core/cluster_state.py
```diff
@@ -200,3 +200,30 @@ def get_unit_hostname(self, unit_name: str | None = None) -> str:
...
         """Return the number of units in the cluster that have their Valkey server started."""
         return len([unit for unit in self.servers if unit.model and unit.is_started])
+
+    @property
+    def can_restore_workflow_proceed(self) -> bool:
+        """True only when every snapshotted participant has reached the current instruction.
+
+        Fail-closed: iterate the snapshotted participant *names* and look each
+        up in the live servers. A participant absent from the live set (it
+        departed mid-step) counts as NOT reached, so the gate stalls rather
```
skourta: This can be better explained. I only understood the comment when I read the code.

src/events/backup.py
```diff
@@ -212,4 +226,200 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+            return "A restore is already in progress."
+        # Stable cluster: all participants active + a resolvable primary + no failover in flight.
+        if not all(s.is_active for s in self.charm.state.servers):
+            return "Not all units are active; wait for the cluster to settle."
+        try:
+            primary_ip = self.charm.sentinel_manager.get_primary_ip()
+        except ValkeyCannotGetPrimaryIPError:
+            return "No primary available; cannot restore."
+        if "failover_in_progress" in (
+            self.charm.sentinel_manager._get_sentinel_client()
```
skourta: If we have to use this then we should make it "public". Do we not have a function that checks this?

src/events/backup.py
```diff
@@ -212,4 +226,200 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+            return
+
+        instruction = cluster.restore_instruction
+        step = unit.restore_step
+        role = unit.restore_role  # "" until DOWNLOAD records it
+
+        try:
+            self._run_restore_step(instruction, step, role)
+        except Exception as e:
+            # Broad catch is deliberate: _restore_teardown -> resume_failover() is
```
skourta: Unclear comment

src/events/backup.py
```diff
@@ -212,4 +226,200 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+                bm.set_restore_step(RestoreStep.COMPLETED)
+
+            case _:
+                # Not our turn (tuple doesn't match an expected transition): no-op.
+                return
+
+    def _do_primary_restore(self) -> None:
+        """Re-download the RDB if missing, then restore in-place; roll back on unhealthy state."""
+        bm = self.charm.backup_manager
+        if not self.charm.workload.path_exists(bm._download_path):
```
skourta: Same if the path is used from outside make it "public"

src/events/backup.py
```diff
@@ -212,4 +226,200 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+            bm.download_backup(self.charm.state.cluster.restore_id)
```
skourta: I thought the download step and restore are separate

src/events/backup.py
```diff
@@ -212,4 +226,200 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+        )
+        participants = ",".join(sorted(s.unit_name for s in self.charm.state.servers))
+        self.charm.state.cluster.update(
+            {
+                "restore_id": backup_id,
+                "restore_instruction": RestoreStep.DOWNLOAD.value,
+                "restore_participants": participants,
+            }
+        )
+        event.set_results({"restore": f"initiated for {backup_id}"})
```
skourta, line 316: So the workflow actually starts when the first non leader unit adds some data into its databag and a relation changed is triggered correct?

src/events/backup.py
skourta: This way we download twice the backup file

src/events/backup.py
```diff
@@ -212,4 +226,200 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+        match (instruction, step):
+            case (RestoreStep.DOWNLOAD, RestoreStep.NOT_STARTED):
+                is_primary = bm.is_local_primary()
+                unit.update({"restore_role": "primary" if is_primary else "replica"})
+                if is_primary:
+                    self.charm.sentinel_manager.suppress_failover()
+                    bm.download_backup(cluster.restore_id)
+                bm.set_restore_step(RestoreStep.DOWNLOAD)
+
+            case (RestoreStep.RESTORE, RestoreStep.DOWNLOAD):
```
skourta: Download and restore can be fused. The non primary do not need to wait between them and primary can do both in a single sweep

src/events/backup.py
```diff
@@ -212,4 +226,200 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+            raise
+
+    def _advance_if_leader(self) -> None:
+        """Advance the instruction once every participant has reached it; clear on COMPLETED."""
+        cluster = self.charm.state.cluster
+        if not self.charm.state.can_restore_workflow_proceed:
+            return
+        instruction = cluster.restore_instruction
+        if instruction == RestoreStep.COMPLETED:
+            self.charm.state.statuses.delete(
```
skourta: This should be calculated from relation data at `get_statuses` if possible and avoid adding and removing statuses whenever possible.

src/events/base_events.py
```diff
@@ -292,6 +295,9 @@ def _on_peer_relation_changed(self, _: ops.RelationChangedEvent) -> None:
 
     def _on_peer_relation_departed(self, _: ops.RelationDepartedEvent) -> None:
         """Handle event received by all units when a unit departs."""
+        if self.charm.state.cluster.is_restore_in_progress:
+            return
```
skourta, line 300: Shouldn't we defer?

src/events/external_clients.py
```diff
@@ -193,7 +193,10 @@ def _on_peer_relation_changed(self, event: ops.RelationChangedEvent) -> None:
         This handler catches all changes from scaling operations, TLS switchover, TLS CA rotation,
         IP changes, etc.
         """
-        if not self.charm.state.unit_server.is_started:
+        if (
+            not self.charm.state.unit_server.is_started
+            or self.charm.state.cluster.is_restore_in_progress
```
skourta, line 200: Don't we risk missing an IP change?

src/events/tls.py
```diff
@@ -112,6 +112,9 @@ def _on_peer_relation_created(self, event: ops.RelationCreatedEvent) -> None:
 
     def _on_peer_relation_changed(self, event: ops.RelationChangedEvent) -> None:
         """Handle TLS related changes to the peer relation."""
+        if self.charm.state.cluster.is_restore_in_progress:
+            return
```
skourta, line 118: Same why aren't we deferring?

src/managers/backup.py
```diff
@@ -305,6 +337,153 @@ def create_backup(self) -> str:
...
+        always correct.
+        """
+        s3_parameters = self.state.cluster.s3_credentials
+        if s3_parameters is None:
+            raise ValkeyRestoreError("S3 credentials unavailable")
+        bucket = self._get_bucket_resource(s3_parameters)
+
+        # boto3 writes the whole object into this BytesIO buffer so we can
+        # inspect the magic header before committing the file to disk.
+        buffer = io.BytesIO()
```
skourta: This is very memory consuming. To ensure a proper restore we would need to have at least the size of the rdb file free in memory.

src/managers/backup.py
```diff
@@ -305,6 +337,153 @@ def create_backup(self) -> str:
...
+            raise ValkeyRestoreError(e) from e
+
+        head = buffer.getvalue()[:16]
+        if not head.startswith(_RDB_MAGIC):
+            raise ValkeyRestoreError(
+                f"Downloaded object for {backup_id} is not a valid RDB stream"
+            )
+
+        buffer.seek(0)
+        self.workload.push_data_file(
```
skourta, line 352: I don't understand if we're downloading into memory why do we write the part file then rename it?

skourta: Is it to protect against possible io writing issues?

src/managers/backup.py
```diff
@@ -305,6 +337,153 @@ def create_backup(self) -> str:
...
+                        raise ValkeyRestoreUnhealthyError("still loading RDB")
+        except Exception as e:  # tenacity reraises the last attempt error
+            raise ValkeyRestoreUnhealthyError(
+                f"Primary did not come up healthy within {RESTORE_LOAD_TIMEOUT_S}s"
+            ) from e
+
+    def wait_until_resynced(self) -> None:
+        """Bounded poll until this replica reports a connected, in-sync link.
+
+        Purpose-built: the stock ``wait_for_replica_fully_synced`` has no ceiling
```
skourta: Comment also hard to understand

tests/integration/backup/test_s3_restore.py
```diff
@@ -0,0 +1,305 @@
...
+
+@pytest.mark.abort_on_fail
+def test_restore_disaster_recovery(
+    juju: jubilant.Juju,
+    microceph: dict,
+    s3_bucket,
+    substrate: Substrate,
+) -> None:
+    """Remove the app entirely, redeploy a fresh cluster, restore from S3 -- data comes back."""
+    _write_key(juju, "dr_key", "dr-value")
```
skourta, line 199: I would very much like our tests to be independently executable. If the cluster is not deployed we can deploy one.

tests/integration/backup/test_s3_restore.py
```diff
@@ -0,0 +1,305 @@
...
+    time.sleep(15)
+    juju.wait(
+        lambda s: jubilant.all_agents_idle(s, APP_NAME),
+        timeout=600,
+        delay=5,
+    )
+
+    # Old data must still be present (restore rolled back or never committed).
+    got = _read_key(juju, _leader_unit_name(juju), "safe_key")
+    assert got == "safe-value", f"Old data lost after corrupt restore; got {got!r}"
```
skourta, line 272: Why aren't we checking on status?

skourta (commented):
I left several comments/questions I will do manual testing next.

src/core/cluster_state.py
delgod: Fixed in b411469

src/events/backup.py
delgod: Fixed in ec94e9f

src/events/backup.py
delgod: Fixed in b411469

src/events/backup.py
delgod: Fixed in c2ae788

src/events/backup.py
delgod: Fixed in c2ae788

src/events/backup.py
delgod: Fixed in c2ae788

src/events/base_events.py
delgod: Deliberately return, not defer: deferring a *-relation-departed event is unsafe (on redelivery the departed unit/databag may be gone), and scale-down is already refused during a restore (storage-detaching raises), so a departure shouldn't happen mid-restore anyway. Quorum reconciles on the next relation event. Rationale documented in b411469.

src/events/external_clients.py
delgod: Not missed, just deferred to the post-restore reconcile. This handler reconciles from current state (it's delta-independent), and restore completion clears restore_id — an app-databag write that fires another relation-changed. So an IP change arriving mid-restore is applied on that next reconcile, not dropped.
Documented in b411469; kept return rather than defer on purpose.

src/events/tls.py
delgod: Same pattern as external_clients: return during restore, and restore completion re-fires relation-changed, which re-runs this handler from current state. CA rotation/renewal would otherwise collide with the primary's restart. Documented in b411469.

src/managers/backup.py
delgod: Fixed in 2883142

src/managers/backup.py
delgod: Fixed in 2883142

src/managers/backup.py
delgod: Fixed in b411469

tests/integration/backup/test_s3_restore.py
delgod: Fixed in eeaf72c

tests/integration/backup/test_s3_restore.py
delgod: Fixed in eeaf72c

src/events/backup.py
delgod: Not quite — it starts on the leader. The restore-backup action (leader-only) writes `restore_id` / `restore_instruction=RESTORE` / `restore_participants` to the app databag in `_on_restore_action`. That app-data write emits relation-changed to every unit, which drives `_on_restore_workflow`; the leader also advances via `update_status`. Each unit writing its own `restore_step` is a consequence of processing a step, not the trigger.

src/events/backup.py
delgod: `RESTORE_IN_PROGRESS` already is derived in `get_statuses` from `is_restore_in_progress`. `RESTORE_FAILED`/`RESTORE_UNHEALTHY` are terminal outcomes of the last restore, and once the restore databag is cleared there's no state left to derive them from — pure derivation would mean adding a `restore_last_result` databag field just to drive the status. I kept the explicit add-on-teardown / delete-on-next-start for now; happy to add the field if you'd prefer no imperative status writes. up to Mehdi

src/managers/backup.py
```diff
@@ -305,6 +311,176 @@ def create_backup(self) -> str:
...
+        a failed query. Sync is judged by ``is_replica_synced`` (ROLE link state).
+        """
+        try:
+            for attempt in Retrying(
+                stop=stop_after_delay(RESTORE_RESYNC_TIMEOUT_S),
+                wait=wait_fixed(5),
+                reraise=True,
+            ):
+                with attempt:
+                    if not self.state.charm.cluster_manager.is_replica_synced():  # pyright: ignore[reportAttributeAccessIssue]
```
reneradoi: This line highlights a conflict with the general design patterns that I noticed: By adding methods like `wait_until_resynced` or `is_local_primary` to the backup manager, it effectively takes over responsibilites from the cluster and sentinel managers. This results in calling one manager from another, which we should avoid, because it creates dependencies between the managers.

As a resolution, I recommend to call the cluster and/or sentinel manager from the backup events to check whether a replica is synced, or whether a Valkey instance is the primary. If the existing methods in the managers don't provide what is required (e.g. retry mechanism), it should be added there.

src/events/backup.py
```diff
@@ -212,4 +209,198 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+            return
+
+        if self.charm.unit.is_leader():
+            self._advance_if_leader()
+
+    def _run_restore_step(self, instruction: RestoreStep, step: RestoreStep, role: str) -> None:
+        """Run exactly the step whose (instruction, prior-step) tuple matches. Else no-op."""
+        unit = self.charm.state.unit_server
+        bm = self.charm.backup_manager
+
```
reneradoi, line 395: nit: Can we avoid these? It makes it harder to understand the code if it is always required to check what exactly `unit.update` is, whereas `self.charm.state.unit_server.update` is precise and understandable at every single point in the code.

src/events/backup.py
```diff
@@ -212,4 +209,198 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+            case (RestoreStep.COMPLETED, RestoreStep.RESYNC):
+                if role == "primary":
+                    bm.cleanup_restore_files()
+                bm.set_restore_step(RestoreStep.COMPLETED)
+
+            case _:
+                # Not our turn: tuple doesn't match a valid transition.
+                return
+
+    def _do_primary_restore(self) -> None:
```
reneradoi, line 437: This is backup/restore business logic and should be moved to the backup manager.

src/events/backup.py
```diff
@@ -212,4 +209,198 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+        # missing) while valkey still serves, so a bad backup-id never bounces
+        # the primary. Outside the try: nothing has changed, nothing to roll back.
+        bm.verify_backup_is_rdb(self.charm.state.cluster.restore_id)
+        try:
+            bm.restore_on_primary()
+        except Exception:
+            bm.roll_back()
+            raise
+
+    def _advance_if_leader(self) -> None:
```
reneradoi, line 465: This is backup/restore business logic and should be moved to the backup manager.

src/managers/backup.py
```diff
@@ -305,6 +311,176 @@ def create_backup(self) -> str:
...
+    def verify_backup_is_rdb(self, backup_id: str) -> None:
+        """Confirm the S3 object starts with the RDB magic, cheaply, before touching valkey.
+
+        A ranged GET of just the first bytes (not the whole object), so a missing
+        or non-RDB backup-id fails while valkey is still serving -- the primary is
+        stopped only once we know the object is plausibly a real snapshot.
+        """
+        s3_parameters = self.state.cluster.s3_credentials
+        if s3_parameters is None:
+            raise ValkeyRestoreError("S3 credentials unavailable")
```
reneradoi, line 320: No need to check here, we do gatekeeping when the `restore` action is started and once in progress, the credentials will not be removed anymore.

src/managers/backup.py
```diff
@@ -305,6 +311,176 @@ def create_backup(self) -> str:
...
+        renames onto ``dump.rdb`` (same partition -> atomic), so the final file
+        never appears partial or invalid. The caller (restore_on_primary) has
+        already moved the old dump aside to the archive partition, so the data
+        partition holds only ~1x the dataset and the new dump is written once,
+        with no cross-device copy to install it.
+        """
+        s3_parameters = self.state.cluster.s3_credentials
+        if s3_parameters is None:
+            raise ValkeyRestoreError("S3 credentials unavailable")
+        bucket = self._get_bucket_resource(s3_parameters)
```
reneradoi, line 343: Same comment: No need to do that here.

src/managers/backup.py
```diff
@@ -305,6 +311,176 @@ def create_backup(self) -> str:
...
+
+        # Stream to a temp file on disk so we can check the magic header first.
+        with tempfile.NamedTemporaryFile() as tmp:
+            try:
+                bucket.download_fileobj(f"{s3_parameters.path}/{backup_id}", tmp)
```
reneradoi: If I am not mistaken, this will download the whole object. What happens if the size of the backup file exceeds the size of the charm container? Shouldn't we stream this to the archive volume directly, which exists for exactly this case?

src/managers/backup.py
```diff
@@ -305,6 +311,176 @@ def create_backup(self) -> str:
...
+        """
+        self.workload.stop_service(self.workload.valkey_service)
+        # On a redelivered hook _pre_restore_path already holds the ORIGINAL data;
+        # don't overwrite it (the only rollback copy). Skip the aside if it exists.
+        if not self.workload.path_exists(self._pre_restore_path):
+            self.workload.move_file(self._dump_path, self._pre_restore_path)
+        # Data partition is now free; download the restore RDB directly onto it.
+        self.download_backup(self.state.cluster.restore_id)
+        self.workload.start_service(self.workload.valkey_service)
+        self._wait_until_loaded()
```
reneradoi: We can decouple here and also have clearer responsibilites by deviding the restore and the post-check: `wait_until_loaded` should be in the cluster manager and called from the event handler, where errors can be handled accordingly. This would also allow for the valkey client to be removed from the backup manager, which currently is a syndrom of "cluttered" design, as the backup manager does not actually need to connect to Valkey itself to perform the tasks it is responsible for (other than `self.workload.exec_stream` for creating the backup, which is encapsulated in the workload already).

src/events/external_clients.py
reneradoi: I like this concept, the early return also helps to not run unnecessary logic and thereby improve performance.

tests/integration/backup/test_s3_restore.py
```diff
@@ -0,0 +1,312 @@
...
+from __future__ import annotations
+
+import base64
+import logging
+import time
+from datetime import datetime, timedelta, timezone
+
+import jubilant
+import pytest
+from tests.integration.backup.test_s3_backup import (
```
reneradoi: Let's move them to a helpers file in the `backup` directory if we need them in multiple tests.

tests/integration/backup/test_s3_restore.py
```diff
@@ -0,0 +1,312 @@
...
+def _deploy_cluster_and_s3(juju: jubilant.Juju, microceph: dict) -> None:
+    """Deploy the valkey cluster + s3-integrator and wire up the S3 relation.
+
+    Idempotent: skips individual steps when the relevant apps / relations
+    already exist (safe to call at the top of each test).
+    """
+    status = juju.status()
+
+    if APP_NAME not in status.apps:
+        juju.deploy(APP_NAME, channel="9/edge", num_units=3, trust=True, base="ubuntu@24.04")
```
reneradoi: Same comment as in the PR for backup integration test: We ALWAYS want to test against our local charm version, otherwise the test is pointless.

tests/integration/backup/test_s3_restore.py
```diff
@@ -0,0 +1,312 @@
...
+
+    # Credentials go through the sync-s3-credentials action, not config;
+    # wait for the agent to settle before dispatching so the action is registered.
+    juju.wait(
+        lambda s: jubilant.all_agents_idle(s, S3_INTEGRATOR_APP),
+        timeout=600,
+    )
+    juju.run(
+        f"{S3_INTEGRATOR_APP}/0",
+        "sync-s3-credentials",
```
reneradoi: Please update this test with the feedback from the backup PR; this action has been replaced by a user secret.

reneradoi (commented):
Hi Mykola! In addition to the detailed comments below, I want to add a couple of general feedback points:

- I like the design of the restore workflow, but I think the separation of concerns can be improved. This might seem a little nitpicking now, but for the long term, this will be very important to maintain clear responsibilities per manager, and clear decoupling between event handling and business logic.
- The integration test has a lot of points that we already brought as feedback for the backup integration test PR. I skipped detailled review and would like to ask you to first update the integration test with the feedback from the previous PR.
- The unit tests should go through ops.testing and be based on events, as far as possible. I see a lot of very low-level unit testing, which is contrary to what we typically do. I would like to keep the style of our project, for a variety of reasons. One of them being that the integration of components through ops has to work at any time (for example `data-interfaces` for the peer relation). Another one is that any kind of refactoring will end up in a total rework of unit tests, which is not what we typically want.

In general, I agree with the concept of the restore workflow. Testing from my side will follow when the design and logic is settled.

src/managers/backup.py
delgod: Fixed. Moved `is_local_primary` → `ClusterManager.is_primary`, and the bounded `wait_until_resynced` / `wait_until_loaded` into `ClusterManager` (raising a generic `ValkeyClusterNotReadyError` on timeout); the backup events call them directly. The backup manager no longer reaches into `cluster_manager` via `state.charm`, and holds no Valkey client for restore. bd32226, e1e9900

src/events/backup.py
delgod: Done — dropped the `unit`/`cluster`/`bm` locals in the restore handlers; each access spells out `self.charm.state.unit_server` / `.cluster` / the manager in full. 1326b81

src/events/backup.py
delgod: The low-level work is now all in the manager — `restore_on_primary` (stop/move-aside/download/start), `roll_back`, and `verify_backup_is_rdb` are `BackupManager` methods. What remains in `_do_primary_restore` is orchestration + error handling: verify → restore → `cluster_manager.wait_until_loaded` → `roll_back` on failure. Two constraints keep that in the event rather than the manager:

- Per your #42, the post-check (`wait_until_loaded`) is called from the event handler "where errors can be handled accordingly," and the rollback wrapper has to cover *both* the restore and that post-check — so it belongs where both are called.
- `verify_backup_is_rdb` deliberately runs *outside* the rollback try (before the primary is stopped), which is the regression fix in 52031de — a bad backup-id must fail while the primary still serves, with nothing to roll back.
- Folding this into the manager would reintroduce the manager→manager coupling from #35 (the backup manager calling `cluster_manager.wait_until_loaded`).

So I've kept `_do_primary_restore` as thin event-layer orchestration.

src/events/backup.py
delgod: https://github.com/canonical/valkey-operator/pull/79#discussion_r3538323645

src/managers/backup.py
delgod: You're right that it can't happen at runtime — `_on_s3_credentials_gone` defers while a restore is in progress (`backup.py:127`), so the credentials can't disappear mid-restore. The reason the check stays is typing, not defense: `state.cluster.s3_credentials` is `S3Parameters | None`, so `_get_bucket_resource(s3_parameters)` needs the `None` narrowed or pyright fails. It's type-narrowing, not a duplicated business rule.
Should I suppress the typing linter?

src/managers/backup.py
delgod: https://github.com/canonical/valkey-operator/pull/79#discussion_r3538347825

src/managers/backup.py
delgod: Fixed. `download_backup` no longer buffers the object in a charm-local temp file — it passes the S3 `StreamingBody` straight into `push_data_file`, which copies in bounded chunks (VM `copyfileobj`, K8s Pebble push), so the charm never holds more than a chunk regardless of RDB size. The full object streams onto the data partition (`dump.rdb.part` → atomic rename); the magic header is validated up front by `verify_backup_is_rdb` (a tiny ranged GET, not a full download). e0d16fa

src/managers/backup.py
delgod: Done — `wait_until_loaded` is now a `ClusterManager` method, called from the event after `restore_on_primary`; `restore_on_primary` no longer health-waits internally. All restore Valkey-client usage is out of the backup manager. One caveat: the `ValkeyClient` import stays because `create_backup`'s `_build_rdb_command` (pre-existing, from #59) still uses it to build the `valkey-cli --rdb -` argv — removing that is a change to the merged backup-create path, out of scope for this PR. bd32226, e1e9900.

tests/integration/backup/test_s3_restore.py
delgod: Done — moved the shared setup into `tests/integration/backup/helpers.py` (`deploy_and_relate_s3`, `S3_INTEGRATOR_APP`, `BACKUP_ID_RE`); both the backup and restore tests now use it. 10a1c90.

tests/integration/backup/test_s3_restore.py
delgod: Fixed — the restore test now deploys the local charm via the `charm` fixture (`deploy_and_relate_s3`), same as the backup test, instead of a Charmhub revision. 10a1c90.

tests/integration/backup/test_s3_restore.py
delgod: Fixed — switched to the Juju user-secret flow (`add_secret` → `grant_secret` → `credentials` config) against s3-integrator 2/edge; the `sync-s3-credentials` action is gone. Shared with the backup test via the helper. 10a1c90.

src/events/backup.py
skourta: `update_status` is unreliable. But even if it is disabled when a unit writes to its databag the leader will continue correct?

src/events/backup.py
skourta: @Mehdi-Bendriss

src/managers/backup.py
```diff
@@ -393,84 +383,24 @@ def set_restore_step(self, step: RestoreStep) -> None:
...
-        # already holds the ORIGINAL data. A second unconditional move-aside would
-        # overwrite it with the (possibly corrupt) restored dump.rdb, destroying the
-        # only copy we can roll back to. Skip the aside if the target already exists.
+        # On a redelivered hook _pre_restore_path already holds the ORIGINAL data;
+        # don't overwrite it (the only rollback copy). Skip the aside if it exists.
         if not self.workload.path_exists(self._pre_restore_path):
             self.workload.move_file(self._dump_path, self._pre_restore_path)
-        self.workload.move_file(self._download_path, self._dump_path)
+        # Data partition is now free; download the restore RDB directly onto it.
+        self.download_backup(self.state.cluster.restore_id)
```
skourta, line 410: We should download the backup file before we stop the service. If the file is very large we can still serve users until we are ready. Just to minimize the downtime.

skourta (commented):
Thank you @delgod. I left a couple of comments until I do some manual testing.

src/events/backup.py
```diff
@@ -33,12 +43,8 @@
 def _safe_error(exc: ValkeyBackupError) -> str:
     """Render a backup error safe to return in an action result.
 
-    Action results are readable by any Juju user, unlike ``juju
-    debug-log``. The raw exception text can carry the S3 endpoint,
-    request/host ids, or RDB stream metadata, so only the structured S3
-    error code -- a fixed, non-sensitive token such as "AccessDenied" --
-    is surfaced. Everything else collapses to a generic message; the full
-    detail stays in the unit log.
+    Action results are world-readable, so surface only the structured S3 error
```
Mehdi-Bendriss:
```suggestion
    Action results are world-readable, so surface only the structured object-storage error
```

src/events/backup.py
```diff
@@ -88,20 +99,15 @@ def _on_s3_credentials_changed(
             event.defer()
             return
 
-        # Parse + normalise + validate the integrator envelope in one step:
-        # S3Parameters trims whitespace, strips the separators that would
-        # corrupt S3 key paths, and rejects an envelope missing a required
-        # field or whose path/bucket strips to empty.
+        # S3Parameters trims/validates the envelope and rejects missing/empty fields.
```
Mehdi-Bendriss: can we replace `envelope` by `object storage integrator payload` or something more natural ?

src/events/backup.py
```diff
@@ -132,17 +141,14 @@ def _on_create_backup_action(self, event: ops.ActionEvent) -> None:
             event.set_results({"error": reason})
             event.fail(reason)
             return
-        # Audit the invocation itself, not just the manager-level transfer
-        # (P1-24): ties a specific Juju action run to the resulting backup,
-        # for forensics if an RDB later turns up somewhere unexpected.
+        # Audit the action invocation (P1-24): ties an action run to its backup.
```
Mehdi-Bendriss: 2 questions here: 
1. What does `P1-24` mean? 
2. what do we audit here? or do we mean by that logging?

src/events/backup.py
```diff
@@ -212,4 +215,195 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+        return None
+
+    def _on_restore_action(self, event: ops.ActionEvent) -> None:
+        """Validate, then initiate the async restore workflow (leader only)."""
+        if reason := self._restore_blocking_reason():
+            event.set_results({"error": reason})
+            event.fail(reason)
+            return
+        backup_id = event.params.get("backup-id", "")
+        if not backup_id:
```
Mehdi-Bendriss: nit, walrus 😬

src/events/backup.py
```diff
@@ -212,4 +215,195 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+            event.set_results({"error": "Must provide backup-id to restore."})
+            event.fail("Must provide backup-id to restore.")
+            return
+        try:
+            if backup_id not in self.charm.backup_manager.list_backups():
+                event.fail(f"backup-id {backup_id} not found.")
+                return
+        except ValkeyBackupError as e:
```
Mehdi-Bendriss, line 290: can we log the exception here?

src/events/backup.py
```diff
@@ -212,4 +215,195 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+            event.fail("Could not list backups. Check juju debug-log.")
+            return
+
+        logger.info(
+            "audit: restore action invoked action_id=%s backup_id=%s unit=%s",
+            event.id,
+            backup_id,
+            self.charm.unit.name,
+        )
+        # Clear any stale RESTORE_FAILED from a prior attempt.
```
Mehdi-Bendriss: 2 things here:
1. what else clears the status after a failed restore? (if the user doesn't re-run the restore action do we just keep the status?)
2. should we set a restore in progress status here or until the flow actually starts?

src/events/backup.py
```diff
@@ -212,4 +215,195 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+        )
+        self.charm.state.statuses.add(
+            status.value,
+            scope="app",
+            component=self.charm.backup_manager.name,
+        )
+        if self.charm.unit.is_leader():
+            self.charm.state.cluster.update(
+                {"restore_id": "", "restore_instruction": "", "restore_participants": ""}
+            )
```
Mehdi-Bendriss: I struggled with this one and took me a while to reason about it, but i think this introduces a significant risk where a failed restore can leave the charm **as a whole** stuck in restore mode when **the valkey primary is not the juju leader**

I tracked down the flow and here is what i see:

The restore workflow stores the global `restore_id` (so "restore in progress") in `state.cluster.restore_id`, which is in the app-level relation data, so this "state" can only be cleared from the juju leader.

The problem is that the unit doing the actual restore is the valkey primary, not Juju leader. So if sentinel says `valkey/1` is the primary, then `valkey/1` runs the RDB swap, even if the juju leader is `valkey/0`.

Now:
1. juju leader sets `restore_id`
2. valkey primary starts the restore
3. restore fails on the valkey primary
4. that unit calls `_restore_teardown()`
5. `_restore_teardown()` only clears `restore_id` when the unit is the juju leader
6. if the failing primary is not the juju leader, `restore_id` remains set forever

That leaves the charm thinking a restore is still in progress since `is_restore_in_progress` is now used as guardrail and the main global orchestration in all the flows: backup/restore actions, peer_relation_changed, scale-down / storage-detaching etc.

similarly, the same is can also happen later on a replica during the `resync` step, because replicas invoke the same `_restore_teardown()`  if `wait_until_resynced()` times out. Even though replicas do not perform the RDB restore, they can still fail the workflow and be unable to clear the global restore marker.

**Note:** we will need an integration test for this (juju leader != primary)

src/events/backup.py
```diff
@@ -212,4 +215,195 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+        stop -> back up the current dump -> download the restore RDB onto the
+        data partition -> restart happen inside restore_on_primary; the cluster
+        manager then confirms the server came up and finished loading. Any failure
+        (bad download, unhealthy server) rolls back to the pre-restore copy before
+        propagating to teardown.
+        """
+        # Pre-stop gate: reject an object that isn't a real RDB (wrong magic /
+        # missing) while valkey still serves, so a bad backup-id never bounces
+        # the primary. Outside the try: nothing has changed, nothing to roll back.
+        self.charm.backup_manager.verify_backup_is_rdb(self.charm.state.cluster.restore_id)
```
Mehdi-Bendriss, line 447: Before `restore_on_primary` which will stop the workload first thing, we should `self.charm.cluster_manager.save_database_blocking()` so we persist all the memory data on disk. (This should be done before the `try`)

src/events/backup.py
```diff
@@ -212,4 +215,195 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+            {"restore_instruction": self.charm.backup_manager.next_restore_step(instruction).value}
+        )
+
+    def _restore_teardown(self, exc: Exception | None = None) -> None:
+        """Resume failover, flag failure, and (leader) clear restore state.
+
+        RESTORE_UNHEALTHY when the cluster never became ready
+        (ValkeyClusterNotReadyError), RESTORE_FAILED otherwise.
+        """
+        self.charm.sentinel_manager.resume_failover()
```
Mehdi-Bendriss: This can raise an exception and this method will exit without gracefully setting that the restore has failed and removing the restore_id.

src/events/backup.py
```diff
@@ -116,8 +122,11 @@ def _on_s3_credentials_changed(
 
     def _on_s3_credentials_gone(self, event: StorageConnectionInfoGoneEvent) -> None:
         """Handle removal of the S3 credentials relation."""
-        if self.charm.state.unit_server.is_backup_in_progress:
-            logger.warning("Backup in progress; deferring credentials_gone")
+        if (
+            self.charm.state.unit_server.is_backup_in_progress
+            or self.charm.state.cluster.is_restore_in_progress
+        ):
+            logger.warning("Backup or restore in progress; deferring credentials_gone")
```
Mehdi-Bendriss: in `_on_s3_credentials_changed` during a restore, what is the failure recovery mechanism? basically how will the charm behave during a restore? (and during a backup)

src/events/backup.py
```diff
@@ -212,4 +215,195 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+
+    def _restore_teardown(self, exc: Exception | None = None) -> None:
+        """Resume failover, flag failure, and (leader) clear restore state.
+
+        RESTORE_UNHEALTHY when the cluster never became ready
+        (ValkeyClusterNotReadyError), RESTORE_FAILED otherwise.
+        """
+        self.charm.sentinel_manager.resume_failover()
+        status = (
+            RestoreStatuses.RESTORE_UNHEALTHY
```
Mehdi-Bendriss, line 512: Are these 2 statuses ever cleaned / removed? I'm not sure they are, especially in the case of say a failed restore etc..

src/managers/backup.py
```diff
@@ -305,6 +307,112 @@ def create_backup(self) -> str:
...
+        self.workload.start_service(self.workload.valkey_service)
+
+    def roll_back(self) -> None:
+        """Restore the pre-restore dump and restart (stop_service FIRST to defeat auto-restart)."""
+        self.workload.stop_service(self.workload.valkey_service)
+        if self.workload.path_exists(self._pre_restore_path):
+            self.workload.move_file(self._pre_restore_path, self._dump_path)
+        self.workload.start_service(self.workload.valkey_service)
```
Mehdi-Bendriss: 1. Similar comment on post-restart operations. 
2. this method is funny, it seems like it's the exact same yet opposite of `restore_on_primary`

src/managers/backup.py
```diff
@@ -305,6 +307,112 @@ def create_backup(self) -> str:
...
+        confirms the server came up healthy (cluster_manager.wait_until_loaded).
+        """
+        self.workload.stop_service(self.workload.valkey_service)
+        # On a redelivered hook _pre_restore_path already holds the ORIGINAL data;
+        # don't overwrite it (the only rollback copy). Skip the aside if it exists.
+        if not self.workload.path_exists(self._pre_restore_path):
+            self.workload.move_file(self._dump_path, self._pre_restore_path)
+        # Data partition is now free; download the restore RDB directly onto it.
+        self.download_backup(self.state.cluster.restore_id)
+        self.workload.start_service(self.workload.valkey_service)
```
Mehdi-Bendriss: Given we are doing a stop / start manually directly, it doesn't do all the graceful pre-stop / post-start operations we do in `_on_restart_workload` (e.g: `reconcile_min_replicas_to_write`, checing that sentinel is healthy etc.).
 
So they should either be all put here, or we should have it orchestrated through a similar graceful rolling mechanism like `_on_restart_workload`

src/events/backup.py
```diff
@@ -212,4 +215,195 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+        if self.charm.state.cluster.is_restore_in_progress:
+            return "A restore is already in progress."
+        # Require a stable cluster: all active, a resolvable primary, no failover in flight.
+        if not all(s.is_active for s in self.charm.state.servers):
+            return "Not all units are active; wait for the cluster to settle."
+        try:
+            self.charm.sentinel_manager.get_primary_ip()
+        except ValkeyCannotGetPrimaryIPError:
+            return "No primary available; cannot restore."
+        if self.charm.sentinel_manager.is_failover_in_progress():
```
Mehdi-Bendriss: Missing exception handling of `ValkeyWorkloadCommandError`

src/events/backup.py
```diff
@@ -212,4 +215,195 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
             return "Valkey is not running on this unit."
         if check_running_operations and self.charm.state.unit_server.is_backup_in_progress:
             return "A backup is already in progress on this unit."
+        if check_running_operations and self.charm.state.cluster.is_restore_in_progress:
+            return "A restore is in progress; backups are paused."
         return None
+
+    def _restore_blocking_reason(self) -> str | None:
```
Mehdi-Bendriss, line 243: 1. I feel like we could extract a lot of common code between `_blocking_reason` and this method 
2. I think we should also sefeguard against TLS transition states
3. I think in `_blocking_reason` we could check it, only if want to `check_running_operations=True` (as the list backups will use the s3 client not valkey)

src/events/backup.py
```diff
@@ -212,4 +215,195 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
             return "Valkey is not running on this unit."
         if check_running_operations and self.charm.state.unit_server.is_backup_in_progress:
```
Mehdi-Bendriss, line 237: Why do we check only on the current unit doing a backup while we check for restore ALL the cluster?

Mehdi-Bendriss (commented):
Thank you Mykola! I really like the concept! I have some comments, one that is major on the juju leader vs valkey primary

src/events/backup.py
delgod: Correct. Multi-unit doesn't rely on update_status — it's the peer relation_changed cascade: the leader advances the app-databag instruction → followers fire relation_changed; each follower records its restore_step → fires relation_changed back on the leader, which runs its step and advances. update_status is just a backstop.

The only case that leaned on it was single-unit (no peer to bounce relation_changed). Fixed in 6a17488: _on_restore_workflow now drives to a fixed point within the hook, so single-unit completes in one hook; multi-unit unchanged.

src/managers/backup.py
delgod: When a user issues a restore action, they don't care anymore about the uptime of incorrect data. So we keep the 1×-disk design (stop → move the current dump to the archive partition → download onto the freed space). Downloading first needs ~2× the data partition or an archive volume. Uptime of incorrect data isn't worth doubling peak disk usage.

src/events/backup.py
delgod: fixed in 6a83ef1

src/events/backup.py
delgod: Done — went with payload, across events and models (6a83ef1).

src/events/backup.py
delgod: P1-24 was an internal requirement id — dropped it. It's just an audit log line tying the action invocation to the backup it produces (6a83ef1).

src/events/backup.py
delgod: Fixed in 6a83ef1

src/events/backup.py
delgod: Done — logger.exception added (2bd3c9c).

src/events/backup.py
delgod: 1. Both terminal statuses (FAILED/UNHEALTHY) are now cleared on the next restore's re-init and on COMPLETED; a failed one persists intentionally until then.
2. RESTORE_IN_PROGRESS is derived in get_statuses from is_restore_in_progress, not set eagerly here (43b0538).

src/events/backup.py
delgod: Good catch — fixed. The failing unit (often the valkey primary, not the leader) now records a per-unit restore_failed marker on its own databag; the leader observes any participant's marker (failed_restore_kind) and tears the whole restore down, clearing the app-level restore_id. Scoped to a per-attempt restore_token so a stale marker can't wedge a re-run. 43b0538, integration-tested in a9ddbf0.

src/events/backup.py
delgod: Done — added pre-stop, outside the try (nothing to roll back yet) (2bd3c9c).

src/events/backup.py
delgod: Covered by the teardown fix: the step runs under a catch-all that records the failure marker and (leader) clears restore_id, so a raise no longer wedges the restore (43b0538).

src/events/backup.py
delgod: A real credentials rotation is now deferred while a backup or restore is in progress — the CA is still stored, current creds stay in the databag, and it re-applies when the op finishes (da1136f).

src/events/backup.py
delgod: Yes now — both cleared via _clear_terminal_restore_statuses on the next restore's re-init and on COMPLETED (43b0538).

src/managers/backup.py
delgod: Post-restart min-replicas is now reasserted for both the restore and the rollback restart (in a finally). The roll_back/restore_on_primary symmetry is intentional — rollback is the mirror image (2bd3c9c).

src/managers/backup.py
delgod: Went with "put them here": inline pre-stop save_database_blocking + post-start reconcile_min_replicas_to_write. No sentinel-health gate — restore restarts only valkey-server (not sentinel), failover is suppressed then resumed + SENTINEL RESET, and valkey readiness is gated by wait_until_loaded (2bd3c9c).

src/events/backup.py
delgod: Done — now caught in _unstable_primary_reason (2bd3c9c).

src/events/backup.py
delgod: 1. Extracted _unstable_primary_reason for the shared primary/failover checks; left the 2-line S3 checks inline (a helper there is more indirection than it's worth).
2. TLS-transition guard added (is_tls_transitioning).
3. Restore preflight uses the valkey path; list-backups stays on the S3 client (2bd3c9c).

src/events/backup.py
delgod: Intentional: a backup runs on a single unit, so it only blocks a second backup on that same unit; a restore is cluster-wide. (Separately, _on_s3_credentials_gone was widened to cluster-wide in da1136f.)

src/managers/backup.py
```diff
@@ -350,4 +461,7 @@ def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObje
         if scope == "app" and self.state.s3_relation and not self.state.cluster.s3_credentials:
```
reneradoi: nit: `and self.state.unit_server.is_started` to avoid this status being displayed when a relation is present from deploy time but the credentials have not been applied yet. This could happen with Terraform deployments.

src/events/backup.py
reneradoi: Let's please add a short comment that single-unit deployments are triggered through the `peer-relation-changed` on app data.

src/events/backup.py
```diff
@@ -212,4 +236,322 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+                logger.exception("resume_failover during leader restore teardown failed")
+        status = (
+            RestoreStatuses.RESTORE_UNHEALTHY
+            if self.charm.state.failed_restore_kind == RestoreFailure.UNHEALTHY.value
+            else RestoreStatuses.RESTORE_FAILED
+        )
+        # Clear restore state FIRST, so a status-write failure (statuses.add,
+        # unlike delete, doesn't swallow one) can't leave the restore wedged.
+        self._clear_restore_state()
+        try:
```
reneradoi: Let's follow existing patterns and avoid error handling for adding or deleting statuses. This should be considered safe and errors be handled through the lib. Error handling around every status handling would unnecessarily inflate the lines of code, especially with this broad kind of exception handling.

src/events/backup.py
```diff
@@ -212,4 +236,322 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+        """Leader-only: flag the failure (UNHEALTHY vs FAILED) and clear restore state.
+
+        ``resume`` resumes failover as a backstop, since a failing peer's own
+        best-effort resume may have raised. The leader-self teardown already
+        resumed, so it passes resume=False to avoid a redundant SENTINEL RESET.
+        """
+        if resume:
+            try:
+                self.charm.sentinel_manager.resume_failover()
+            except Exception:
```
reneradoi: This should only raise `ValkeyWorkloadCommandError`.

src/events/backup.py
```diff
@@ -212,4 +236,322 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+    ) -> None:
+        """Drive the restore state machine as far as this unit can in one hook.
+
+        Each pass runs this unit's matching step and (leader) advances the shared
+        instruction once all participants catch up, looping until no local
+        progress is made. So a single-unit restore finishes in one hook; a
+        multi-unit one runs to the next cross-unit barrier, where the peer
+        relation_changed cascade (update_status as backstop) resumes it.
+        """
+        while True:
```
reneradoi: I don't think we need a while loop here. In single-unit deployments the workflow should also proceed through updates to the app databag. Or why did you add it?

src/events/backup.py
```diff
@@ -212,4 +236,322 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+
+        restore_on_primary does stop -> move dump aside -> download -> restart;
+        the caller then confirms it loaded. Any failure rolls back to the
+        pre-restore copy before propagating to teardown.
+        """
+        # Pre-stop (outside the try, nothing to roll back yet): reject a non-RDB
+        # object before bouncing the primary, and persist memory to disk so the
+        # rollback copy restore_on_primary moves aside is current, not stale.
+        self.charm.backup_manager.verify_backup_is_rdb(self.charm.state.cluster.restore_id)
+        self.charm.cluster_manager.save_database_blocking()
```
reneradoi: Heads up: This has been refactored, you should now use `cluster_manager.save_dataset_before_shutdown` which is save against pebble kill.

Edit: This is also the reason why the unit tests are failing now.

src/events/backup.py
```diff
@@ -212,4 +236,322 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+            self.charm.backup_manager.roll_back()
+            raise
+        finally:
+            # A restart (restore or rollback) resets runtime min-replicas-to-write
+            # to the rendered 1, so reassert the topology-correct value or a small
+            # cluster is write-frozen. Best-effort: a raise in a finally must not
+            # mask the failure or false-fail a success.
+            try:
+                self.charm.cluster_manager.reconcile_min_replicas_to_write()
+            except Exception:
```
reneradoi, line 462: This should only raise a `ValkeyWorkloadCommandError`.

src/events/backup.py
```diff
@@ -212,4 +236,322 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+            role = self.charm.state.unit_server.restore_role  # "" until RESTORE records it
+
+            try:
+                self._run_restore_step(instruction, step, role)
+            except Exception as e:
+                # Catch everything: teardown must record the failure marker and
+                # resume failover on ANY error (service-control errors sit outside
+                # the restore-error hierarchy), or the restore wedges.
+                logger.exception("Restore step failed; tearing down")
+                self._restore_teardown(e)
```
reneradoi: How do we give feedback to users about a failed restore on the leader unit? If I understand correctly, the restore state is cleaned up on teardown, and on the next execution, a restore is no longer in progress and we just return. 

So from a user perspective, it looks like the restore was successful (confirmed by testing).

src/managers/backup.py
reneradoi: No don't suppress the static type check, we can leave it as is.

src/workload_k8s.py
```diff
@@ -197,6 +197,78 @@ def restart(self, service: str) -> None:
                 f"Failed to start service {service}: {e}"
             ) from e
 
+    @override
+    def stop_service(self, service: str) -> None:
```
reneradoi: We should integrate this and `start_service` into the existing methods`start` and `stop` to avoid code duplication. It should be as simple as adding an optional parameter `service`, and if this is available, stop/start only this service, otherwise all.

src/workload_k8s.py
```diff
@@ -197,6 +197,78 @@ def restart(self, service: str) -> None:
...
+            pebble.TimeoutError,
+            pebble.ConnectionError,
+            pebble.APIError,
+        ) as e:
+            logger.error("Failed to stop service %s: %s", service, e)
+            raise ValkeyServicesCouldNotBeStoppedError(
+                f"Failed to stop service {service}: {e}"
+            ) from e
+        try:
+            if self.container.get_service(service).is_running():
```
reneradoi: Please use `self.alive` instead.

src/workload_k8s.py
```diff
@@ -197,6 +197,78 @@ def restart(self, service: str) -> None:
...
+    def start_service(self, service: str) -> None:
+        try:
+            self.container.start(service)
+        except (pebble.ChangeError, pebble.ConnectionError, pebble.APIError) as e:
+            raise ValkeyServicesFailedToStartError(
+                f"Failed to start service {service}: {e}"
+            ) from e
+
+    @override
+    def service_running(self, service: str) -> bool:
```
reneradoi: Please use `alive` instead (can be adjusted to check only one service as well).

src/workload_vm.py
```diff
@@ -190,6 +191,60 @@ def restart(self, service: str) -> None:
                 "Failed to restart service %s: %s", service, e
             ) from e
 
+    @override
+    def stop_service(self, service: str) -> None:
```
reneradoi: Same comments on integrating into the existing methods please, to avoid code duplication.

reneradoi (commented):
I've found a couple of things that we need to fix/clarify. In addition: The LDAP feature has been merged, please also consider LDAP event handlers for restore safeguards.

src/managers/backup.py
delgod: Done in 4ab475b — gated on unit_server.is_started.

src/events/backup.py
delgod: Done in 4ab475b — now save_dataset_before_shutdown() (also fixes the failing tests).

src/events/backup.py
delgod: Done in 4ab475b — narrowed to ValkeyWorkloadCommandError.

src/workload_k8s.py
delgod: Done in 9f1c325 — folded into start/stop via an optional service param; start_service/stop_service removed.

src/workload_k8s.py
delgod: Done in 9f1c325 — now verifies via alive(service).

src/workload_k8s.py
delgod: Done in 9f1c325 — service_running removed; alive takes an optional service.

src/workload_vm.py
delgod: Done in 9f1c325 — same refactor on VM.

src/events/backup.py
delgod: Drives a single-unit restore to its fixed point in one hook; Juju doesn't guarantee a unit sees relation-changed for its own app-databag writes, so otherwise it stalls until update-status.

src/events/backup.py
delgod: Teardown sets a blocked status (RESTORE_FAILED/UNHEALTHY) and clears restore_id, so it should show blocked.

src/events/backup.py
reneradoi: I understand that this is guaranteed, because the single unit will always be the leader. From the docs:

```
Note that units only receive relation-changed events for other units’ changes. The one exception to this 
rule is for peer relations’ application data bags. The application leader will receive a relation-changed 
event for the changes that it writes to a peer relation’s application data bag. 
```

https://canonical.com/juju/docs/juju-cli/3.6/reference/hook/#endpoint-relation-changed

src/events/backup.py
delgod: LLM Answer: reconcile already swallows ValkeyWorkloadCommandError internally, so narrowing here is dead code — and it's in a finally, so a non-ValkeyWorkloadCommandError (e.g. a raw OSError from the VM CLI exec) would escape and false-fail a successful restore.

src/events/backup.py
delgod: Done in c919929d0b4a71394ca57725b120431bd2787c7c — dropped the wrapper. _clear_restore_state() runs first, so a status-write failure can't wedge; the hook just re-runs.

src/workload_k8s.py
```diff
@@ -267,6 +310,12 @@ def stop(self) -> None:
             raise ValkeyServicesCouldNotBeStoppedError(
                 f"Failed to stop Valkey services: {e}"
             ) from e
+        # Single-service stop verifies the SPECIFIC service went down (a sibling
+        # staying up is expected); the all-services stop needs no such check.
+        if service and self.alive(service):
+            raise ValkeyServicesCouldNotBeStoppedError(
```
reneradoi: How about a more clean approach?

```
def stop(service: str|None=None, check_alive: bool=False):
    ...
    if not check_alive:
        return
    
    if self.alive(service=service if service else None):
        raise ...
```

src/workload_k8s.py
```diff
@@ -171,17 +171,22 @@ def pebble_layer(self) -> pebble.Layer:
...
+                    self.valkey_service, self.sentinel_service, self.metric_service
+                )
         except (
             pebble.ChangeError,
             pebble.ConnectionError,
             pebble.APIError,
         ) as e:
             raise ValkeyServicesFailedToStartError(f"Failed to start Valkey services: {e}") from e
-        if not self.alive():
+        if not service and not self.alive():
```
reneradoi: Same idea here to be more clean and simple:
```
def start(service: str|None=None, check_alive: bool=True):
    ...
    if not check_alive:
        return
    
    if not self.alive(service=service if service else None):
        raise
```

In the existing call(s) of `start/stop` you don't have to change anything, in the calls from backup/restore, you can add the `check_alive` parameter according to your needs.

By this, we can avoid intransparency and ambiguous behavior.

Same applies in VM workload.

src/events/backup.py
reneradoi: I've tested without the while loop on a single unit, and it works. By putting the restore-id into the peer app data, it triggers a relation changed. Subsequent relation changes are triggered by updating the restore instructions one-by-one. Also failure cases work, by resetting the restore-id in peer app data, it also advances the workflow.

reneradoi (commented):
Some final comments from my side for the sake of maintainability, other than that I'm good. Testing was fine, also a few error cases (invalid backup file, pebble services stopped during restore, ...) where the restore was stopped and the database recovered.

src/events/backup.py
reneradoi: I've confirmed that this works, thank you.

src/events/backup.py
delgod: Added. 51840f9.

src/events/backup.py
delgod: You were right — verified and removed the loop. 51840f9.

src/workload_k8s.py
delgod: Thank you; check_alive is now an explicit param. 57d5281.

src/workload_k8s.py
delgod: thank you, answer https://github.com/canonical/valkey-operator/pull/79#discussion_r3635711082

src/workload_k8s.py
reneradoi: Nice, thank you!

src/core/cluster_state.py
```diff
@@ -215,6 +216,66 @@ def number_units_started(self) -> int:
...
+        more specific status.
+        """
+        token = self.cluster.restore_token
+        participants = set(self.cluster.restore_participants)
+        kinds = {
+            server.restore_failure_kind(token)
+            for server in self.servers
+            if server.unit_name in participants
+        }
+        kinds.discard("")
```
skourta: Why do we need this?

src/events/backup.py
skourta: Let's add logs in all scenarios for tracibility in the logs in case of errors.

src/events/backup.py
```diff
@@ -212,4 +236,316 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+        the caller then confirms it loaded. Any failure rolls back to the
+        pre-restore copy before propagating to teardown.
+        """
+        # Pre-stop (outside the try, nothing to roll back yet): reject a non-RDB
+        # object before bouncing the primary, and persist memory to disk so the
+        # rollback copy restore_on_primary moves aside is current, not stale.
+        self.charm.backup_manager.verify_backup_is_rdb(self.charm.state.cluster.restore_id)
+        self.charm.cluster_manager.save_dataset_before_shutdown()
+        try:
+            self.charm.backup_manager.restore_on_primary()
```
skourta: should we use `valkey-check-rdb` to check the file is valid before we restore?

src/events/backup.py
```diff
@@ -212,4 +236,316 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+            self.charm.cluster_manager.wait_until_loaded(RESTORE_LOAD_TIMEOUT_S)
+        except Exception:
+            self.charm.backup_manager.roll_back()
+            raise
+        finally:
+            # A restart (restore or rollback) resets runtime min-replicas-to-write
+            # to the rendered 1, so reassert the topology-correct value or a small
```
skourta: Can you please explain why it resets the config to 1?

src/events/backup.py
```diff
@@ -212,4 +236,316 @@ def _blocking_reason(self, check_running_operations: bool = True) -> str | None:
...
+        instruction = self.charm.state.cluster.restore_instruction
+        if instruction == RestoreStep.COMPLETED:
+            self._clear_terminal_restore_statuses()
+            self._clear_restore_state()
+            return
+        self.charm.state.cluster.update(
+            {"restore_instruction": self.charm.backup_manager.next_restore_step(instruction).value}
+        )
+
+    def _restore_teardown(self, exc: Exception | None = None) -> None:
```
skourta: I'm confused what a teardown is. Is it just clearing state and statuses?

src/events/external_clients.py
```diff
@@ -193,7 +193,12 @@ def _on_peer_relation_changed(self, event: ops.RelationChangedEvent) -> None:
         This handler catches all changes from scaling operations, TLS switchover, TLS CA rotation,
         IP changes, etc.
         """
-        if not self.charm.state.unit_server.is_started:
+        # Skip during a restore; not dropped — restore completion clears restore_id,
+        # firing another relation-changed that reconciles from current state.
+        if (
```
skourta: There was a helper for this on the backup event handler maybe move is to state and reuse it?

src/managers/cluster.py
```diff
@@ -152,6 +165,46 @@ def wait_for_replica_fully_synced(self, primary_endpoint: str):
...
+                        client.info_persistence(hostname=self.state.endpoint).get("loading", "1")
+                        != "0"
+                    ):
+                        raise ValkeyClusterNotReadyError("still loading dataset")
+        except Exception as e:  # tenacity reraises the last attempt's error
+            raise ValkeyClusterNotReadyError(
+                f"Server did not become ready within {timeout_s}s"
+            ) from e
+
+    def wait_until_resynced(self, timeout_s: int) -> None:
```
skourta: This feels redundant maybe fuse the functions and add a boolean?

tests/integration/backup/test_s3_restore.py
```diff
@@ -0,0 +1,385 @@
...
+    """Return the unit name of the current Valkey primary node."""
+    primary_ip = get_primary_ip(juju, APP_NAME)
+    return get_unit_name_from_primary_ip(juju, primary_ip, substrate)
+
+
+# ── tests ────────────────────────────────────────────────────────────────────
+
+
+@pytest.mark.abort_on_fail
+def test_restore_rollback(
```
skourta: Can we please add a one valkey unit cluster test for the happy paht?
This is mostly because the syncing mechanims are different from multiple units

skourta (commented):
Thank you @delgod.
When manually testing I noticed:
- We add the status about missing/wrong credentials early on even in the creds are valid. It is then cleared when all units have them. This is a little bad from a UX point of view.
- Very hard to follow the restore on the logs. We hardly log anything.

reneradoi (approved):
Thank you Mykola!

src/events/backup.py
delgod: According to LLM research, `valkey-check-rdb` can fail in some cases, such as empty keys; worse, it passes on duplicate keys, which would fail on a real startup.
So running the server straight away is more reliable.

src/core/cluster_state.py
delgod: `restore_failure_kind()` is `""` for units that didn't fail this attempt, so the set needed the discard. Folded into the comprehension — https://github.com/canonical/valkey-operator/pull/96/changes/26fed56a58c454ca07bf5f8e198068f07b8e3688.

src/events/backup.py
delgod: Done — restore.* log trail on every rejection, step, advance, completion, download, rollback and failure — https://github.com/canonical/valkey-operator/pull/96/changes/26fed56a58c454ca07bf5f8e198068f07b8e3688

src/events/backup.py
delgod: valkey.conf ships min-replicas-to-write 1; the 0-below-3-units value is a non-persistent CONFIG SET, so any restart reloads 1 and a 1–2 unit cluster would be write-frozen. Comment reworded — https://github.com/canonical/valkey-operator/pull/96/changes/26fed56a58c454ca07bf5f8e198068f07b8e3688

src/events/backup.py
delgod: Nothing is torn down (rollback already happened). Renamed: _restore_teardown → _fail_restore (unit records its failure marker), _clear_failed_restore → _finish_failed_restore (leader clears app state + sets terminal status) — https://github.com/canonical/valkey-operator/pull/96/changes/26fed56a58c454ca07bf5f8e198068f07b8e3688

src/events/external_clients.py
delgod: This already is the state property. The backup-handler helper also folds in a cluster-wide backup check, which must not pause client-endpoint reconciliation, so it isn't the same predicate. Kept.

src/managers/cluster.py
delgod: Fused behind _wait_until(condition, timeout, failure) with a predicate instead of a boolean —https://github.com/canonical/valkey-operator/pull/96/changes/c0efccc3d1160577873201af3f4f8cb21081b23b

tests/integration/backup/test_s3_restore.py
delgod: Added test_restore_single_unit (redeploy 1 unit → backup → mutate → restore → verify → write again); passed locally on VM — https://github.com/canonical/valkey-operator/pull/96/changes/aeb49693c3765bf55ee5b3ace0f3d3427c7eb820

src/managers/sentinel.py
```diff
@@ -376,26 +376,34 @@ def is_failover_in_progress(self) -> bool:
...
+        re-render (Sentinel persists SET to its own conf, so a restart won't).
         """
         client = self._get_sentinel_client()
         for endpoint in self.all_sentinel_endpoints():
-            client.set(
+            if not client.set(
                 endpoint, PRIMARY_NAME, "down-after-milliseconds", str(SENTINEL_DOWN_AFTER_MS)
-            )
+            ):
+                logger.warning("Failed to reset down-after-milliseconds on sentinel %s", endpoint)
```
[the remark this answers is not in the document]
delgod: Yes, for a sentinel unreachable at resume time (it persists SET to its conf). Fixed: each unit re-checks its own sentinel on update-status outside a restore and resets it — https://github.com/canonical/valkey-operator/pull/96/changes/fd774a1d2c74e85a923491d816b02595e00dbc10

src/core/models.py
```diff
@@ -260,6 +266,25 @@ def restore_role(self) -> str:
...
+
+    def restore_failure_kind(self, restore_token: str) -> str:
+        """Return this unit's failure kind for ``restore_token``, else "".
+
+        Token-scoped so a stale marker from a prior attempt (restore_id is the
+        backup-id and repeats on a re-run) isn't misread against the current one.
+        """
+        if not self.model or not self.model.restore_failed:
+            return ""
+        kind, _, token = self.model.restore_failed.partition(":")
```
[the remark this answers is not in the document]
delgod: partition never raises on a marker without : (empty-token compat); split would need a length guard. Kept.

delgod (commented):
> We add the status about missing/wrong credentials early on even in the creds are valid. It is then cleared when all units have them. 

pre-existing 9/edge behaviour (backup PR #59) — transient "blocked" until the leader has processed the relation.

skourta (approved):
Thank you @delgod!

PR #81, opened 2026-07-06
[DPE-10186] Add LDAP support

reneradoi:
This PR adds LDAP support to Valkey, based on the specification in [DA269](https://docs.google.com/document/d/1guxZipx-h5kYIKDS0hVQR_o35jtOVqRFci7V_TbcVG4/edit?usp=sharing).

Design:
- new charm relations for `ldap` and `ldap-ca-cert`
- receive `entity-permissions` request through client relation from Data Integrator to construct "auth roles" (matched against `ldap-map` config of Valkey)
- new event handler for LDAP related events
- moved the auth related parts of the `ConfigManager` to a new `AuthManager`
- added functionality for LDAP authentication and authorization to the `AuthManager`
- LDAP is enabled/disabled by reloading config and ACLs, no need for restart
- does not require `ldaps`, also works with `startTLS`
- safeguards on config and relation state are checked before enabling LDAP and results are displayed to users with blocked statuses

Workarounds / Limitations:
- added `certificate_transfer` charmlib because `glauth-k8s` only supports the `certificate-transfer` interface in v0
- disabled the connection failure detection of the Valkey LDAP module because it depends on the `WhoAmI` command which is not supported by GLAuth (see https://glauth.github.io/docs/limitations.html)

Unrelated:
- refactor method for `config set` in `ValkeyClient` class
- add `+ping +role` permissions to external client users to allow users to check if they are connected to a primary or replica instance (addresses #80)


src/core/cluster_state.py
```diff
@@ -200,3 +214,93 @@ def get_unit_hostname(self, unit_name: str | None = None) -> str:
...
+
+        for relation in self.external_client_relations:
+            if not relation.app:
+                continue
+
+            repository = OpsRelationRepository(self.model, relation, relation.app)
+            if not (relation_data := repository.get_data()):
+                continue
+
+            if not (requests := relation_data.get("requests", [])):
```
Mehdi-Bendriss: nit, the default `[]` is not needed

src/core/models.py
```diff
@@ -367,3 +378,48 @@ def tls_client_private_key(self) -> PrivateKey | None:
...
+        """LDAP relation data object."""
+        if not self.relation or not self.relation.app:
+            return {}
+
+        return self.relation.data[self.relation.app]
+
+    @property
+    def urls(self) -> list:
+        """The URLs to connect to LDAP."""
+        return ast.literal_eval(self.relation_data.get("urls", ""))
```
Mehdi-Bendriss: i would rather we use `json.loads()`

src/core/models.py
```diff
@@ -367,3 +378,48 @@ def tls_client_private_key(self) -> PrivateKey | None:
...
+
+    @property
+    def ldaps_urls(self) -> list:
+        """The URLs for LDAP over TLS."""
+        return ast.literal_eval(self.relation_data.get("ldaps_urls", ""))
```
Mehdi-Bendriss: similar comment on json.loads

src/core/models.py
```diff
@@ -367,3 +378,48 @@ def tls_client_private_key(self) -> PrivateKey | None:
...
+
+    @property
+    def starttls(self) -> str:
```
Mehdi-Bendriss, line 417: why starttls? doesn't GLAuth and Authentik in the future always support ldaps?

src/events/external_clients.py
```diff
@@ -135,7 +135,7 @@ def _on_bulk_resources_requested(
                 continue
 
             if not (password := self.charm.client_manager.get_password(username)):
-                password = self.charm.config_manager.generate_password()
+                password = self.charm.auth_manager.generate_password()
```
Mehdi-Bendriss, line 138: I really like the `auth_manager` idea!

src/events/ldap.py
```diff
@@ -0,0 +1,238 @@
...
+        self.charm.cluster_manager.reload_ldap_settings(ldap_config)
+
+        logger.info("Update ACL configuration")
+        # only update Valkey ACLs, we do not support LDAP for Sentinel
+        self.charm.auth_manager.set_acl_file()
+        self.charm.cluster_manager.reload_acl_file()
+
+    def _on_sync_ldap_users(self, event: ops.ActionEvent) -> None:
+        """Handle the action event for sync-ldap-users."""
+        if not self.charm.state.unit_server.is_started:
```
Mehdi-Bendriss, line 214: should this action be run on every unit? or only once in the cluster? leader?

src/managers/config.py
```diff
@@ -114,6 +110,14 @@ def get_config_properties(self, primary_endpoint: str) -> dict[str, str]:
         tls_config = self.generate_tls_config()
         config_properties.update(tls_config)
 
+        # LDAP related configuration
+        config_properties["loadmodule"] = self.workload.lib_dir.as_posix() + "/libvalkey_ldap.so"
```
Mehdi-Bendriss, line 114: do we need as_posix when using a Path?

src/managers/config.py
```diff
@@ -184,69 +188,35 @@ def generate_sentinel_tls_config(self) -> dict[str, str]:
...
-            acl_content += f"user {username} on #{password_hash} {permissions}\n"
+        ldap_urls = (
+            self.state.ldap.ldaps_urls if self.state.ldap.ldaps_urls else self.state.ldap.urls
+        )
+        ldap_servers = " ".join(url for url in ldap_urls)
+
+        ldap_config = {
+            "ldap.auth_mode": "search+bind",
+            "ldap.servers": ldap_servers,
+            "ldap.use_starttls": "no" if self.state.ldap.ldaps_urls else "yes",
```
Mehdi-Bendriss, line 209: similar comment on starttls

Mehdi-Bendriss (commented):
Thank you René! Great work and really clean! I only have small comments and questions

src/common/client.py
```diff
@@ -295,21 +277,16 @@ def acl_load(self, hostname: str) -> bool:
         """
         return self.exec_cli_command(["acl", "load"], hostname=hostname) == "OK"
 
-    def reload_tls(self, tls_config: dict[str, str], hostname: str) -> None:
-        """Trigger to load the TLS settings."""
+    def load_config_settings(self, config_settings: dict[str, str], hostname: str) -> None:
```
skourta: nit: the name is a little misleading. When I read load I think get not set. It also breaks the convention of naming the client methods like the valkey cli commands they use

src/statuses.py
```diff
@@ -171,3 +171,29 @@ class BackupStatuses(Enum):
...
+        message="LDAP: Missing `entity-permissions`",
+    )
+
+    LDAP_MAP_PERMISSION_REQUEST_MISSING = StatusObject(
+        status="blocked",
+        message="LDAP: Permission request missing for group in `ldap-map`",
+    )
+
+    LDAP_NOT_ENABLED = StatusObject(
+        status="blocked", message="LDAP not yet enabled, see `debug-log` for details"
```
skourta, line 198: nit: All LDAP statuses start with the prefix except for this one

src/core/cluster_state.py
```diff
@@ -200,3 +214,93 @@ def get_unit_hostname(self, unit_name: str | None = None) -> str:
...
+
+    @property
+    def ldap(self) -> LDAPState:
+        """The ldap relation state."""
+        return LDAPState(
+            relation=self.ldap_relation,
+        )
+
+    @property
+    def is_ldap_valid(self) -> bool:
```
skourta, line 226: nit: maybe we should move this to the manager? This is logic to me.

src/core/cluster_state.py
```diff
@@ -200,3 +214,93 @@ def get_unit_hostname(self, unit_name: str | None = None) -> str:
...
+            return False
+
+        if not self.is_ldap_permission_config_valid:
+            logger.warning("Permission request missing for LDAP group in `ldap-map`")
+            return False
+
+        return True
+
+    @property
+    def is_ldap_permission_config_valid(self) -> bool:
```
skourta, line 256: Same.

src/managers/auth.py
```diff
@@ -0,0 +1,321 @@
...
+
+        Returns:
+            str: ACL lines for the LDAP users.
+        """
+        acl_content = ""
+        if not self.state.is_ldap_valid:
+            return acl_content
+
+        # get non-LDAP users to avoid adding duplicate usernames to ACL files
+        internal_users = (user.value for user in CharmUsers)
```
skourta: This returns a generator. Let's use a set or a list instead.

src/managers/auth.py
```diff
@@ -0,0 +1,321 @@
...
+        client_users = []
+        if external_client_users := self.state.cluster.external_users_credentials:
+            client_users = (username for username in external_client_users)
```
skourta: client_users is declared as a list then here it is replaced by a generator.

src/managers/auth.py
```diff
@@ -0,0 +1,321 @@
...
+    def _get_ldap_connection(self) -> ldap3.Connection:
+        """Get a connection to the related LDAP provider."""
+        ldap_urls = (
+            self.state.ldap.ldaps_urls if self.state.ldap.ldaps_urls else self.state.ldap.urls
+        )
+        ldap_ca_cert = self.workload.read_file(self.workload.tls_paths.ldap_ca)
+
+        tls_context = ldap3.Tls(
+            ciphers="ALL",
+            version=ssl.PROTOCOL_TLSv1,
```
skourta: Is this requirered by glauth? TLS 1 is deprecated I believe.

src/events/ldap.py
```diff
@@ -0,0 +1,238 @@
...
+            event.defer()
+            return
+
+    def _update_ldap_config(self) -> None:
+        """Update the current LDAP configuration."""
+        if not self.charm.state.unit_server.is_started:
+            return
+
+        if (
+            self.charm.state.ldap_ca_cert_relation
```
skourta, line 190: This will also be executed when disabling LDAP. Which causes an extra deferal. When the CA is deleted but the relation still exist.

src/managers/cluster.py
```diff
@@ -192,7 +192,20 @@ def get_version(self) -> str:
...
+            client.load_config_settings(tls_config, hostname=self.state.endpoint)
+        except ValkeyWorkloadCommandError:
+            logger.error("Error loading TLS settings")
+            raise ValkeyTLSLoadError("Could not load TLS settings")
+
+    def reload_ldap_settings(self, ldap_config: dict[str, str]) -> None:
+        """Update Valkey auth by loading the LDAP settings."""
+        if not ldap_config:
+            # disable LDAP by unsetting the server connection
+            ldap_config["ldap.servers"] = ""
```
skourta: nit: declare new dict to avoid mutating the arguemnt passed.

skourta (commented):
Minor comments. Thank you @reneradoi. This is a very clean PR.

src/core/cluster_state.py
reneradoi: updated

src/core/models.py
reneradoi: Good idea, updated

src/core/models.py
reneradoi: updated

src/core/models.py
reneradoi: Depends on the configuration of the LDAP provider: By default, it is configured with `startTLS` only. LDAPS can be enabled, but as we do not require it from our side, I would not narrow it down and allow both (if - for whatever reason - a setup cannot support LDAPS).

src/events/ldap.py
reneradoi: Originally I wanted it to only be run on the leader, but I went away from it and now it should be executed on every unit. This will also be explained in the docs.

Reasoning: If only the leader orchestrates this, it would add a lot of complexity for the workflow, and even more, it would require a feedback loop in case it would fail on one unit. The solution that we have now is much simpler and more robust.

src/managers/config.py
reneradoi: Yes, we need it as a string for the configuration file. Otherwise we would create a new pathops object.

src/managers/config.py
reneradoi: Same concern: Why restricting if not required? We prefer `LDAPS` if the provider supports it, but can fall back to `startTLS` if not.

src/common/client.py
reneradoi: Good point, I've changed it to `config_set`.

src/statuses.py
reneradoi: With the prefix the status message would be `LDAP: LDAP not yet enabled` which sounds ... interesting. So I left it out here.

src/core/cluster_state.py
reneradoi: Agree and had it like this in the first place, but went away from it for two reasons:
- this is also required in other managers, which would create dependency between multiple managers or require code duplication instead
- it exclusively computes state and is more of a validator than business logic

src/core/cluster_state.py
reneradoi: See above

src/managers/auth.py
reneradoi: Fixed

src/managers/auth.py
reneradoi: Fixed

src/managers/auth.py
reneradoi: Uh this is a very good catch, thank you! I confused this to be the protocol which auto-negotiates the highest possible version between server and client, but that was a misconception from my side. I've fixed it, and also moved away from the deprecated value to use `PROTOCOL_TLS_CLIENT` instead.

src/events/ldap.py
reneradoi: Yes this is by design. Otherwise we could end up in a situation where `is_ldap_valid` still returns true even though the relation is going away.

src/managers/cluster.py
reneradoi: Fixed.

src/events/ldap.py
Mehdi-Bendriss: I understand the reasoning, but as this is an action/UX decision, @delgod should approve

reneradoi: I'll update the spec.

reneradoi: I just noticed that I had already updated the relevant part in the spec before I asked Mykola for explicit approval of the action (on Smail's request). It no longer says that this action has to only be run on the leader, and it got approved.

reneradoi: After confirming with Mykola, I have now adjusted the implementation to allow the action only on the leader unit and trigger the user sync on other units through peer relation data.

skourta (approved):
Thank you @reneradoi

src/core/models.py
```diff
@@ -133,6 +132,7 @@ class PeerAppModel(PeerModel):
     tls_client_private_key: ExtraSecretStr = Field(default=None)
     external_client_users: ClientUsersSecret = Field(default="")
     client_user_epoch: float = Field(default=0)
+    ldap_user_epoch: float | int = Field(default=0)
```
Mehdi-Bendriss, line 135: good idea :)

Mehdi-Bendriss (approved):

PR #89, opened 2026-07-20
[DPE-10671] Fix save on shutdown

reneradoi:
This PR addresses #86:
- disables save on scale down
- cleans up remaining "temp" files on storage attached
- removes workarounds on integration tests added previously

The logic changes target both K8s and VM intentionally to avoid deviation between substrates.

### DIFF
--- a/src/common/client.py
+++ b/src/common/client.py
@@ -284,7 +284,7 @@ class ValkeyClient(CliClient):
         for key, value in config_settings.items():
             cmd.append(key)
             cmd.append(value)
-        logger.debug("Loading config settings for LDAP")
+        logger.debug("Loading config settings")
 
         self.exec_cli_command(command=cmd, hostname=hostname, json_output=False)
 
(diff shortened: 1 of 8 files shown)


src/common/client.py
```diff
@@ -284,9 +284,9 @@ def config_set(self, config_settings: dict[str, str], hostname: str) -> None:
         for key, value in config_settings.items():
             cmd.append(key)
             cmd.append(value)
-        logger.debug("Loading config settings for LDAP")
+        logger.debug("Loading config settings")
 
-        self.exec_cli_command(command=cmd, hostname=hostname)
+        self.exec_cli_command(command=cmd, hostname=hostname, json_output=False)
```
reneradoi, line 289: This was a finding at the side, not related to the changes in this PR.

Mehdi-Bendriss (approved):
Thank you René!

skourta (approved):
Thank you

PR #95, opened 2026-08-14
fix(vm):  deflake the integration suite

delgod:
Root-causes and fixes the recurring integration flakes. One charm fix; everything else is test/CI.

### Charm
- **VM**: a unit that changed IP kept serving a certificate with stale SANs — `config-changed`
  can run while Juju's binding still reports the old address, and nothing retried afterwards.
  The reconcile now also runs from `update-status`, so a missed window self-heals. Unit tests added.
- **TLS**: the tls-certificates lib re-emits `certificate_available` even when the
  certificate is unchanged, and the relation is app-shared — so any unit's renewal
  restarted Sentinel on *every* unit. That RestartLock churn is what kept
  `k8s/test_certificate_rotation` from ever settling (the original flaky-TLS failure).
  The handler now skips already-applied certificates: a fingerprint in the unit databag,
  recorded only after the restart is handed off, so a failed apply is still retried.
  Restarting on a real change stays. Unit tests added.
  
### Tests
- The `no_ip_change` network cut now actually cuts (iptables drop). The old 1 kbit throttle let
  Sentinel heartbeats trickle through, so `+odown` quorum never formed and no failover was ever
  attempted — the cause of the recurring `test_network_cut_primary[no_ip_change-*]` failures.
- `test_failover_topology_update` polls for real failover convergence: `sentinel failover`
  returns on initiation, so the agents-idle wait passed instantly and writes hit the demoting primary.
- Waits spanning a client-TLS enable widened to 900s (the rolling sentinel restart behind it
  takes ~6–9 min for 3 units); other waits stay at 600s.
- `juju ssh` timeout in the signal helper sized for the connection (60s), not the instant pkill.

### CI
- Bootstrap Canonical K8s with pod CIDR `10.42.0.0/16`: the default `10.1.0.0/16` collides with
  the GitHub runner's own subnet, so a pod handed the gateway address could not reach the apiserver.
- Capture valkey/sentinel server log files as artifacts (they log to files, not stdout — this is
  what made the network-cut root cause visible).
- Retry the setup network fetches (snap store, uv installer) that killed runs before any test ran.

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -248,7 +248,7 @@ jobs:
                 > ~/logs/"$name"-valkey-server.txt 2>&1
               { sudo juju ssh "$unit" "sudo cat $log_dir/sentinel.log"; } \
                 > ~/logs/"$name"-valkey-sentinel.txt 2>&1
-              { sudo juju ssh "$unit" 'sudo snap logs charmed-valkey.sentinel -n 500'; } \
+              { sudo juju ssh "$unit" 'sudo snap logs charmed-valkey -n 500'; } \
                 > ~/logs/"$name"-valkey-service.txt 2>&1
             done
           fi
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -196,6 +196,42 @@ jobs:
         timeout-minutes: 3
         if: ${{ success() || (failure() && steps.spread.outcome == 'failure') }}
         run: script -q -c 'sudo jhack tail --printer raw --replay --no-watch' /dev/null | tee ~/logs/jhack-tail.txt || true
+      - name: Capture Valkey server logs
+        timeout-minutes: 3
+        # Valkey and Sentinel log to files (config.py sets `logfile`), not stdout/journald,
+        # so `pebble logs` / `snap logs` miss them; read the files directly and keep a
+        # separate service-lifecycle capture. Best-effort -- never fail the job.
+        if: ${{ success() || (failure() && steps.spread.outcome == 'failure') }}
+        env:
+          SPREAD_JOB: ${{ matrix.job.spread_job }}
+        run: |
+          set +e
+          units=$(sudo juju status --format json 2>/dev/null | python3 -c \
+            "import json,sys; d=json.load(sys.stdin); print('\n'.join((d.get('applications',{}).get('valkey',{}).get('units') or {}).keys()))")
+          if [[ "$SPREAD_JOB" == *:tests/spread/k8s/* ]]; then
+            log_dir=/var/log/valkey
+            for unit in $units; do
+              pod=${unit/\//-}  # valkey/0 -> valkey-0
+              { sudo k8s kubectl -n testing exec "$pod" -c valkey -- \
+                cat "$log_dir/valkey.log"; } > ~/logs/"$pod"-valkey-server.txt 2>&1
+              { sudo k8s kubectl -n testing exec "$pod" -c valkey -- \
+                cat "$log_dir/sentinel.log"; } > ~/logs/"$pod"-valkey-sentinel.txt 2>&1
+              { sudo k8s kubectl -n testing exec "$pod" -c valkey -- \
+                pebble changes; } > ~/logs/"$pod"-valkey-service.txt 2>&1
+            done
+          else
+            log_dir=/var/snap/charmed-valkey/common/var/log/valkey
+            for unit in $units; do
+              name=${unit/\//-}
+              { sudo juju ssh "$unit" "sudo cat $log_dir/valkey.log"; } \
+                > ~/logs/"$name"-valkey-server.txt 2>&1
+              { sudo juju ssh "$unit" "sudo cat $log_dir/sentinel.log"; } \
+                > ~/logs/"$name"-valkey-sentinel.txt 2>&1
+              { sudo juju ssh "$unit" 'sudo snap logs charmed-valkey.sentinel -n 500'; } \
+                > ~/logs/"$name"-valkey-service.txt 2>&1
+            done
+          fi
+          true
       - name: Upload logs
         timeout-minutes: 5
         if: ${{ success() || (failure() && steps.spread.outcome == 'failure') }}
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -23,9 +23,19 @@ jobs:
         uses: actions/checkout@v5
       - name: Set up environment
         run: |
+          # snapd intermittently cannot reach the store on a fresh runner
+          retry() {
+            for attempt in 1 2 3
+            do
+              "$@" && return 0
+              echo "::warning::'$*' failed (attempt $attempt of 3); retrying in 15s"
+              sleep 15
+            done
+            return 1
+          }
           sudo apt update
-          sudo snap install charmcraft --classic
-          sudo snap install go --classic
+          retry sudo snap install charmcraft --classic
+          retry sudo snap install go --classic
           go install github.com/canonical/spread/cmd/spread@latest
           pipx install tox poetry
           # to build Valkey-glide during tests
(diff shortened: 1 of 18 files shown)


.github/workflows/integration_test.yaml
```diff
@@ -196,6 +217,42 @@ jobs:
         timeout-minutes: 3
         if: ${{ success() || (failure() && steps.spread.outcome == 'failure') }}
         run: script -q -c 'sudo jhack tail --printer raw --replay --no-watch' /dev/null | tee ~/logs/jhack-tail.txt || true
+      - name: Capture Valkey server logs
```
skourta: Thank you for this

tests/integration/ha/helpers/helpers.py
```diff
@@ -55,21 +65,29 @@ def lxd_cut_network_from_unit_with_ip_change(machine_name: str) -> None:
...
-    override_command = f"lxc config device override {machine_name} eth0"
-    try:
-        subprocess.check_call(override_command.split())
-    except subprocess.CalledProcessError:
-        # Ignore if the interface was already overridden.
-        pass
+    for rule in NETWORK_CUT_RULES:
+        # drop a leftover copy first so re-cutting cannot stack a duplicate rule
+        _lxd_delete_iptables_rule(machine_name, rule)
+        subprocess.check_call(f"lxc exec {machine_name} -- iptables --insert {rule}".split())
```
skourta: @Mehdi-Bendriss this diverges from the other products. Should we update the other products if this is stricter and more stable?

spread.yaml
```diff
@@ -125,8 +125,53 @@ prepare: |
 
   if [[ $SPREAD_SUITE == *"k8s"* ]]
   then
+    # Canonical K8s defaults its pod CIDR to 10.1.0.0/16 — the same range the GitHub-hosted
+    # Azure runners use for their host network, so a pod handed the subnet gateway address
+    # cannot reach the apiserver and its first hook dies. concierge has no pod-cidr option,
+    # but it adopts an already-clustered node, so bootstrap here first with a safe CIDR.
+    POD_CIDR="10.42.0.0/16"
```
skourta: we can specify the cider via concierge
https://github.com/canonical/concierge/blob/main/presets/k8s.yaml#L21

src/core/models.py
```diff
@@ -150,6 +150,7 @@ class PeerUnitModel(PeerModel):
     client_cert_ready: bool = Field(default=False)
     tls_ca_rotation: str = Field(default="")
     tls_certificate_expiring: bool = Field(default=False)
+    applied_client_cert_fingerprint: str = Field(default="")
```
skourta: why not read from the disk instead of adding a databag field?

skourta (commented):
Amazing work on this @delgod! Thank you.
Minor comments!

tests/integration/ha/helpers/helpers.py
delgod: I am open to other ways that would actually work for ValKey.
The shared DP helper is an LXD throttle (egress=0kbit/ingress=1kbit), not a cut: 1 kbit ≈ 125 B/s still carries Sentinel PING/PONG.
The current code might be good enough for other products because they have heavy transaction traffic between members, not a light quorum ping/pong.

spread.yaml
delgod: `load-balancer.cidrs` (Service LB IP pool), not pod CIDR

src/core/models.py
delgod: `_on_certificate_available`  writes the cert to disk before `reload_tls_settings` + `restart_workload.emit`.
 The `ValkeyCertificatesNotReadyError` branch returns without defer and relies  on the next re-emission to finish; a disk comparison would then say "already there → skip" and the reload/restart would never run.

tests/integration/ha/helpers/helpers.py
Mehdi-Bendriss: @skourta not a priority, but yes. Our throttle was a best effort network cut (without IP change)

.github/workflows/integration_test.yaml
```diff
@@ -196,6 +217,42 @@ jobs:
...
+            done
+          else
+            log_dir=/var/snap/charmed-valkey/common/var/log/valkey
+            for unit in $units; do
+              name=${unit/\//-}
+              { sudo juju ssh "$unit" "sudo cat $log_dir/valkey.log"; } \
+                > ~/logs/"$name"-valkey-server.txt 2>&1
+              { sudo juju ssh "$unit" "sudo cat $log_dir/sentinel.log"; } \
+                > ~/logs/"$name"-valkey-sentinel.txt 2>&1
+              { sudo juju ssh "$unit" 'sudo snap logs charmed-valkey.sentinel -n 500'; } \
```
reneradoi: I think this should be `charmed-valkey.server` if we want to capture the Valkey service logs.

src/events/base_events.py
```diff
@@ -376,39 +381,63 @@ def _on_leader_elected(self, event: ops.LeaderElectedEvent) -> None:
...
+
+        Returns:
+            bool: True if the address is reconciled (or there was nothing to do), False if
+                the reconcile could not complete and must be retried.
+        """
+        # on k8s we use hostnames, so the address is irrelevant and the binding unread
+        if self.charm.state.substrate != Substrate.VM:
+            return True
+
+        if not (
```
reneradoi: nit: I'd like to invert for better readability:
```
if (
model.private_ip
and bind_address == model.private_ip
):
    return True
```

A positive statement is easier to understand than a double negative.

src/managers/tls.py
```diff
@@ -73,6 +74,38 @@ def set_ca_rotation_state(self, state: TLSCARotationState) -> None:
...
+        The fingerprint is recorded only once an apply has been handed off
+        (restart emitted, or CA rotation started — the rotation state machine
+        owns activation from there), so a deferred/failed apply is still
+        retried instead of being skipped.
+        """
+        if self.state.unit_server.tls_client_state != TLSState.TLS:
+            return False
+        if self.state.unit_server.tls_ca_rotation_state != TLSCARotationState.NO_ROTATION:
+            return False
+        if not self.state.unit_server.model.client_cert_ready:
```
reneradoi: It could happen that a certificate-available arrives before the peer relation is fully set up, so we need to safeguard:
```
if not self.state.unit_server.model:
    return False
```

Previously, the logic was robust against this and would just log a message that the peer relation is not ready yet. Now it could crash.

reneradoi (commented):
Very nice findings, thank you Mykola for tracking the failures down to the root causes! I only have minor change requests, other than that I'm fine with the PR.

.github/workflows/integration_test.yaml
delgod: Fixed in b96ac57 — now snap logs charmed-valkey, which covers server and sentinel. Both services' logfiles are already captured above; this file is the service-manager view (VM counterpart of pebble changes).

tests/integration/tls/test_certificate_rotation.py
```diff
@@ -199,11 +210,19 @@ def test_ca_rotation_by_config_change(juju: jubilant.Juju) -> None:
     assert old_certificate, "Failed to get current certificate"
 
     logger.info("Rotating the CA certificate")
-    tls_config = {"certificate-validity": "10d", "ca-common-name": "new-valkey-ca"}
+    # The short CA validity schedules the provider-side CA renewal test_ca_rotation_by_expiration
```
skourta: nit: This couples the tests together. We can no longer run them independently.

skourta (approved):
Small comment not critical. I'm ok to merge after CI runs green (maybe run it a couple of times?).
Can you also please merge the main branch again.

src/events/base_events.py
delgod: Done in e63d40c, but as two early returns rather than one positive term — the literal inversion isn't equivalent. Empty private_ip (nothing recorded until start / leader-elected) would fall through into the reconcile and reconfigure/restart a unit that hasn't started. New test test_update_status_does_not_reconcile_before_an_address_is_recorded fails against the inversion, passes here.

src/managers/tls.py
delgod: Can't crash: the function opens on tls_client_state, which returns NO_TLS when model is unbuilt (core/models.py:260-263) and returns False before line 97. The "not ready" log is intact too — record_applied_certificate goes through RelationState.update() (core/models.py:182-187).
Pinned it with a test instead of a redundant guard (f87c93b). Say the word if you'd still like the explicit if not ...model: return False.

tests/integration/tls/test_certificate_rotation.py
delgod: Fixed the hidden part of it: the module-level _ca_rotation_config_time global is replaced by a module-scoped rotated_ca fixture. [0ffe2c3](https://github.com/canonical/valkey-operator/pull/95/commits/0ffe2c3d698ce19600113105f43c76a9e4c87249)

I don't want to make expiry test fully standalone because of +15–18 min tax.

src/managers/tls.py
reneradoi: You are right, this will never happen because we would already return earlier in such a case. We can keep it as it is.

reneradoi (approved):
Thank you Mykola!

PR #96, opened 2026-08-18
[DPE-10650] S3 restore followup

delgod:
Follow-up to #79 (S3 restore) addressing the last review round (skourta, 2026-08-17).
Only the four commits on top of #79 are new; base becomes `9/edge` once #79 lands.

- **fix(restore): self-heal a sentinel left failover-suppressed** — `resume_failover` is
  best-effort on failure paths and Sentinel persists `SENTINEL SET` to its own conf, so a
  sentinel unreachable at that moment came back with the 24 h `down-after` and no automatic
  failover until the next config re-render. Each unit now re-checks its own sentinel on
  `update-status` (outside a restore) and resets it.
- **refactor(restore): log the restore trail** — `restore.*` log lines for every action
  rejection, step, barrier advance, completion, download, rollback and failure
  (`juju debug-log | grep restore.` shows the whole run); the failure helpers are renamed
  to say what they do (`_fail_restore` / `_finish_failed_restore`); comment/nit cleanups.
- **refactor(cluster): fuse the bounded waits** — `_wait_until(condition, timeout, failure)`
  replaces the duplicated tenacity loops in `wait_until_loaded` / `wait_until_resynced`.
- **test(restore): single-unit happy path** — `test_restore_single_unit` (1-unit app:
  leader == primary, no resync, cascade driven only by the leader's own app-databag
  relation-changed); also asserts the unit is writable after the restart.

No databag/schema changes; behaviour change is limited to the sentinel self-heal.

### DIFF
--- a/src/events/backup.py
+++ b/src/events/backup.py
@@ -86,7 +86,6 @@ class BackupEvents(ops.Object):
             self.charm.on[PEER_RELATION].relation_departed, self._on_restore_workflow
         )
         self.framework.observe(self.charm.on.update_status, self._on_restore_workflow)
-        self.framework.observe(self.charm.on.update_status, self._reconcile_failover_suppression)
 
     # ── event handlers ──────────────────────────────────────────────────
 
@@ -335,9 +334,13 @@ class BackupEvents(ops.Object):
         Juju's one self-delivery guarantee. No in-hook loop; every guard below is
         idempotent, so a redelivered hook re-runs and converges.
         """
-        # Restore done: each unit clears its own per-unit state once restore_id is gone.
+        # Restore done: each unit clears its own per-unit state once restore_id is gone,
+        # and self-heals a sentinel a failed restore left failover-suppressed --
+        # suppression is only legitimate while a restore runs.
         if not self.charm.state.cluster.is_restore_in_progress:
             self._clear_local_restore_state()
+            if self.charm.state.unit_server.is_active:
+                self.charm.sentinel_manager.reconcile_failover_suppression()
             return
 
         # Leader ends a failed restore on a participant failure or departure
@@ -378,19 +381,6 @@ class BackupEvents(ops.Object):
 
         if self.charm.unit.is_leader():
             self._advance_if_leader()
-
-    def _reconcile_failover_suppression(self, _: ops.UpdateStatusEvent) -> None:
-        """Self-heal this unit's sentinel if a failed restore left failover suppressed.
-
-        Suppression is only legitimate while a restore runs, so this unit's
-        sentinel is re-checked on every update-status outside one.
-        """
-        if (
-            not self.charm.state.unit_server.is_active
-            or self.charm.state.cluster.is_restore_in_progress
-        ):
-            return
-        self.charm.sentinel_manager.reconcile_failover_suppression()
 
     def _clear_local_restore_state(self) -> None:
         """Clear this unit's per-unit restore fields once a restore has ended."""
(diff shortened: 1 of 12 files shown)


src/events/backup.py
```diff
@@ -275,29 +276,29 @@ def _unstable_primary_reason(self) -> str | None:
...
+        backup_id = event.params.get("backup-id", "")
         if reason := self._restore_blocking_reason():
-            event.set_results({"error": reason})
-            event.fail(reason)
+            self._reject_restore(event, backup_id, reason)
             return
-        if not (backup_id := event.params.get("backup-id", "")):
-            event.set_results({"error": "Must provide backup-id to restore."})
-            event.fail("Must provide backup-id to restore.")
+        if not backup_id:
```
skourta: Consider adding this in the _restore_blocking_reason. This and the other checks

src/events/backup.py
```diff
@@ -362,34 +370,62 @@ def _on_restore_workflow(
...
+            logger.exception(
+                "restore.step_failed instruction=%s step=%s role=%s", instruction, step, role
+            )
+            self._fail_restore(e)
             return
 
         if self.charm.unit.is_leader():
             self._advance_if_leader()
 
+    def _reconcile_failover_suppression(self, _: ops.UpdateStatusEvent) -> None:
```
skourta: This can be moved to the sentinel manager.

skourta (commented):
Thank you @delgod only two comments so far.
The PR is still marked as draft so when it is ready for review I will recheck it.

src/events/backup.py
delgod: Done. _restore_blocking_reason now takes backup_id and owns all the gates. 00e0b3d.

src/events/backup.py
delgod: Moved: SentinelManager.reconcile_failover_suppression() now does the check 62676ab.

reneradoi (commented):
No concerns in general, I only have a design remark for the added logging: All the log messages are bloating the event flow, in some cases we even add private methods only for avoiding duplicating log messages. 

Example: `BackupEvents._record_restore_step()` consists of only two lines: the log line and a call to the manager function. It makes to code much harder to read and I would prefer if we could be more concise. Could you please try to add the log output to the manager methods where possible, and leave them out of the event handler? This can be supported by not logging the unit name as part of the log message, which is also duplicate information (already logged by Juju for each line).

If this could be applied all over the events, it would increase readability a lot.

src/events/backup.py
```diff
@@ -86,6 +86,7 @@ def __init__(self, charm: "ValkeyCharm"):
             self.charm.on[PEER_RELATION].relation_departed, self._on_restore_workflow
         )
         self.framework.observe(self.charm.on.update_status, self._on_restore_workflow)
+        self.framework.observe(self.charm.on.update_status, self._reconcile_failover_suppression)
```
skourta: nit: can't we fuse these two event handlers? `_reconcile_failover_suppression` is just a proxy for its manager method with a condition

skourta (approved):
One nit comment.
Thank you @delgod 
Can you please check the unit tests and rebase?

src/events/backup.py
delgod: Fused — the self-heal moved into _on_restore_workflow's "no restore in progress" branch, which already holds that condition. fef8aae.

skourta (approved):

reneradoi (approved):
Thank you Mykola!

PR #98, opened 2026-08-20
build(deps): use released valkey-glide instead of the git fork

delgod:
Swaps the `valkey-glide` git fork for the released PyPI package — `^2.5` in both pyprojects (locks resolve 2.5.1).

The fork existed for upstream PR 5124 (mTLS), which merged Jan 2026 and has shipped since 2.4.0. It costs **272 s of Rust build on every spread job**, ~34 jobs per CI run.

**`poetry install --only integration`: 272 s → 0.6 s.**

### Two gotchas, both handled here

**The sdist is incomplete.** Wheels ship `glide` *and* `glide_shared`; every sdist from 2.2.10 to 2.5.1 ships only `glide` — which imports `glide_shared`. charmcraft installs with `--no-binary=:all:`, so glide-runner failed every hook with `ModuleNotFoundError: No module named 'glide_shared'`. Fixed by pinning that one package to its wheel: `poetry-pip-extra-args: ["--only-binary=valkey-glide"]`.

**glide renames error codes.** Valkey replies `NOPERM`, glide ≥2.4 renders `PermissionDenied`. Two assertions in `test_client_relation.py` now accept either. `NOAUTH`/`WRONGPASS` still arrive verbatim via `ClosingError`, so `helpers.py` is untouched.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -221,12 +221,13 @@ Any code touching addresses, file paths, services, or networking must handle bot
   tests are still a TODO. Channel is `9/edge`.
 - Always invoke a specific env (`tox run -e <env>`); bare `tox` errors on an undefined `static` env
   reference (legacy `env_list` entry).
-- `valkey-glide` is a git dependency on a fork (pending upstream PR 5124): `poetry lock` (run by
-  `format`) needs network and may bump the fork commit. Building the integration group locally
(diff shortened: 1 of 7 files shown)


Mehdi-Bendriss (approved):
Thanks Mykola! This was a long awaited fix. 
@skourta missed your window there :D

skourta (approved):

reneradoi (approved):
Very nice, this was still on the to-do list.

PR #99, opened 2026-08-20
docs: Update LDAP guide for VM/K8s

izmalk:
This is the first of many. To serve as a good exemplar of how to utilise tabs for VM/K8s docs unification.

I initially planned to use deploy guide for this purpose, but the guide is too simple and all the commands are the same. Ended up selecting the LDAP guide. At least we have cross-model for VM here.

- [x] Have you updated relevant documentation?


reneradoi (approved):
Very nice, thank you Vladimir!

skourta (approved):

Issue #88, opened 2026-07-17
Grant client|id, client|tracking and ping to valkey_client relation users


f-atwi:
The ACL generated for valkey_client relation users (_get_client_user_acl_lines in src/managers/config.py) grants only data categories 
(-@all +@read +@write +@keyspace +@pubsub +@transaction +info) 

The workload for the HRMS charm fails with NoPermissionError on CLIENT ID, CLIENT TRACKING,  and PING which prevents the app from starting.
We verified that granting client|id, client|tracking, and ping resolves it.

Could these be added to the relation user's permissions?

reneradoi:
Thank you again for reporting! As discussed, we will add these permissions in near future and I will update you when there is news.

reneradoi:
Hey @f-atwi, support for these permissions has been added and released to `9/edge`.

f-atwi:
> Hey [@f-atwi](https://github.com/f-atwi), support for these permissions has been added and released to `9/edge`.

Thank you so much!!

Issue #82, opened 2026-07-06
Scripting permissions


hk21702:
Please grant scripting permissions for users or allow the configuration of permissions.

https://github.com/canonical/valkey-operator/blob/95c61c40cd355e4137f513ce1a9edd28c8a980e0/src/managers/config.py#L243

We have a use case that requires the use of operations such as `script load`, `eval`, and `evalsha`.

https://github.com/wikimedia/mediawiki-services-jobrunner/blob/fe525ace2c9924d71a02a548a5a542ea9ff3e4fd/src/RedisJobChronService.php#L217
https://doc.wikimedia.org/mediawiki-core/REL1_46/php/RedisConnRef_8php_source.html

reneradoi:
Thank you again for reporting! As discussed, we will think about how to safely enable scripting permissions and allocate time for the implementation. I will update you when there is news.

reneradoi:
Hey @alexdlukens-canonical, we plan to allow scripting permissions by default to client relation users. This will be added soon, I will reply here once merged.

reneradoi:
Hi @hk21702 and @alexdlukens-canonical, this feature has been added and released to `9/edge`.

alexdlukens-canonical:
@reneradoi Thanks for the update. I confirm that rev73 works for my usecase, without my manual workarounds for permissions. Cheers!

PR #101, opened 2026-08-21
[DPE-10892] Add scripting and caching permissions

reneradoi:
This PR addresses:
- #82
- #88 

The permissions for external client users are adjusted to allow client-side caching and LUA scripting. To safeguard memory consumption for client-side caching, the configuration `tracking-table-max-keys` is explicitly set to 1M (same as default).

Regarding scripting: The scripting permissions do not impact the keyspace permissions, or in other words: Scripts can only be executed on a key if the user has permissions for it. This was confirmed by testing:
```
$ juju deploy data-integrator
$ juju deploy ./valkey_ubuntu@24.04-amd64.charm --trust --resource valkey-image=ghcr.io/canonical/charmed-valkey:9.0.1-26.04_edge
$ juju config data-integrator prefix-name="test:*"
$ juju integrate data-integrator valkey                                                                             
$ valkey-cli -h 10.1.0.12 -p 6379
10.1.0.12:6379> auth relation-5-ee6157ff1cd25188 <password>
OK                                                                
10.1.0.12:6379> set mykey 42
(error) NOPERM No permissions to access a key
10.1.0.12:6379> set test:mykey 42
OK
10.1.0.12:6379> get mykey
(error) NOPERM No permissions to access a key
10.1.0.12:6379> get test:mykey
"42"
10.1.0.12:6379> EVAL "return server.call('SET', KEYS[1], ARGV[1])" 1 foo bar
(error) NOPERM No permissions to access a key
10.1.0.12:6379> EVAL "return server.call('SET', KEYS[1], ARGV[1])" 1 test:foo bar
OK
10.1.0.12:6379> EVAL "return server.call('GET', KEYS[1])" 1 foo
(error) NOPERM No permissions to access a key
10.1.0.12:6379> EVAL "return server.call('GET', KEYS[1])" 1 test:foo
"bar"
```

Integration test failures for storage reuse on VM are addressed in #102

### DIFF
--- a/src/managers/config.py
+++ b/src/managers/config.py
@@ -119,6 +119,9 @@ class ConfigManager(ManagerStatusProtocol):
         ldap_config = self.generate_ldap_config()
         config_properties.update(ldap_config)
 
+        # Client-side caching: limit memory consumption for key-invalidation table
+        config_properties["tracking-table-max-keys"] = "1000000"
+
         return config_properties
 
     def _generate_replica_config(self, primary_endpoint: str) -> dict[str, str]:
--- a/src/managers/auth.py
+++ b/src/managers/auth.py
@@ -122,6 +122,11 @@ class AuthManager(ManagerStatusProtocol):
         Returns:
             str: ACL lines for the external client users.
         """
+        valkey_base_permissions = (
+            "-@all +@read +@write +@keyspace +@pubsub +@transaction +info +ping +role "
+        )
+        valkey_scripting_permissions = ""
+        valkey_client_caching_permissions = "+client|id +client|tracking "
         sentinel_base_permissions = "-@all +auth +client +command +hello +ping +role "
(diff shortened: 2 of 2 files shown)


src/managers/config.py
```diff
@@ -119,6 +119,9 @@ def get_config_properties(self, primary_endpoint: str) -> dict[str, str]:
         ldap_config = self.generate_ldap_config()
         config_properties.update(ldap_config)
 
+        # Client-side caching: limit memory consumption for key-invalidation table
+        config_properties["tracking-table-max-keys"] = "1000000"
```
skourta: Isn't this the default value?

skourta (commented):
One small comment otherwise looks good.
Thank you @reneradoi

src/managers/config.py
reneradoi: Yes that is correct. It's a safeguard to make this explicit and avoid issues if the default should change or go away upstream.

Mehdi-Bendriss (approved):
Thanks René!

skourta (approved):
Thank you @reneradoi

PR #102, opened 2026-08-24
[DPE-10894] Improve scale-down procedure

reneradoi:
This PR attempts to improve the stability of scaling down / removing the application.

As seen in multiple CI runs ([example 1](https://github.com/canonical/valkey-operator/actions/runs/32492483752/job/97336924041?pr=101#step:13:28025), [example 2](https://github.com/canonical/valkey-operator/actions/runs/32610038957/job/97337902711#step:13:28909)), the removal of the entire application may take unusually long and never complete. This needs improvement to avoid users not being able to remove a Valkey cluster, or being tempted to run removal with `--force`.

The cause for this issue is that if the connection to the other units is not possible, an exception is raised from the charm because of Sentinel health check errors. This exception occurs _after_ the workload on this unit has already been stopped. If the event handler returns, earlier checks/actions on the `storage_detaching` event are bound to fail, resulting in a never-healing error.

The proposed solution is to decouple unit removal and Sentinel removal. The goal is to not have any operation _after_ the actual workload stop that could raise. Once stopped, the `storage_detaching` hook should not fail anymore.  

The proposed design:
- the responsibility for Sentinel removal/reset and health check is now at the Sentinel Manager, the `storage_detaching` hook of the departing unit does not care about Sentinel state anymore
- the leader unit sets a flag `sentinel_reset_required` to the application databag on `peer-relation-departed`
- the leader unit checks on `peer-relation-changed` if one or more Sentinels need to be removed from the cluster and executes the removal and the health checks on all remaining Sentinels
- the Sentinel removal must only happen after the Sentinel service on the departing unit was stopped
- multiple departing Sentinels are collected and only once all are stopped, the removal is executed
- the workflow is save for Juju leadership changes during removal (ensured through `leader-elected`)
- if the removal of departed Sentinels was not successful, a status message is displayed

### DIFF
--- a/src/events/base_events.py
+++ b/src/events/base_events.py
@@ -708,7 +708,7 @@ class BaseEvents(ops.Object):
             if ip != self.charm.state.unit_server.get_endpoint(self.charm.state.substrate)
         ]
 
-        if active_sentinels:
+        if active_sentinels and self.charm.app.planned_units() != 0:
             logger.debug("Resetting sentinel states on active units: %s", active_sentinels)
             self.charm.sentinel_manager.reset_sentinel_states(active_sentinels)
 
--- a/src/events/base_events.py
+++ b/src/events/base_events.py
@@ -11,7 +11,10 @@ import ops
 
 from common.custom_events import UnitFullyStartedEvent
 from common.exceptions import (
+    CannotSeeAllActiveSentinelsError,
+    NotAllDepartingSentinelsStoppedError,
     RequestingLockTimedOutError,
+    SentinelIncorrectReplicaCountError,
     ValkeyACLLoadError,
     ValkeyBackupInProgressError,
     ValkeyCannotGetPrimaryIPError,
@@ -255,7 +258,7 @@ class BaseEvents(ops.Object):
         except (ValkeyWorkloadCommandError, ValueError) as e:
             logger.error("Failed to start topology observer: %s", e)
 
-    def _on_peer_relation_changed(self, _: ops.RelationChangedEvent) -> None:
+    def _on_peer_relation_changed(self, event: ops.RelationChangedEvent) -> None:
         """Handle event received by all units when a unit's relation data changes."""
         if self.charm.state.cluster.is_restore_in_progress:
             return
@@ -277,6 +280,19 @@ class BaseEvents(ops.Object):
             lock.process()
 
         if not self.charm.state.unit_server.is_active:
+            return
+
+        try:
+            self.charm.sentinel_manager.remove_departed_sentinels_from_cluster()
+        except (
+            CannotSeeAllActiveSentinelsError,
+            NotAllDepartingSentinelsStoppedError,
+            SentinelIncorrectReplicaCountError,
+            ValkeyCannotGetPrimaryIPError,
+            ValkeyWorkloadCommandError,
+        ) as e:
+            logger.error(e)
+            event.defer()
             return
 
         # return early during TLS switchover to avoid unnecessary operation during rolling restart for sentinel
@@ -309,7 +325,13 @@ class BaseEvents(ops.Object):
         if self.charm.state.unit_server.is_started:
             self.charm.cluster_manager.reconcile_min_replicas_to_write()
 
-        if not self.charm.unit.is_leader() or not self.charm.state.unit_server.is_active:
+        if not self.charm.unit.is_leader():
+            return
+
+        # trigger Sentinel reset and health check, executed on relation-changed
+        self.charm.state.cluster.update({"sentinel_reset_required": True})
+
+        if not self.charm.state.unit_server.is_active:
             return
 
         try:
@@ -354,6 +376,19 @@ class BaseEvents(ops.Object):
             return
 
         if self.charm.state.unit_server.is_active:
+            try:
+                self.charm.sentinel_manager.remove_departed_sentinels_from_cluster()
+            except (
+                CannotSeeAllActiveSentinelsError,
+                NotAllDepartingSentinelsStoppedError,
+                SentinelIncorrectReplicaCountError,
+                ValkeyCannotGetPrimaryIPError,
+                ValkeyWorkloadCommandError,
+            ) as e:
+                logger.error(e)
+                event.defer()
+                return
+
             try:
                 self.charm.topology_manager.start_observer()
             except (ValkeyWorkloadCommandError, ValueError) as e:
@@ -654,7 +689,6 @@ class BaseEvents(ops.Object):
             scope="unit",
             component=self.charm.cluster_manager.name,
         )
-        # TODO consider quorum when removing unit
 
         self.charm.status.set_running_status(
             ScaleDownStatuses.SCALING_DOWN.value,
@@ -662,6 +696,7 @@ class BaseEvents(ops.Object):
             component_name=self.charm.cluster_manager.name,
             statuses_state=self.charm.state.statuses,
         )
+
         # if unit has primary then failover
         try:
             primary_ip = self.charm.sentinel_manager.get_primary_ip_for_scale_down()
@@ -691,9 +726,8 @@ class BaseEvents(ops.Object):
             logger.info("Waiting for replica to be fully-synced before saving the dataset")
             self.charm.cluster_manager.wait_for_replica_fully_synced(primary_ip)
 
+        # stop valkey
         self.charm.cluster_manager.save_dataset_before_shutdown()
-
-        # stop valkey and sentinel processes
         try:
             self.charm.workload.stop()
         except ValkeyServicesCouldNotBeStoppedError as e:
@@ -701,21 +735,7 @@ class BaseEvents(ops.Object):
 
         self.charm.state.unit_server.update({"start_state": StartState.NOT_STARTED.value})
 
-        # reset sentinel states on other units
-        active_sentinels = [
-            ip
-            for ip in active_sentinels
-            if ip != self.charm.state.unit_server.get_endpoint(self.charm.state.substrate)
-        ]
-
-        if active_sentinels and self.charm.app.planned_units() != 0:
-            logger.debug("Resetting sentinel states on active units: %s", active_sentinels)
-            self.charm.sentinel_manager.reset_sentinel_states(active_sentinels)
-
-            # check health after scale down
-            self.charm.sentinel_manager.verify_expected_replica_count(active_sentinels)
-
-        if active_sentinels:
+        if len(active_sentinels) > 1:
             scale_down_lock.release_lock(primary_ip=primary_ip)
 
         self._set_state_for_going_away()
--- a/src/events/base_events.py
+++ b/src/events/base_events.py
@@ -714,7 +714,8 @@ class BaseEvents(ops.Object):
 
             # check health after scale down
             self.charm.sentinel_manager.verify_expected_replica_count(active_sentinels)
-            # release lock
+
+        if active_sentinels:
             scale_down_lock.release_lock(primary_ip=primary_ip)
 
         self._set_state_for_going_away()
--- a/src/managers/sentinel.py
+++ b/src/managers/sentinel.py
@@ -14,7 +14,9 @@ from tenacity import retry, retry_if_result, stop_after_attempt, wait_fixed
 
 from common.client import SentinelClient
 from common.exceptions import (
+    CannotSeeAllActiveSentinelsError,
     SentinelFailoverError,
+    SentinelIncorrectReplicaCountError,
     ValkeyCannotGetPrimaryIPError,
     ValkeyWorkloadCommandError,
 )
@@ -236,6 +238,9 @@ class SentinelManager(ManagerStatusProtocol):
                 logger.warning(
                     "Sentinel at %s does not see all other sentinels after reset.", sentinel_ip
                 )
+                raise CannotSeeAllActiveSentinelsError(
+                    "Sentinel at %s does not see all other sentinels after reset." % sentinel_ip
+                )
 
     @retry(
         wait=wait_fixed(5),
@@ -315,6 +320,12 @@ class SentinelManager(ManagerStatusProtocol):
                 number_replicas := len(client.replicas_primary(hostname=sentinel_ip))
             ):
                 logger.warning(
+                    "Sentinel at %s sees %d replicas, expected %d.",
+                    sentinel_ip,
+                    number_replicas,
+                    expected_replicas,
+                )
+                raise SentinelIncorrectReplicaCountError(
                     "Sentinel at %s sees %d replicas, expected %d.",
                     sentinel_ip,
                     number_replicas,
--- a/src/managers/sentinel.py
+++ b/src/managers/sentinel.py
@@ -15,6 +15,7 @@ from tenacity import retry, retry_if_result, stop_after_attempt, wait_fixed
 from common.client import SentinelClient
 from common.exceptions import (
     CannotSeeAllActiveSentinelsError,
+    NotAllDepartingSentinelsStoppedError,
     SentinelFailoverError,
     SentinelIncorrectReplicaCountError,
     ValkeyCannotGetPrimaryIPError,
@@ -35,7 +36,7 @@ from literals import (
     K8sService,
     Substrate,
 )
-from statuses import CharmStatuses
+from statuses import CharmStatuses, ScaleDownStatuses
 
 logger = logging.getLogger(__name__)
 
@@ -202,6 +203,11 @@ class SentinelManager(ManagerStatusProtocol):
 
         return True
 
+    def is_sentinel_active(self, sentinel_ip: str) -> bool:
+        """Check if the sentinel is active and responds to a ping."""
+        client = self._get_sentinel_client()
+        return client.ping(hostname=sentinel_ip)
+
     def failover(self) -> None:
         """Trigger a failover in the cluster.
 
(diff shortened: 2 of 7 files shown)


skourta (commented):
Shouldn't we gate this if planned_units is 0?

src/managers/sentinel.py
```diff
@@ -238,9 +236,6 @@ def reset_sentinel_states(self, sentinel_ips: list[str]) -> None:
                 logger.warning(
                     "Sentinel at %s does not see all other sentinels after reset.", sentinel_ip
                 )
-                raise CannotSeeAllActiveSentinelsError(
-                    "Sentinel at %s does not see all other sentinels after reset." % sentinel_ip
-                )
```
Mehdi-Bendriss: I followed the code path here, and it seems we are sure this exception will get raised. In which case, the reset_sentinel_states will never get applied by the departing unit (which is the only one ever attempting to reset the state), which also only invokes this if there are other sentinels.
We should most likely keep this exception, and have this run by another sentinel on peer_departing or similar instead.

src/managers/sentinel.py
```diff
@@ -325,12 +320,6 @@ def verify_expected_replica_count(self, sentinel_ips: list[str]) -> None:
                     number_replicas,
                     expected_replicas,
                 )
-                raise SentinelIncorrectReplicaCountError(
```
Mehdi-Bendriss: same comment

Mehdi-Bendriss (commented):
Thanks René. I left a couple of comments

src/managers/sentinel.py
reneradoi: I agree with you, we should keep raising. And I also agree with Smail, we could gate this for planned units != 0. If no unit remains, it does not make sense to reset the sentinel state, because all sentinels will be gone.

I also investigated your idea about moving responsibility for updating sentinel state to the leader unit, but that would be much more complex (to safely determine whether a sentinel has already been stopped on a departing unit). There is no real need to move it to the leader, so I decided to discard and keep it simple.

src/managers/sentinel.py
reneradoi: see above.

src/managers/sentinel.py
Mehdi-Bendriss: Hmm I see 2 ways to orchestrate this from the leader (or another) unit, in peer_relation_departing:
1. [sync] from any unit: 
    => while departing_unit in list_of_sentinels => sleep => then reset the state
    - which would be a bit risky given we don't know how long can storage detaching run   
2. [async] in leader unit: 
    => if departing_unit in list_of_sentinels => put `reset_state_required=True` in the peer rel data => defer => when coming back to the hook check the flag if there => reset => remove the flag 

I prefer the 2nd

reneradoi: I have investigated and tested this, and then moved away from it. The issue here is: How to reliably detect `if departing_unit in list_of_sentinels`?
- take list of sentinels from relation state (`state.servers`): will fail because the departing unit is not in the peer relation anymore, but the sentinel on the departing unit is still running and failover did not happen yet
- take list of sentinels from ops (`model.app.get_unit`): same as above
- take list of sentinels from sentinel: will fail because, until sentinel was reset, the departing unit will always be there (chicken egg problem)

The only way I can think of is checking the number of planned units against the number of sentinels from sentinel itself, but this is pretty risky and I would not consider it a reliable solution.

That's why I reverted back to gating the sentinel reset and not run it if planned units == 0. Please let me know if you have other ideas.

src/managers/sentinel.py
```diff
@@ -219,6 +225,34 @@ def failover(self) -> None:
...
+            SentinelIncorrectReplicaCountError: If any sentinel sees an incorrect number of replicas
+        """
+        if not self.state.cluster.model.sentinel_reset_required:
+            return
+
+        logger.info("Resetting Sentinels after unit removal")
+        primary_ip = self.get_primary_ip()
+        active_sentinels = self.get_active_sentinel_ips(primary_ip)
+
+        if len(active_sentinels) != self.state.model.app.planned_units():
```
skourta: How would this impact scale up? We should perhaps change the condition for it to only apply to scale downs.

src/events/base_events.py
```diff
@@ -691,30 +726,16 @@ def _scale_down_unit(self) -> None:
...
-        # reset sentinel states on other units
-        active_sentinels = [
-            ip
-            for ip in active_sentinels
-            if ip != self.charm.state.unit_server.get_endpoint(self.charm.state.substrate)
-        ]
-
-        if active_sentinels:
-            logger.debug("Resetting sentinel states on active units: %s", active_sentinels)
-            self.charm.sentinel_manager.reset_sentinel_states(active_sentinels)
```
skourta: Removing these would also diverge from the [docs ](https://redis.io/docs/latest/operate/oss_and_stack/management/sentinel/#adding-or-removing-sentinels)

skourta (commented):
Thank you @reneradoi.
I left some comments. Prefer to discuss in a call.

src/managers/sentinel.py
reneradoi: Through `sentinel_reset_required`, it is gated to only happen when a unit was removed (the flag is set by the leader on `relation_departed`).

If scale up happens during scale down, there could be these scenarios:
a) new unit Sentinel already started, departing unit Sentinel not down yet
- `active_sentinels` would report planned_units + 1
-  no reset happens

b) new unit Sentinel already started, departing unit Sentinel down
- `active_sentinels` would report == planned_units
- reset happens, as new Sentinel is already up, no impact other than removing the departing one

c) new unit Sentinel not yet started, departing unit Sentinel down
- `active_sentinels` would report planned_units - 1
- no reset happens until new unit Sentinel is started

d) new unit Sentinel not yet started, departing unit Sentinel not down yet
- `active_sentinels` would report == planned_units
- reset happens, departing Sentinel gets removed
- once new unit Sentinel is started, it will be added to the cluster by the charm.

If I miss something, please let me know, otherwise this is save against scale-up and scale-down in parallel.

src/events/base_events.py
reneradoi: We don't remove them, they are still there. It's just the responsibility for coordinating the workflow that moves from the departing unit to the Juju leader unit.

skourta (commented):
The issue has two parts to it. An issue with how we aquire the lock and the other with how we clean sentinel state after a unit has been stopped. Both issues are triggered by a scale-down of a multi-unit application, where the units are stopped at the same time.

The test times out because `_scale_down_unit` performs cross-unit Sentinel cleanup after `workload.stop()`. When two units are removed at the same time, the lock holder (valkey/5) issues SENTINEL RESET against the survivor and cleans its SENTINEL SENTINELS view. The other departing unit (valkey/6) reads `len(get_active_sentinel_ips()) == 1`, thinks it is the "last unit" and skips the lock tearing down in parallel. valkey/5's `reset_sentinel_states` then asserts against a an old list of active sentinels which contains valkey/6, so `target_sees_all_others` can never succeed and it raises `CannotSeeAllActiveSentinelsError`. Because juju only writes databag when the hook succeeds, the peer databag writes are never written. When the retried storage-detaching reenters from the top it dies at `save_dataset_before_shutdown`. The abandoned cleanup leaves dead sentinels in the units that are not scaled down which might even cause issues for HA. 

Timeline from logs of one of the runs:
- 06:30:58 — valkey/5 asks for the scale-down lock.
- 06:30:59 — valkey/5 gets the lock.
- 06:31:01 — valkey/6 asks for the same lock, is refused, and its hook fails. This is normal; it will try again.
- 06:31:01 — valkey/5 saves its data, then stops its own Valkey and Sentinel.
- 06:31:09 — valkey/5 sends SENTINEL RESET to the survivor (10.127.23.222). This makes the survivor briefly forget all of its peers.
- 06:31:10 — valkey/6 tries again. It asks the survivor who is alive. The survivor has just forgotten everyone, so it answers "only me". valkey/6 believes it is the last unit, skips the lock, and starts shutting down too.
- 06:31:12-18 — valkey/6 saves, stops, cleans up the survivor, and finishes fine at 06:31:20.
- 06:31:15 — meanwhile valkey/5 tries to clean up valkey/6, using a list it made before it stopped. But valkey/6 is now dying.
- 06:31:21 → 06:31:57 — Connection refused, then No route to host. valkey/5 gives up with an error, so its hook fails. Its notes are thrown away, and the cleanup is never finished.
- 06:32:11 onward — every retry now fails at the very first step, saving data to its own dead server: 06:32:11, 06:32:25, 06:33:01, 06:33:50, 06:35:21, 06:38:02, 06:43:18.
- 06:43:18 — the test has been waiting 1000 seconds for the unit to disappear. It gives up and fails.

reneradoi:
Thank you @skourta for sharing your findings. The PR addresses these by:
- having a central responsibility for the Sentinel reset, avoiding race conditions between competing units
- thereby only performing one Sentinel reset for multiple removals and only if actually required
- not relying on databag updates on departing units for the cleanup

The most important thing (from my point of view is) to not have any action or command after workload stop that could raise. Coming back to the storage detaching hook after the workload is down must absolutely be avoided.

src/managers/sentinel.py
skourta: We should be careful gating on planned_units. If we scale down by more than one unit (3->1) for example. The leader will expect 1 active sentinel and raise because after 1 unit scales down we still 2.

skourta:
> Thank you @skourta for sharing your findings. The PR addresses these by:
> 
> * having a central responsibility for the Sentinel reset, avoiding race conditions between competing units
> * thereby only performing one Sentinel reset for multiple removals and only if actually required
> * not relying on databag updates on departing units for the cleanup
> 
> The most important thing (from my point of view is) to not have any action or command after workload stop that could raise. Coming back to the storage detaching hook after the workload is down must absolutely be avoided.

I believe the race for the lock still persists. If the leader resets sentinel states and if in the gap when sentinel is rediscovering the rest it tries to acquire the lock it will skip the local thinking it is an only sentinel.

reneradoi:
> I believe the race for the lock still persists. If the leader resets sentinel states and if in the gap when sentinel is rediscovering the rest it tries to acquire the lock it will skip the local thinking it is an only sentinel.

The reset will only happen when all units are gone (`active_sentinels == planned_units`).

src/managers/sentinel.py
reneradoi: Yes, the leader will raise here and the event will be deferred until `active_sentinels == planned_units`.

In the integration test for scaling, we can see this:
- the two units are gone at 09:48:50 ([link](https://github.com/canonical/valkey-operator/actions/runs/33158857630/job/98823459138?pr=102#step:13:27862))
- the leader runs the reset at 09:49:19 ([link](https://github.com/canonical/valkey-operator/actions/runs/33158857630/job/98823459138?pr=102#step:13:28391))

src/managers/sentinel.py
Mehdi-Bendriss: Thanks René, i read it carefully, and indeed your changes reset every remaining Sentinel. 

My concern is more on the orchestration state: `sentinel_reset_required` doesn't store the identity of any departed sentinels, and `len(active_sentinels) == planned_units()` can accept the wrong set if a departing Sentinel still responds while a "survivor" is temporarily unreachable (base_events L328, sentinel L238). 

For ex:
- Scale down from 5 to 3
- 1 departing sentinel stopped
- The other departing sentinel still answers PING
- 1 real survivor is temporarily unreachable (for whatever temporary reason)
- 3 Sentinels answer PING, matching planned_units() == 3

The count is correct, but one of those 3 sentinels is still departing. So we may reset the wrong set. Which is why i suggested an identified list of sentinels.

A few other points:
- i just saw on [the docs](https://valkey.io/topics/sentinel/#adding-or-removing-sentinels) is that this procedure must happen in a rolling fashion, but also within at least 30 seconds interval.
- we also need to clear that flag at some point, and I think i missed this 
- we need to handle the reset on update status too

reneradoi: I understand the concern, the previous condition for checking if the Sentinel reset can be executed was to broad. I have added another condition to the check, additionally ensuring that not only the number matches, but also only those Sentinels are active that are expected from the deployment state (through the `is_active` flag of the respective unit). This is now safe against false-positives on the planned units check.

To the other points:
- The procedure is happening in rolling fashion one by one, and we respect the recommended wait time (through retries) in the `target_sees_all_others` method of `reset_sentinel_states`.
- The flag is only cleared if the whole procedure was successful ([here](https://github.com/canonical/valkey-operator/blob/shutdown-no-raise-after-stopping/src/managers/sentinel.py#L259))
- If any step fails, the method `remove_departed_sentinels_from_cluster` will raise to the event handler, which will defer the `relation-changed` event.
- We don't have to add it to `update-status` because deferred events will be issued again before `update_status` anyway.

src/managers/sentinel.py
reneradoi: I've enhanced the check condition, it no longer relies only on the planned units. This also covers potential false-positives when scale-down and scale-up happen at the same time.

src/managers/sentinel.py
```diff
@@ -219,6 +225,39 @@ def failover(self) -> None:
...
+
+        logger.info("Resetting Sentinels after unit removal")
+        primary_ip = self.get_primary_ip()
+        active_sentinels = self.get_active_sentinel_ips(primary_ip)
+        expected_sentinels = self.all_sentinel_endpoints()
+
+        # ensure that no departing sentinels is still alive before the reset
+        if (
+            any(sentinel for sentinel in active_sentinels if sentinel not in expected_sentinels)
+            or len(active_sentinels) != self.state.model.app.planned_units()
```
Mehdi-Bendriss:
```suggestion
            len(expected_sentinels) != self.state.model.app.planned_units()
            or len(active_sentinels) != self.state.model.app.planned_units()
            or set(active_sentinels) != set(expected_sentinels)
```

src/managers/sentinel.py
```diff
@@ -346,10 +385,17 @@ def get_active_sentinel_ips(self, hostname: str) -> list[str]:
...
         client = self._get_sentinel_client()
 
-        return [hostname] + [
+        all_sentinels = [hostname] + [
             sentinel["ip"] for sentinel in client.sentinels_primary(hostname=hostname)
         ]
 
+        active_sentinels = []
+        for sentinel in all_sentinels:
+            if self.is_sentinel_active(sentinel):
```
Mehdi-Bendriss: This is a deviation from the previous behavior, and now we're relying on ping where other parts of the code uses it to test on "membership".
So we need to separate "reachability" and "membership", where a unit not answering `PING` must be treated as an existing but unavailable member, not as a nonexistent member. 
If expected members are unreachable, scale-down should stop and retry, not bypass the lock or failover.
(re: `request_lock` and failover in `_scale_down_unit`)

src/managers/sentinel.py
Mehdi-Bendriss: re on the 30sec rolling reset, it seems to be still missing.
We reset the next sentinel immediately when discovery succeeds. The 5 sec tenacity.retry is only polling after a failure, it is not adding a delay between resets.

src/events/base_events.py
```diff
@@ -691,30 +726,16 @@ def _scale_down_unit(self) -> None:
...
-
-            # check health after scale down
-            self.charm.sentinel_manager.verify_expected_replica_count(active_sentinels)
-            # release lock
+        if len(active_sentinels) > 1:
             scale_down_lock.release_lock(primary_ip=primary_ip)
```
Mehdi-Bendriss: Given the lock lease has a 5min TTL, i think we don't need to crash this hook if it fails to reach the primary for whatever reason? 
```suggestion
            try:
                scale_down_lock.release_lock(primary_ip=primary_ip)
            except ValkeyWorkloadCommandError:
                logger.warning("Failed to release scale-down lock, lock will be released when its TTL expires")
```

Mehdi-Bendriss (commented):
Thanks René! I left some comments

src/managers/sentinel.py
reneradoi: updated

src/events/base_events.py
reneradoi: updated

src/managers/sentinel.py
reneradoi: Some context for the 30 second recommendation between resets: The reason for this is because it takes up to 30 seconds for the Sentinels to reach quorum. Instead of waiting a fixed time for this, we could ensure quorum by checking for the primary in the cluster. This would allow for earlier continuation if possible and raise to the event handler (+ defer) if it fails, avoiding further Sentinel resets.

I have added the primary check and would prefer to do this because it actually ensures we can continue, instead of just waiting.

src/managers/sentinel.py
reneradoi: I see your point, but I disagree here: The deviation is only relevant for scale down (the method is not used anywhere else), and I think it is actually good to deviate from previous behavior because it increases the chance of users not running force removal.

To explain: If the cluster is in unhealthy state, we might currently not be able to scale down or remove the application cleanly. This could lead to users forcing it, which in turn would then not allow the charm to save the data before shutting down cleanly, and could leave behind cloud resources. Forced removal would also bypass the lock and - even more important - failover.

The change I am introducing here is more robust against unavailable Sentinels in the event of scale down, thus reducing the chance of users forcing it.

With the responsibility for resetting Sentinels being moved to the leader unit and another event, we can make sure that scale down / removal succeeds even in edge or failure scenarios.  I think this is really worth it. We still keep the safeguard of not leaving the cluster behind in unavailable state if only part of the units are scaled down (by deferring the `peer_relation_changed` event if the Sentinel reset fails).

Mehdi-Bendriss (approved):
Thanks René!

skourta (approved):
Thank you @reneradoi 
Tested multiple scenarios all worked well.

PR #103, opened 2026-08-25
drop the protobuf and rustup installs left over from the valkey-glide fork

delgod:
Follow-up to #98. Now that `valkey-glide` installs as a published wheel, nothing in CI compiles Rust or protobuf any more — but every job still installs the toolchain for it. This removes the dead installs.

No behavioural change to any test; the only reason to look closely is whether something *else* was quietly relying on that toolchain (it isn't — see below).

### What goes

| Where | Removed |
|---|---|
| `ci.yaml` | the whole `Install dependencies` step in **unit-test** and in **static** — `apt update` + `libprotobuf-dev protobuf-compiler` for a package neither dependency group contains |
| `integration_test.yaml` | protobuf, `rustup`, and `rustup default 1.90.0` in both setup steps, plus the `apt update` that fed them (no `apt install` is left in either step) |
| `spread.yaml` | `rustup default 1.90.0` from the global `prepare` |
| `charmcraft.yaml`, `requirer-charm/charmcraft.yaml` | the `libprotobuf-dev` / `protobuf-compiler` build-packages |
| `tests/AGENTS.md` | the line telling contributors they need protobuf headers and a Rust toolchain to build the integration group |

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,11 +36,6 @@ jobs:
         run: |
           pipx install tox
           pipx install poetry
-      # to build Valkey-glide during tests
-      - name: Install dependencies
-        run: |
-          sudo apt update
-          sudo apt install libprotobuf-dev protobuf-compiler
       - name: Run tests
         run: tox run -e unit
       - name: Upload Coverage to Codecov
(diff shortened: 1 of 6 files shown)


skourta (approved):

reneradoi (approved):
No concerns from my side, we only added the protobuf dependency for building Valkey glide.

PR #104, opened 2026-08-25
refactor(backup): make the backup path backend-neutral

delgod:
Preparation for a second object store (Azure Blob). Adding one previously meant
editing every layer that named S3.

- `StorageBackend` (`common/storage_backend.py`) owns client construction, key
  layout and error translation; `BackupManager` keeps the policy.
- `BACKUP_CREDENTIAL_FIELDS` registry drives relation discovery, mutual
  exclusion and credential selection; guards and statuses read it.
- Errors carry `safe_code`, so botocore leaves `events/` entirely.

Two behavior changes: two related integrators now block on
`BACKUP_BACKENDS_CONFLICT` instead of one silently winning, and removing one
integrator re-drives the remaining backends' credentials handlers, so a backend
that the conflict guard had blocked gets configured instead of waiting for an
unrelated event to re-fire its handler.


src/managers/backup_backend.py
```diff
@@ -0,0 +1,224 @@
+#!/usr/bin/env python3
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Remote object-store backends for Valkey RDB backup/restore."""
```
reneradoi: This does not seem to be a real manager. We should try to be careful here, to not mess up the architecture. I'd rather consider this a "client" and move it to `common`, as it is intended to glue to the specific object storage providers.

src/managers/backup.py
```diff
@@ -121,68 +115,20 @@ def _pre_restore_path(self) -> pathops.PathProtocol:
...
-                endpoint_url=s3_parameters.endpoint,
-                config=Config(
-                    request_checksum_calculation="when_required",
-                    response_checksum_validation="when_required",
-                ),
-                verify=verify,
-            ),
-        )
-        return s3.Bucket(s3_parameters.bucket)
+    def _backend_for(self, params: "BackupCredentials") -> StorageBackend:
```
reneradoi: Can we rename this to `storage_backend` please? It is confusing to read `self._backend_for(params).ensure_container()`.

reneradoi: Another comment/question here: Is there no way to instantiate the storage backend directly when instantiating the backup manager class? By the way it is now, it is very hard to track down methods through the code (at least my IDE can't). It would also help to not have to do model validation at the level of calling a method, but rather at instantiation.

src/managers/backup.py
```diff
@@ -475,13 +392,16 @@ def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObje
         if scope == "unit" and self.state.unit_server.is_backup_in_progress:
             status_list.append(BackupStatuses.BACKUP_IN_PROGRESS.value)
 
-        if (
+        if scope == "app" and self.state.backup_backends_conflict:
```
reneradoi: nit: return early `if scope == "unit"`.

reneradoi (commented):
Not many concerns from my side, ony a few comments and questions. I've gone through the previous code and compared it, looks fine.

One question to your PR description:
> credentials are cleared on relation-broken so peers can't keep using them

I don't really see a behaviour change, this also happens in the current version (see https://github.com/canonical/valkey-operator/blob/9/edge/src/events/backup.py#L157). Am I missing something?

src/common/storage_backend.py
```diff
@@ -0,0 +1,224 @@
+#!/usr/bin/env python3
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Remote object-store backends for Valkey RDB backup/restore."""
```
delgod: Moved to `src/common/storage_backend.py` in 51f5db3.

src/managers/backup.py
delgod: Renamed — call sites now read `self.storage_backend.list_object_ids()`. 51f5db3.

src/managers/backup.py
delgod: Done in 51f5db3

src/managers/backup.py
delgod: Not in `__init__`: `charm.py:68` constructs `BackupManager` on every hook, including ones where no integrator is related and `state.active_backup_credentials` is `None`, so the constructor would have to raise charm-wide or hold a `None`.
Made it a `functools.cached_property` instead (51f5db3). A manager instance lives for exactly one hook, so every operation in that hook now shares one backend object rather than building a new one per call. It raises StorageBackendError when no credentials are stored, which the callers' existing handlers translate into their own backup/restore error — that also removed four duplicated `params is None` checks.

delgod:
> I don't really see a behaviour change, this also happens in the current version (see https://github.com/canonical/valkey-operator/blob/9/edge/src/events/backup.py#L157). Am I missing something?

Description updated.

src/managers/backup.py
reneradoi: That is fine for me, much better now, thank you!

reneradoi (approved):

src/common/storage_backend.py
```diff
@@ -0,0 +1,224 @@
...
+    The single place a credentials type maps to a backend: a new backend is added
+    here and nowhere above, so BackupManager never learns which stores exist.
+    ``ca_path`` is the charm-local CA bundle, used only by backends that need one.
+    """
+    if isinstance(params, S3Parameters):
+        return S3Backend(params, ca_path)
+    raise StorageBackendError(f"No storage backend for {type(params).__name__} credentials")
+
+
+def _s3_error_code(exc: ClientError) -> str:
```
skourta: Shouldn't this be inside the S3Backend class?

skourta (approved):
Thank you @delgod!
One nit.

src/common/storage_backend.py
delgod: fixed in 737bc212d065615643d40b6ac63d24216b94d63a

reneradoi (approved):
Re-approval

skourta (approved):

PR #105, opened 2026-08-25
docs - Post URL migration clean up

izmalk:
We merge the temporary branch to the default one.
It contains all the required configurations to publish the docs on canonical.com as well as updated Sphinx Stack.


reneradoi (approved):
I have to trust that the PR does what it is supposed to, can't review the changes for a lack of context/understanding, but that's ok. One thing for your attention: The `compare_url` check is failing. The message makes sense because some pages where removed, but better check yourself to be sure.

izmalk:
Yes, the compare URL is pointing towards pages that never even existed in the repo. That is because it track the previous state of the template repo that had them (style guides for MD and RST). THis should be resolved after we merge this to the default `9/edge` - next check will compare against the default branch.

Mehdi-Bendriss (approved):
Thanks Vladimir, this said the `Check for removed URLs / build-docs` check is failing. Is this getting addressed in a next PR?

izmalk:
> Check for removed URLs / build-docs check is failing.

Yes, it is failing, sorry about that, but I assure you that everything is good and this is a false positive.

The way this check works - it checks out the PR branch, then the base branch, builds both and checks the difference of all URLs served, including redirects.

So, the reason this particular check is getting triggered - as part of a major Sphinx Stack upgrade, I deleted some [boiler-plate redirects](https://github.com/canonical/valkey-operator/pull/105/changes#diff-85933aa74a2d66c3e4dcdf7a9ad8397f5a7971080d34ef1108296a7c6b69e7e3L234-L239) that we had before for no good reason (we inherited them from an older template, but there is no harm in having them).

```text
redirects = {
    'reference/doc-cheat-sheet-myst/': '../myst-syntax-reference',
    'reference/doc-cheat-sheet/': '../rst-syntax-reference',
    'reference/style-guide-myst/': '../myst-syntax-reference',
    'reference/style-guide/': '../rst-syntax-reference',
}
```

The best way to fix the check is to merge this, so the next check would compare against the new version.
Alternatively, I can re-introduce the redundant redirects, as they do no harm. That would make the check green, but the redirects are redundant.

PR #85, opened 2026-07-14
[DPE-10363] Move integration tests from MicroK8s to Canonical K8s

reneradoi:
The goal of this PR is to replace MicroK8s with Canonical K8s as Kubernetes platform for integration tests.

As a workaround for #86 , we add enough storage for larger datasets on the integration tests for scaling and storage reuse.

PR #106, opened 2026-08-26
feat(backup): add the Azure Blob backup backend

delgod:
Adds Azure Blob as a backup store alongside S3.

Relate `azure-storage-integrator` on `azure-credentials`. The two backends
are mutually exclusive: relate both and the app blocks; remove one and the
survivor converges on its own.

Six commits, each green and reviewable on its own.

Two things worth a reviewer's attention:
- The account name is passed to `BlobServiceClient` explicitly. Given a bare
  key the SDK infers it from the hostname, which fails outright for any
  endpoint that isn't `<account>.blob.*` (emulator, custom domain).
- `_az_error_code` reads `.value`; `error_code` is a `(str, Enum)` whose
  `str()` is `StorageErrorCode.AUTHENTICATION_FAILED`, and action results
  are world-readable.

### DIFF
--- a/src/common/storage_backend.py
+++ b/src/common/storage_backend.py
@@ -8,18 +8,23 @@ from __future__ import annotations
 
 import logging
 import pathlib
+from enum import Enum
 from typing import IO, TYPE_CHECKING, BinaryIO, NamedTuple, Protocol, cast
 from urllib.parse import urlparse
 
 import boto3
+from azure.core.exceptions import HttpResponseError, ResourceExistsError
+from azure.storage.blob import BlobServiceClient
 from boto3.s3.transfer import TransferConfig
 from botocore.config import Config
 from botocore.exceptions import ClientError
 
 from common.exceptions import StorageBackendError
-from core.models import BackupCredentials, S3Parameters
+from core.models import AzureStorageParameters, BackupCredentials, S3Parameters
+from literals import AZURE_HTTPS_PROTOCOLS
 
 if TYPE_CHECKING:
+    from azure.storage.blob import BlobClient, ContainerClient
     from mypy_boto3_s3.literals import BucketLocationConstraintType
     from mypy_boto3_s3.service_resource import Bucket, S3ServiceResource
 
@@ -83,6 +88,8 @@ def build_backend(params: BackupCredentials, ca_path: pathlib.Path) -> StorageBa
     """
     if isinstance(params, S3Parameters):
         return S3Backend(params, ca_path)
+    if isinstance(params, AzureStorageParameters):
+        return AzureBackend(params)
     raise StorageBackendError(f"No storage backend for {type(params).__name__} credentials")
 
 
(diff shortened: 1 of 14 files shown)


tests/integration/backup/conftest.py
```diff
@@ -155,3 +157,169 @@ def s3_bucket(microceph):
...
+    Reachable from charm units at the host's routable IP -- never loopback,
+    which from a unit resolves to the unit itself (same rationale as the
+    MicroCeph fixture). `endpoint` embeds the dev account in the path and
+    `connection-protocol` is plain `http`, which azure-storage-integrator
+    accepts for Blob REST access, so it talks to the emulator, not real Azure.
+    """
+    host_ip = _host_ip()
+    if not (port := _ensure_azurite(host_ip)):
+        pytest.skip("Azurite needs a host docker daemon or a running microk8s")
+
```
reneradoi: Hm, not sure if that is really helpful. In practise it means we will always skip integration tests, see https://github.com/canonical/valkey-operator/actions/runs/33184650060/job/98931635312?pr=106#step:8:2802

This is because:
- we run K8s integration tests on Canonical K8s
- in order to be able to use Canonical K8s, we purge the existing docker setup on integration test preparation (see https://github.com/canonical/valkey-operator/blob/9/edge/.github/workflows/integration_test.yaml#L112)

src/common/storage_backend.py
```diff
@@ -225,3 +232,125 @@ def delete(self, backup_id: str) -> None:
...
+        return f"azure://{host}/{self.params.container}/{self.params.path}"
+
+    def _account_url(self) -> str:
+        """Explicit endpoint (emulator/private) if given, else the derived Blob host."""
+        if self.params.endpoint:
+            return self.params.endpoint
+        scheme = "https" if self.params.connection_protocol in AZURE_HTTPS_PROTOCOLS else "http"
+        return f"{scheme}://{self.params.storage_account}.blob.core.windows.net"
+
+    def _service(self) -> BlobServiceClient:
```
reneradoi: Instead of building a blob service client that builds a container client that builds a blob client, we could simplify and just build a container client and (if required) get a blob client from that.

src/common/storage_backend.py
```diff
@@ -225,3 +232,125 @@ def delete(self, backup_id: str) -> None:
...
+        return code.value if isinstance(code, Enum) else str(code)
+
+    def ensure_container(self) -> None:
+        """Create the configured container; tolerates an already-existing one."""
+        try:
+            self._container().create_container()
+        except ResourceExistsError:
+            # Subclass of HttpResponseError, so this clause must stay first.
+            logger.info("Using existing container %s", self.params.container)
+        except HttpResponseError as e:
```
reneradoi: Here and in other places of error handling: Should we not also consider other errors? The `HttpResponseError` only covers 4xx and 5xx error codes, but how about network connectivity or others? I believe it would be best to use the `AzureError` as a catch-all exception handler next to the use-case relevant ones (ResourceExistsError in this case).  

See https://learn.microsoft.com/en-us/azure/developer/python/sdk/fundamentals/errors#exception-hierarchy

reneradoi (commented):
Thank you Mykola! The previous work for unifying the backup/restore architecture really pays off, this new feature can be integrated very cleanly.

I've left only three comments, two of them important ones (testing setup is skipped, handling all errors). Once the testing setup topic is addressed, I will also perform testing myself.

src/common/storage_backend.py
```diff
@@ -225,3 +232,125 @@ def delete(self, backup_id: str) -> None:
...
+    """Azure Blob object store via azure-storage-blob."""
+
+    def __init__(self, params: "AzureStorageParameters"):
+        self.params = params
+
+    @property
+    def location(self) -> str:
+        """Account host, container and prefix -- the audit trail's destination.
+
+        ``hostname``, not the raw URL: an endpoint carrying a SAS token or
```
skourta: Is this from an old implementation?

tests/integration/backup/conftest.py
```diff
@@ -155,3 +157,169 @@ def s3_bucket(microceph):
...
+
+def _start_azurite_microk8s() -> None:
+    """Run Azurite as a microk8s Deployment behind a NodePort; idempotent.
+
+    The manifest goes in on stdin rather than as a path: microk8s is a strict
+    snap with a private /tmp, so it cannot read a file the test process wrote
+    to a temp directory.
+    """
+    _run("microk8s", "kubectl", "apply", "-f", "-", input=_AZURITE_MANIFEST)
```
skourta: Why are we using microk8s on this branch? Was this branched out before we moved to canonical k8s?

tests/integration/backup/conftest.py
skourta: +1 both tests are skipped.

tests/integration/backup/conftest.py
```diff
@@ -155,3 +157,169 @@ def s3_bucket(microceph):
...
+# ── Azurite (Azure Blob emulator) ─────────────────────────────────────────────
+# Unlike MicroCeph (a snap, installed above), Azurite ships only as an OCI image,
+# so it needs a container runtime. Two are tried in order -- a host `docker`
+# daemon (what the CI runners have) and microk8s behind a NodePort (what a local
+# k8s dev box has) -- and the suite skips when neither is available.
+AZURITE_PORT = 10000
+AZURITE_NODE_PORT = 30000
+AZURITE_CONTAINER = "valkey-azurite"
+AZURITE_NAMESPACE = "valkey-azurite"
+_AZURITE_IMAGE = "mcr.microsoft.com/azure-storage/azurite:latest"
```
skourta: Should we pin a version rather than `latest`?

src/common/storage_backend.py
skourta: +1 this would also means we will miss killing the process of the `valkey-cli --rdb` on upload if it raises.

skourta (commented):
Thank you @delgod 
I am mostly worried about the skipped integration tests, the use of microk8s, and the error handling as @reneradoi pointed out.

tests/integration/backup/conftest.py
delgod: Both spread tasks run for real now. 66d5298
 I am not sure if it is good idea to run `sudo npm install azurite` inside the test.
 Any better idea?

src/common/storage_backend.py
delgod: done 2ee8b45

src/common/storage_backend.py
delgod: Both fixed. Every clause is AzureError now (ServiceRequestError was escaping), and S3Backend had the same hole — BotoCoreError is disjoint from ClientError — so that's widened too (3c09364). create_backup also gained a catch-all proc.kill(), so no future SDK exception can orphan the producer (c06e113).

src/common/storage_backend.py
delgod: No — location is a live Protocol member, mirroring S3Backend.location. The "SAS token" wording was stale though (shared key only), so I trimmed it to the userinfo rationale S3 uses. 499b756

tests/integration/backup/conftest.py
delgod: Not a stale branch — #85 is an ancestor here; microk8s was a local dev-box fallback. Gone now, no cluster involved at all (66d5298). tests/AGENTS.md was part of the confusion (it still claimed CI pinned microk8s 1.34-strict); corrected in 2c8ca74.

tests/integration/backup/conftest.py
reneradoi: I don't like npm...

... [but we could snap it!](https://github.com/canonical/azurite-snap)
```
$ sudo snap install azurite --edge
azurite (edge) 3.37.0 from Rene Radoi (reneradoi) installed
$ sudo snap start azurite.blob
Started.
$ sudo snap logs azurite.blob
2026-09-03T10:19:01+02:00 systemd[1]: Started snap.azurite.blob.service - Service for snap application azurite.blob.
2026-09-03T10:19:03+02:00 azurite.blob[7167]: Azurite Blob service is starting on 127.0.0.1:10000
2026-09-03T10:19:03+02:00 azurite.blob[7167]: Azurite Blob service successfully listens on http://127.0.0.1:10000
```

src/common/storage_backend.py
```diff
@@ -225,3 +235,126 @@ def delete(self, backup_id: str) -> None:
...
+    def upload(self, backup_id: str, reader: IO[bytes]) -> None:
+        """Stream ``reader`` into the blob for ``backup_id`` (sequential block blob).
+
+        ``length=None`` + ``max_concurrency=1``: the source is a non-rewindable
+        pipe from ``valkey-cli --rdb -``, so the SDK must stage blocks
+        sequentially off ``read()`` rather than seek around a known-size stream.
+        """
+        try:
+            self._blob(backup_id).upload_blob(
+                reader, blob_type="BlockBlob", overwrite=True, length=None, max_concurrency=1
```
reneradoi: Should we really overwrite if the backup file already exists? Furthermore, would this deviate from the behavior of the S3 backend?

reneradoi (approved):
One minor question remaining, but just for confirmation. Other than that, I'm good.

src/common/storage_backend.py
delgod: fixed in e36276fd969544f1cc0f39801ce61786c644820d

skourta (approved):
Thank you Mykola.
It looks good to me.

src/common/storage_backend.py
reneradoi: Thank you

tests/integration/backup/conftest.py
delgod: it is outdated

